PR #8, opened 2024-07-17
Refactor list comprehension & bug fixes

danielgerlag:
- Fixes the `WHERE clauses interpreted as list filters when following a WITH projection without an AS` bug
- Refactored expression evaluation logic out of the reduce function, and into the expression evaluator
- Encapsulated list comprehension evaluation around the `IteratorExpression`
- Removed redundant expression variants
- Added LazyScalar function type for functions that do not want the args pre-resolved
- Additional test cases

### DIFF
--- a/core/src/evaluation/expressions/mod.rs
+++ b/core/src/evaluation/expressions/mod.rs
@@ -8,7 +8,7 @@ use async_recursion::async_recursion;
 use chrono::{
     Datelike, Duration, FixedOffset, NaiveDate, NaiveDateTime, NaiveTime, Timelike, Weekday,
 };
-use drasi_query_ast::ast::{self, ParentExpression};
+use drasi_query_ast::ast;
 
 use crate::evaluation::temporal_constants;
 use crate::evaluation::variable_value::duration::Duration as DurationStruct;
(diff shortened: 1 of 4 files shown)


query-cypher/src/lib.rs
```diff
@@ -241,6 +234,32 @@ peg::parser! {
                 "[" __* c:expression() ** (__* "," __*) __* "]" { ListExpression::list(c) }
             }
 
+            #[cache_left_rec]
+        rule iterator() -> Expression
```
ruokun-niu, line 238: wow :heart_eyes: looks so much better

core/src/evaluation/expressions/mod.rs
```diff
@@ -1649,30 +1549,126 @@ impl ExpressionEvaluator {
...
-                ast::UnaryExpression::Identifier(ident) => ident,
-                _ => return Err(EvaluationError::InvalidType),
+        expression: &ast::Expression,
+    ) -> Result<(Arc<str>, VariableValue), EvaluationError> {
+        match expression {
+            ast::Expression::BinaryExpression(exp) => match exp {
+                ast::BinaryExpression::Eq(var, val) => {
+                    let variable = match *var.clone() {
+                        ast::Expression::UnaryExpression(exp) => match exp {
+                            ast::UnaryExpression::Identifier(ident) => ident,
```
ruokun-niu, line 1562: Sorry what would be an example of an identifier expression again?

danielgerlag: like a variable, eg. `x` from the expression: `x = 1`

PR #11, opened 2024-07-23
Review ZonedTime and ZonedDateTime functions

ruokun-niu:


core/src/evaluation/functions/temporal_instant/temporal_instant.rs
```diff
@@ -260,9 +265,14 @@ impl ScalarFunction for Time {
         _expression: &ast::FunctionExpression,
         args: Vec<VariableValue>,
     ) -> Result<VariableValue, EvaluationError> {
-        if args.len() != 1 {
+        if args.len() > 1 {
             return Err(EvaluationError::InvalidArgumentCount("time".to_string()));
         }
+        if args.len() == 0 {
+            let local = Local::now().time();
+            let timezone = FixedOffset::east_opt(0).unwrap();
```
danielgerlag, line 273: Is `east_opt(0)` the same thing as UTC?

ruokun-niu: Yes, I believe so. This essentially means that we have a 0-second offset from UTC. We are using this instead of the Utc time because ZoneTime takes in a localtime and a FixedOffset

PR #13, opened 2024-08-05
Updated EvaluationError variants and removed .unwrap() in expressions and aggregating functions

ruokun-niu:
The remainings will be updated in a separate PR


core/src/evaluation/functions/aggregation/avg.rs
```diff
@@ -63,7 +66,10 @@ impl AggregatingFunction for Avg {
             }
             VariableValue::Integer(n) => {
                 *count += 1;
-                *sum += n.as_i64().unwrap() as f64;
+                *sum += match n.as_i64() {
+                    Some(n) => n as f64,
```
agentofreality, line 82: Shouldn't we be summing i64s and converting to f64 at the point of division?

core/src/evaluation/functions/aggregation/linear_gradient.rs
```diff
@@ -205,8 +205,14 @@ impl AggregatingFunction for LinearGradient {
...
 fn extract_parameter(p: &VariableValue) -> Result<f64, EvaluationError> {
     let result = match p {
-        VariableValue::Float(n) => n.as_f64().unwrap(),
-        VariableValue::Integer(n) => n.as_i64().unwrap() as f64,
+        VariableValue::Float(n) => match n.as_f64() {
+            Some(n) => n,
+            None => return Err(EvaluationError::ConversionError),
+        }
+        VariableValue::Integer(n) => match n.as_i64() {
+            Some(n) => n as f64,
```
agentofreality, line 237: Why is this an f64?

core/src/evaluation/functions/aggregation/linear_gradient.rs
```diff
@@ -205,8 +205,14 @@ impl AggregatingFunction for LinearGradient {
...
+            None => return Err(EvaluationError::ConversionError),
+        },
         VariableValue::Duration(d) => d.duration().num_milliseconds() as f64,
```
agentofreality, line 243: Not sure why these are being cast to f64s either?

core/src/evaluation/functions/aggregation/avg.rs
ruokun-niu: I think the value of `sum` is defined to be a `f64` in the accumulator. We need to convert to f64 in order to add this value to the `sum`

core/src/evaluation/functions/aggregation/linear_gradient.rs
ruokun-niu: From my understanding, the x and y values in the accumulator are `f64`.  This helper function `extract_parameter` also returns `Result<f64, EvaluationError>`

PR #20, opened 2024-08-12
Remove .unwrap() in Cypher Scalar functions

ruokun-niu:


core/src/evaluation/functions/cypher_scalar/timestamp.rs
```diff
@@ -13,16 +13,25 @@ impl ScalarFunction for Timestamp {
...
+            });
         }
         let now = std::time::SystemTime::now();
-        let since_epoch = now.duration_since(std::time::UNIX_EPOCH).unwrap();
+        let since_epoch = match now.duration_since(std::time::UNIX_EPOCH) {
+            Ok(d) => d,
+            Err(e) => {
+                return Err(FunctionError {
+                    function_name: "timestamp".to_string(),
+                    error: FunctionEvaluationError::InvalidFormat { expected: "A valid Duration".to_string() },
```
danielgerlag: Could we include the format(s) expected? eg. `hh:mm:ss`, etc...

core/src/evaluation/functions/temporal_instant/temporal_instant.rs
```diff
@@ -67,31 +74,49 @@ impl ScalarFunction for Date {
...
+                            let tz_str = match tz.as_str() {
+                                Some(tz_str) => tz_str,
+                                None => "UTC",
+                            };
                             let tz: Tz = match tz_str.parse() {
                                 Ok(tz) => tz,
-                                Err(_) => return Err(EvaluationError::InvalidType),
+                                Err(_) => return Err(FunctionError {
+                                    function_name: expression.name.to_string(),
+                                    error: FunctionEvaluationError::InvalidFormat { expected: "A valid Date".to_string() },
```
danielgerlag: Could we give format examples, eg. `yyyy-mm-dd` or `d mmmm yyyy`, etc...

danielgerlag: or a link to the ISO standard

PR #21, opened 2024-08-14
Removed .unwrap() in VariableValue + resolved warnings

ruokun-niu:
# Description

Finals bits of unwrap to clean up

## Type of change

Fixes: #issue_number

### DIFF
--- a/core/src/evaluation/functions/future/future_element.rs
+++ b/core/src/evaluation/functions/future/future_element.rs
@@ -3,8 +3,8 @@ use std::sync::Arc;
 use crate::evaluation::context::SideEffects;
 use crate::evaluation::functions::ScalarFunction;
 use crate::evaluation::variable_value::VariableValue;
-use crate::evaluation::{FunctionError, FunctionEvaluationError};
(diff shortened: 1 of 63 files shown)


core/src/evaluation/functions/future/future_element.rs
```diff
@@ -73,7 +89,13 @@ impl ScalarFunction for FutureElement {
                     context.get_transaction_time(),
                     due_time,
                 )
-                .await?;
```
danielgerlag, line 76: If you implement `Into<FunctionError> for IndexError`, you can just use the `?` operator and it will automatically convert it to a FunctionError

ruokun-niu: Ah good point

ruokun-niu: But the problem is that in order to create the FunctionError, we would need to provide a function name.

danielgerlag: right, that's unfortunate.

PR #28, opened 2024-10-07
Updated codecov versioning

ruokun-niu:
# Description

- Updated the version of Codecov Github Action to allow PRs from a forked repo to be able to execute the workflow.
- Updated codecov configurations:
  - Updated the precision, rounding mode and the green range of the values
- Failed upload to codecov will result the pipeline to fail
## Type of change

Fixes: #issue_number

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,6 @@
 codecov:
-  require_ci_to_pass: yes
+  require_ci_to_pass: true
+  allow_pseudo_compare: false 
 
 coverage:
   precision: 2  # Precision of the coverage value
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,15 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "30...100"
+  status:
+    project:
+      default:
+        informational: true
+        target: auto
+        threshold: 0%
+    patch:
+      default: false
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,14 +2,19 @@ codecov:
   require_ci_to_pass: yes
 
 coverage:
-  precision: 2
+  precision: 2  # Precision of the coverage value
   round: down
-  range: "30...100"
+  range: "50...100"   # The value range where you want the value to be green 
   status:
     project:
       default:
         informational: true
-        target: auto
+        target: auto  # default; compares coverage to the previous base commit
         threshold: 0%
     patch:
       default: false
+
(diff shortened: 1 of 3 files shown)


ruokun-niu:
Screenshot 2024-10-07 at 11 25 31 AM
:grin:

agentofreality:
@ruokun-niu you mention that you updated it so that failed uploads to codecov will not fail the build. Do we expect failed uploads? If the upload is failing, how do we get notified?

ruokun-niu:
> @ruokun-niu you mention that you updated it so that failed uploads to codecov will not fail the build. Do we expect failed uploads? If the upload is failing, how do we get notified?

I think this is the default behavior and the only way we can tell is by going into the Github actions. My initial intuition is that if the upload fails but everything before that succeeded, this indicates that everything before the last step worked and we were able to generate a code coverage but were unable to upload it.

However, now that I think about it more, maybe we should just set to true as discussed in this issue https://github.com/codecov/codecov-action/issues/1348.

ruokun-niu:
Also I think Radius set this setting to false because tokenless uploads can be throttled and block the build 
https://github.com/radius-project/radius/commit/1da9e8e3a1ad3846e169815e2ad8d9afcae9e53f

ruokun-niu:
> Please clarify in the main PR Description whether failed codecov push is failing the build or not.
> 
> Also, you say above about something "weird with the workflow" that you are investigating, but dont clarify the result of that investigation.

Sorry about that. I have updated the main PR description. 

The "weird thing" that I have encountered yesterday was that the upload were failing because codecov complained that it was unable to perform a tokenless upload because the repository is private. However, `drasi-core` is public at the moment.

I dug a little deeper and noticed that the base commit for codecov is pointing at a wrong, outdated one. I believe that this is a commit in the repo before we erased our history. 
Screenshot 2024-10-09 at 8 06 55 AM
Screenshot 2024-10-09 at 8 07 06 AM

Codecov failed to switch to pointing to a new commit as the base commit. Since the commit that it is currently pointing to does not belong to any repo, it might have mistakenly categorized it as a private repo. I want to spend some time today trying to see if we can change the base commit on codecov. Will keep the team posted during sync and in this PR

ruokun-niu:
@agentofreality  @danielgerlag  I am closing this PR as I have found an alternative to Codecov that is much easier to use. I have created another PR (https://github.com/drasi-project/drasi-core/pull/30) and I wrote a short paragraph that explains my reasonings and compares the two coverage reporting tools. Thanks!

PR #38, opened 2024-12-12
Nested Property support

danielgerlag:
- adds support for nested properties of deep objects
  - eg. `p.Metadata.Name`
  - eg. `p.Metadata["Name"]` (this syntax is documented here: https://neo4j.com/docs/cypher-manual/current/syntax/operators/#query-operators-property)
- adds support for escaping identifier names using backticks as per https://neo4j.com/docs/cypher-manual/current/syntax/naming/#symbolic-names-escaping-rules
- Existing list comprehension functionality can be used to navigate arrays within an object.

Tests added to exercise the following syntax:

```cypher
MATCH (p:Pod)
  RETURN
    p.metadata.name as name,
    p.metadata.namespace as namespace,
    p.metadata.labels.app as app,
    p.metadata.annotations.`dapr.io/app-id` as app_id,
    p.metadata.annotations['dapr.io/app-id'] as app_id2,
    p.spec.containers[0].image as container_0_image,
    reduce(acc=0, x in p.status.containerStatuses | acc + x.restartCount) as total_restart_count,
    head(container IN p.status.containerStatuses WHERE container.name = 'sidecar').containerID as sidecar_container_id
```


ruokun-niu (approved):

core/src/evaluation/variable_value/zoned_datetime.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::evaluation::temporal_constants::{self, UTC_FIXED_OFFSET};
```
amansinghoriginal, line 15: This is fixing the code, but not necessarily related to the Nested properties right?

query-cypher/src/lib.rs
```diff
@@ -228,6 +228,8 @@ peg::parser! {
                 a:(@) __* "%" __* b:@ { BinaryExpression::modulo(a, b) }
                 a:(@) __* "^" __* b:@ { BinaryExpression::exponent(a, b) }
                 --
```
amansinghoriginal, line 230: It seems that we've taken the choice of an inline grammar definition over having a separate `.peg` file.
Was the reason to avoid having an extra step in the build process?

amansinghoriginal (approved):

agentofreality (approved):

core/src/evaluation/variable_value/zoned_datetime.rs
danielgerlag: Yes, I think it was the lint tool

query-cypher/src/lib.rs
danielgerlag: I don't think there was a conscious decision, it started as an experiment and evolved from there.

Issue #43, opened 2025-04-08
Add support for Collect function in Cypher queries


amansinghoriginal:
### Overview of feature request

If one has a graph of orders, line items and products, then one can write a query today like:

```cypher
MATCH (o:Order {orderId: 'order123'})-[:HAS]->(li:LineItem)
RETURN o.orderId AS orderId,
       o.date AS orderDate,
       li.productId AS productId,
       li.quantity AS quantity
```

If one order has 3 line items, the above query will return 3 rows in the result for that same order - once for each line item. But using the `collect` function I could write a query like this:

```cypher
MATCH (o:Order {orderId: 'order123'})-[:HAS]->(li:LineItem)
RETURN o.orderId AS orderId, 
       o.date AS orderDate, 
       collect({productId: li.productId, quantity: li.quantity}) AS lineItems
```

When the result set changes, drasi emits diffs. 
In the example above, when the collection changes, we can emit a diff.
For instance, if a new line item gets added, drasi could emit a diff like:

```
Updating {
  before: {
    "key": "order1", 
    "orderId": "order123",
    "userId": "user456", 
    "orderTotal": 99.95,
    "lineItems": [
      {"productId": "prod001", "quantity": 2},
      {"productId": "prod002", "quantity": 5}
    ]
  },
  after: {
    "key": "order1", 
    "orderId": "order123",
    "userId": "user456", 
    "orderTotal": 99.95,
    "lineItems": [
      {"productId": "prod001", "quantity": 2},
      {"productId": "prod002", "quantity": 5},
      {"productId": "prod003", "quantity": 1}
    ]
  }
}
```

### Would you like to support us?

- [x] Yes, I would like to support you

PR #49, opened 2025-04-24
Implement `collect()` function for aggregating values into lists

amansinghoriginal:
# Description

This PR implements the collect() function for Drasi Core which enables collecting values into lists during query execution. This is a common aggregation function in Cypher that allows for creating structured results with nested collections.

Fixes: #43 

## Type of change
- [ ] This pull request fixes a bug in Drasi and has an approved issue (issue link required).
- [x] This pull request adds or changes features of Drasi and has an approved issue (issue link required).
- [ ] This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Drasi (issue link optional).

## Features
- Added implementation of the collect() aggregation function
- Support for collecting any value type (primitives, maps, etc.) into lists
- Proper handling of additions, updates, and deletions of collection elements
- Support for null value handling (nulls are ignored in collection)
- Implemented efficient state tracking with a count-based accumulator

## Implementation Details
The implementation tracks collection elements efficiently using a counting approach:

- Uses a HashMap to store unique values and their counts
- Properly handles duplicates by maintaining count information
- Implements revert operations to correctly remove elements when needed
- Emits appropriate diffs when collections change

## Testing
- Added unit tests covering basic functionality, duplicates, nulls, and edge cases
- Added a comprehensive end-to-end test with a movie database scenario


core/src/evaluation/functions/.DS_Store
danielgerlag, line 1: We should gitignore this file

core/src/evaluation/functions/aggregation/mod.rs
```diff
@@ -80,6 +84,13 @@ pub enum ValueAccumulator {
         cov: f64,
     },
     Value(ElementValue),
+    CollectMultiset {
```
danielgerlag, line 87: You would need to update the RocksDb and Redis storage models to support these new variants, this will only work for "In Memory"

core/src/evaluation/functions/aggregation/mod.rs
```diff
@@ -80,6 +84,13 @@ pub enum ValueAccumulator {
...
+        list: Vec<ElementValue>,
```
danielgerlag, line 88: What would the trade offs be of just using `ValueAccumulator::Value(ElementValue::List)`, instead of these new variants?

core/src/evaluation/functions/aggregation/mod.rs
```diff
@@ -80,6 +84,13 @@ pub enum ValueAccumulator {
...
+        counts: std::collections::HashMap<ElementValue, usize>,
+    },
+    CollectCounts {
```
danielgerlag, line 91: What are the purposes / differences of each of these new accumulator variants?

shared-tests/src/in_memory/mod.rs
```diff
@@ -122,6 +122,17 @@ mod remap {
     }
 }
 
+mod collect_movies {
+    use super::InMemoryQueryConfig;
+    use crate::use_cases::*;
+
+    #[tokio::test]
+    pub async fn collect_movies() {
```
danielgerlag, line 130: We should surface this test in the Redis & RocksDb tests as well.

PR #51, opened 2025-05-10
Optional match clause

danielgerlag:
# Description

This PR adds support for the [OPTIONAL MATCH](https://neo4j.com/docs/cypher-manual/current/clauses/optional-match/) clause in cypher.


ruokun-niu (approved):

core/src/path_solver/mod.rs
```diff
@@ -354,16 +373,22 @@ fn merge_node_match<'b>(
     mtch: &NodeMatch,
     slots: &'b mut Vec<match_path::MatchPathSlot>,
     alias_map: &'b mut HashMap<Arc<str>, usize>,
+    path_index: usize,
+    optional: bool,
 ) -> Result<usize, EvaluationError> {
     match &mtch.annotation.name {
         Some(alias) => {
             if let Some(slot_num) = alias_map.get(alias) {
+                slots[*slot_num].optional = optional && slots[*slot_num].optional;
```
amansinghoriginal, line 382: Will this work for a scenario like:

```
MATCH (n:Person)
OPTIONAL MATCH (n)-[:LIKES]->(m:Movie)
  RETURN n.name, m.title
```

Given this graph:
  (Alice:Person)
  (Bob:Person)-[:LIKES]->(Inception:Movie)

I guess the expected results will be:
  "Alice"   | NULL       ← Alice should appear even though she doesn't have movies she liked.
  "Bob"     | "Inception"
  
  I guess here we will create a slot for the `Person` variable for the first match clause.
And then when we merge the slot from second (Optional) Match clause, we will end up marking `slots[n_slot].optional = true && false = false`, **thereby making the slot NOT optional??**

Would this mean that in the example scenario above, the entry for Alice will not show up with NULL ?

amansinghoriginal (approved):
Overall looks good to me, but I had a doubt about the scenario posted.

core/src/path_solver/mod.rs
danielgerlag: Yes it should show up with a NULL.

We have an equivalent test in the PR:

```
MATCH 
  (i:Invoice)
OPTIONAL MATCH
  (i)-[:RECONCILED_TO]->(p:Payment)
RETURN
  i.id AS id,
  i.amount AS amount,
  p.amount AS payment_amount
```
 Where we assert that an invoice with no payment shows up with a `null` for the payment amount.

PR #57, opened 2025-06-10
Add collect() Aggregation Function

amansinghoriginal:
# Description

## Summary
This PR implements the `collect()` aggregation function in Drasi Core to solve the known issue where queries with multiple WITH clauses containing different grouping keys produce multiple partial results instead of properly aggregated results.

## Problem Solved
When using queries like:
```cypher
MATCH (p:products)-[:PRODUCT_TO_ORDER_ITEM]->(oi:orderItem),
      (r:reviews)-[:REVIEW_TO_PRODUCT]->(p)
WITH p, r, count(oi) as orderCount, avg(oi.quantity) as avgQuantity
WITH p, orderCount, avgQuantity, avg(r.rating) as avgRating, count(r) as reviewCount
RETURN p.productId, orderCount, avgQuantity, avgRating, reviewCount
```

The query would produce multiple partial results per product (e.g., 7 results for one product, 4 for another) instead of one properly aggregated result per product.

## Solution
The `collect()` function aggregates values into lists, avoiding the cartesian product issue:

```cypher
MATCH (p:products)-[:PRODUCT_TO_ORDER_ITEM]->(oi:orderItem)
WITH p, collect(oi.quantity) as order_quantities
OPTIONAL MATCH (r:reviews)-[:REVIEW_TO_PRODUCT]->(p)
WITH p, order_quantities, collect(r.rating) as review_ratings
RETURN p.productId, size(order_quantities) as orderCount, 
       size(review_ratings) as reviewCount
```

## Implementation Details
- Implements the `AggregatingFunction` trait with proper `apply()`, `revert()`, and `snapshot()` methods
- Handles null values correctly (ignores them)
- Supports collecting any variable value type
- Works seamlessly with all storage backends (in-memory, Garnet, RocksDB) without requiring backend-specific changes

## Testing
- 9 unit tests covering edge cases
- 7 shared integration tests
  - Solution validation with collect()
  - Filtering with WHERE clauses
  - Collecting complex objects
  - Multiple collect() in same WITH clause
  - Mixed data types
- Tests verified on all storage backends

### DIFF
--- a/core/src/evaluation/functions/aggregation/collect.rs
+++ b/core/src/evaluation/functions/aggregation/collect.rs
@@ -1,0 +1,585 @@
+// Copyright 2024 The Drasi Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{fmt::Debug, sync::Arc};
+
+use async_trait::async_trait;
+use drasi_query_ast::ast;
+
+use crate::{
+    evaluation::{
+        variable_value::VariableValue, ExpressionEvaluationContext, FunctionError,
+        FunctionEvaluationError,
+    },
+    interface::ResultIndex,
+    models::ElementValue,
+};
+
+use super::{super::AggregatingFunction, Accumulator, ValueAccumulator};
+
+/// Collect aggregation function that collects all values into a list
+pub struct Collect {}
+
+#[async_trait]
+impl AggregatingFunction for Collect {
+    fn initialize_accumulator(
+        &self,
+        _context: &ExpressionEvaluationContext,
+        _expression: &ast::FunctionExpression,
+        _grouping_keys: &Vec<VariableValue>,
+        _index: Arc<dyn ResultIndex>,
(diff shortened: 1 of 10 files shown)


core/src/evaluation/functions/aggregation/collect.rs
```diff
@@ -0,0 +1,585 @@
...
+                return Err(FunctionError {
+                    function_name: "Collect".to_string(),
+                    error: FunctionEvaluationError::CorruptData,
+                })
+            }
+        };
+
+        // For revert, we need to remove the value from the list
+        // This is tricky because we need to find and remove the exact value
+        // For now, we'll remove the first occurrence
```
ruokun-niu, line 111: What would happen if we have duplicate values? I am just curious about what is the expected behavior

amansinghoriginal: If there are duplicates, it will remove the first occurrence (the one with lowest index). This effectively treats duplicates with FIFO semantics.

amansinghoriginal: Both the OpenCypher and GQL standards state that COLLECT does not guarantee order. Aggregations into lists are generally considered to be from an unordered multiset.

In a streaming system like Drasi, "order" could means arrival order.

So if the collected list contains [A, B, A, C]) and a revert operation is triggered for A, the implementation will remove only the first occurrence of A it finds. The resulting list would be [B, A, C].

amansinghoriginal: In Drasi's terms, if we're collecting different entities (nodes) in the graph - they will anyways be unique because of Element ID.

ruokun-niu: Makes sense. Thanks for the detailed explanation

danielgerlag (approved):

ruokun-niu (approved):

PR #65, opened 2025-07-03
Implemented separate GQL functions and GROUP BY

ama2369:
# Pull Request: GQL GROUP BY Implementation and Function Registry

## Overview

This pull request introduces GROUP BY from Graph Query Language (GQL) and contains two different function registries for Cypher and GQL. The implementation ensures that queries without an explicit GROUP BY are grouped implicitly (like Cypher), while queries with an explicit GROUP BY behave as described in the standard.

## Function Registry: Cypher and GQL
The function registry now supports both Cypher and GQL, allowing both query languages to use the same underlying functions while supporting different names or aliases as needed. For example, Cypher uses `toUpper` and GQL uses `upper`, but both are mapped to the same implementation. The registry manages these naming differences for each language, ensuring consistency and reducing duplication.

### 1. Implicit Grouping (No GROUP BY clause)
- If no `GROUP BY` clause is specified, all non-aggregated expressions in the RETURN clause automatically become grouping keys. This follows Cypher's behavior. 

### 2. Explicit GROUP BY
- If a `GROUP BY` clause is present, only the listed expressions/columns are used as grouping keys.
- Any non-aggregated column in `RETURN` that is not in the `GROUP BY` clause is **invalid**.

### 3. Grouping Keys
- **Grouping keys must NOT be aggregate functions or their aliases.**
- You may use an alias from the `RETURN` clause as a grouping key if:
  - The alias refers to a non-aggregate expression.
  - The alias is unique among projected columns.
  - The alias is actually defined in the `RETURN` clause.

### 4. RETURN Clause Requirements
- Every projected column in the `RETURN` clause must either:
  - Appear in the `GROUP BY` clause, or
  - Be the result of an aggregate function.

### 6. Two-Stage Plan
- If `GROUP BY` specifies more keys than are projected, the engine uses a **two-stage plan**:
  - First, group by all keys.
  - Then, further project only the keys in the RETURN.

#### 1. Every grouping key listed in GROUP BY is present in the output (single-stage plan)
```cypher
RETURN z.type AS zone_type, v.color AS vehicle_color, count(v) AS vehicle_count
GROUP BY zone_type, vehicle_color
```
- `GROUP BY` keys exactly match non-aggregate columns in `RETURN`.

#### 2. Fewer Projected Keys Than GROUP BY Keys (two-stage plan)
```cypher
RETURN z.type AS zone_type, count(v) AS vehicle_count
GROUP BY zone_type, v.color
```
- Engine groups by both `zone_type` and `v.color`, then projects only `zone_type` and aggregates.

#### 3. More Projected Keys Than GROUP BY Keys (error)
```cypher
RETURN z.type AS zone_type, v.color AS vehicle_color, count(v) AS vehicle_count
GROUP BY vehicle_color
```
- `zone_type` is not in `GROUP BY` and is not an aggregate—**this is an error**.

## Implementation Details

The explicit GROUP BY logic is handled in the function handle_explicit_group_by. It works as follows:

1. Partition the RETURN expressions into aggregates (A) and non-aggregates (K).
2. Build a lookup set from for the expressions in GROUP BY (G).
3. For each expression in K:
    - If the expression is in G: the query is valid.
    - Else: Error (must be grouped or aggregated).
4. If all G are present in K: 
    - Single-stage plan, emit original RETURN clause.
5. Else: 
    - Two-stage plan:
        - Stage 1: GROUP BY the full set G and compute all A.
        - Stage 2: Project the original RETURN expressions.

## Two-stage plan
When the `GROUP BY` clause contains more keys than are present in the projected output (`RETURN`), the engine uses a **two-stage plan**. 

In stage 1, group by all keys specified in the `GROUP BY` clause. The result set at this stage includes all grouping keys and aggregate columns. The second stage projects only the columns requested in the original `RETURN` clause.


core/src/evaluation/functions/aggregation/mod.rs
```diff
@@ -37,12 +37,31 @@ use self::lazy_sorted_set::LazySortedSet;
 
 use super::{Function, FunctionRegistry};
 
-pub trait RegisterAggregationFunctions {
-    fn register_aggregation_functions(&self);
+pub trait RegisterCypherAggregationFunctions {
+    fn register_cypher_aggregation_functions(&self);
```
danielgerlag, line 41: We should keep the `core` crate clean of any specific language implementation. Perhaps we should make these traits common and put them in the `ast` crate, and then have implementations in each specific language crate.

core/src/query/query_builder.rs
```diff
@@ -15,12 +15,14 @@
 use std::{collections::HashMap, sync::Arc};
 
 use drasi_query_ast::api::QueryParser;
+use drasi_query_ast::ast::Query;
 use drasi_query_cypher::CypherParser;
+use drasi_query_gql::GQLParser;
```
danielgerlag, line 20: Let's remove the dependency on the 2 parsers crates from the core crate. I think the only reason the cypher one is there is for a default. We can remove the default and force the user to pass in an implementation of the `QueryParser` trait, rather than an enum. The core crate should not have knowledge of all the language implementations out there, they should be able to just plug-in without modifications to the core crate.

query-gql/src/lib.rs
```diff
@@ -0,0 +1,623 @@
...
+        }
+    }
+
+    // Non-aggregated columns in RETURN that are not in the GROUP BY clause are invalid and cause an error.
+    let invalid: Vec<_> = return_non_aggs.iter()
+        .filter(|e| !group_by_keys.iter().any(|k| expr_covers(e, k)))
+        .cloned()
+        .collect();
+    if !invalid.is_empty() {
+        panic!("Projection contains non-grouped columns");
```
danielgerlag: Instead of a panic, I think we should bubble up a `QueryParseError` to the callers of the parse function. A panic terminates the process.

query-gql/src/lib.rs
```diff
@@ -0,0 +1,623 @@
...
+    two_step_parts(
+        &match_clauses,
+        &where_clauses,
+        full_keys,
+        aggregates,
+        &return_expressions,
+    )
+}
+
+fn expr_covers(expr: &Expression, key: &Expression) -> bool {
```
danielgerlag: Let's rename this `fn` to something more descriptive.

query-gql/src/lib.rs
```diff
@@ -0,0 +1,623 @@
...
+         // If the RETURN expression is an alias (a.name AS person_name) and the GROUP BY references the alias name (person_name), it is a match.
+        (Expression::UnaryExpression(UnaryExpression::Alias { alias, .. }),
+         Expression::UnaryExpression(UnaryExpression::Identifier(id))) => {
+            id == alias
+        }
+        _ => false,
+    }
+}
+
+fn two_step_parts(
```
danielgerlag: Let's make the name of this fn more explicit about the group by functionality.

query-gql/src/lib.rs
```diff
@@ -0,0 +1,623 @@
...
+        return_clause: ProjectionClause::Item(
+            original_return.iter().map(strip_alias).collect()
+        ),
+    };
+
+    vec![stage1, stage2]
+}
+
+// If the expression is an alias (x + 1 AS sum), replace it with just the alias identifier.
+fn strip_alias(expr: &Expression) -> Expression {
```
danielgerlag: are we stripping the alias or preserving the alias? The name is a bit mid-leading.

PR #67, opened 2025-07-15
Implemented GQL GROUP BY

ama2369:
# Pull Request: GQL GROUP BY Implementation

## Overview

This pull request introduces GROUP BY from Graph Query Language (GQL). The implementation ensures that queries without an explicit GROUP BY are grouped implicitly (like Cypher), while queries with an explicit GROUP BY behave as described in the standard.

### 1. Implicit Grouping (No GROUP BY clause)
- If no `GROUP BY` clause is specified, all non-aggregated expressions in the RETURN clause automatically become grouping keys. This follows Cypher's behavior. 

### 2. Explicit GROUP BY
- If a `GROUP BY` clause is present, only the listed expressions/columns are used as grouping keys.
- Any non-aggregated column in `RETURN` that is not in the `GROUP BY` clause is **invalid**.

### 3. Grouping Keys
- **Grouping keys must NOT be aggregate functions or their aliases.**
- You may use an alias from the `RETURN` clause as a grouping key if:
  - The alias refers to a non-aggregate expression.
  - The alias is unique among projected columns.
  - The alias is actually defined in the `RETURN` clause.

### 4. RETURN Clause Requirements
- Every projected column in the `RETURN` clause must either:
  - Appear in the `GROUP BY` clause, or
  - Be the result of an aggregate function.

### 6. Two-Stage Plan
- If `GROUP BY` specifies more keys than are projected, the engine uses a **two-stage plan**:
  - First, group by all keys.
  - Then, further project only the keys in the RETURN.

#### 1. Every grouping key listed in GROUP BY is present in the output (single-stage plan)
```cypher
RETURN z.type AS zone_type, v.color AS vehicle_color, count(v) AS vehicle_count
GROUP BY zone_type, vehicle_color
```
- `GROUP BY` keys exactly match non-aggregate columns in `RETURN`.

#### 2. Fewer Projected Keys Than GROUP BY Keys (two-stage plan)
```cypher
RETURN z.type AS zone_type, count(v) AS vehicle_count
GROUP BY zone_type, v.color
```
- Engine groups by both `zone_type` and `v.color`, then projects only `zone_type` and aggregates.

#### 3. More Projected Keys Than GROUP BY Keys (error)
```cypher
RETURN z.type AS zone_type, v.color AS vehicle_color, count(v) AS vehicle_count
GROUP BY vehicle_color
```
- `zone_type` is not in `GROUP BY` and is not an aggregate—**this is an error**.

## Implementation Details

The explicit GROUP BY logic is handled in the function handle_explicit_group_by. It works as follows:

1. Partition the RETURN expressions into aggregates (A) and non-aggregates (K).
2. Build a lookup set from for the expressions in GROUP BY (G).
3. For each expression in K:
    - If the expression is in G: the query is valid.
    - Else: Error (must be grouped or aggregated).
4. If all G are present in K: 
    - Single-stage plan, emit original RETURN clause.
5. Else: 
    - Two-stage plan:
        - Stage 1: GROUP BY the full set G and compute all A.
        - Stage 2: Project the original RETURN expressions.

## Two-stage plan
When the `GROUP BY` clause contains more keys than are present in the projected output (`RETURN`), the engine uses a **two-stage plan**. 

In stage 1, group by all keys specified in the `GROUP BY` clause. The result set at this stage includes all grouping keys and aggregate columns. The second stage projects only the columns requested in the original `RETURN` clause.

## Caveats
Grouping without aggregation functions is not supported. This is similar to RETURN DISTINCT, which is also currently unsupported.

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -7,7 +7,6 @@ on:
     branches:
       - main
       - feature/*
-      - release/*
       - feature-gql
 
 jobs:
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -4,8 +4,11 @@ on:
   push:
     branches: ['*']
   pull_request:
-    branches: [ "main" ]
-  
+    branches:
+      - main
+      - feature/*
+      - release/*
+      - feature-gql
 
 jobs:
   clippy_check:
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -21,7 +21,6 @@ on:
   pull_request:
     branches:
       - main
-      - release/*
       - feature/*
       - feature-gql
 
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -21,6 +21,9 @@ on:
   pull_request:
     branches:
       - main
+      - release/*
+      - feature/*
+      - feature-gql
 
 env:
   CARGO_TERM_COLOR: always
--- a/query-gql/src/lib.rs
+++ b/query-gql/src/lib.rs
@@ -1,0 +1,626 @@
+#![allow(clippy::redundant_closure_call)]
+// Copyright 2024 The Drasi Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use drasi_query_ast::{
+    api::{QueryParseError, QueryParser},
+    ast::{
+        self, Expression, MatchClause, ParentExpression, ProjectionClause, QueryPart,
+        UnaryExpression,
+    },
+};
+use peg::{error::ParseError, str::LineCol};
+use std::{collections::HashSet, sync::Arc};
+
+#[cfg(test)]
+mod tests;
+
+peg::parser! {
+    grammar gql() for str {
+        use drasi_query_ast::ast::*;
+
+        rule kw_match()     = ("MATCH" / "match")
+        rule kw_optional()  = ("OPTIONAL" / "optional")
+        rule kw_where()     = ("WHERE" / "where")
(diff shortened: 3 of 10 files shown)


query-gql/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+            = r:return_clause() { r }
+
+        rule part(config: &dyn GQLConfiguration) -> Vec<QueryPart>
+            = match_clauses:( __* m:(match_clause() ** (__+) )? { m.unwrap_or_default().into_iter().flatten().collect() } )
+              where_clauses:( __* w:(where_clause() ** (__+) )? { w.unwrap_or_default() } )
+              __*
+              return_clause:(with_or_return())
+              group_by_exprs:( __* g:group_by_clause()? { g } )
+              __*
+              { build_query_parts(match_clauses, where_clauses, return_clause, group_by_exprs, config).unwrap() }
```
ruokun-niu: If `build_query_parts` return an error and we unwrap, I think it will result in a `panic`. I believe we should avoid using `unwrap()` here

query-gql/src/tests.rs
```diff
@@ -0,0 +1,516 @@
...
+struct TestConfig {}
+
+impl GQLConfiguration for TestConfig {
+    fn get_aggregating_function_names(&self) -> HashSet<String> {
+        let mut set = HashSet::new();
+        set.insert("count".into());
+        set.insert("sum".into());
+        set.insert("min".into());
+        set.insert("max".into());
+        set.insert("avg".into());
```
ruokun-niu: nit: Seems like these aggregating functions are not used?

danielgerlag (approved):

ruokun-niu:
@ama2369 You might want to up also update the Github Actions workflow to target your feature branch as well. Seems like most of the checks are not even triggered. Let me know you need any help.

For example, the test GH action is currently targetting only the `main` branch. https://github.com/ama2369/drasi-core/blob/feature-gql/.github/workflows/test.yml#L3-L7

.github/workflows/ci-lint.yml
```diff
@@ -4,8 +4,11 @@ on:
   push:
     branches: ['*']
   pull_request:
-    branches: [ "main" ]
-  
+    branches:
+      - main
+      - feature/*
+      - release/*
```
ruokun-niu: We dont need `release/*` here as we won't be doing releases in the drasi-core branch

.github/workflows/coverage.yaml
```diff
@@ -21,6 +21,9 @@ on:
   pull_request:
     branches:
       - main
+      - release/*
```
ruokun-niu: We dont need `release/*` here as we won't be doing releases in the drasi-core branch

ruokun-niu (approved):

PR #68, opened 2025-07-18
Implemented FILTER, LET, and YIELD keywords

ama2369:
# GQL FILTER, LET, and YIELD Implementation

## Overview
This pull request introduces support for `FILTER`, `LET`, and `YIELD` from Graph Query Language (GQL), enabling variable creation, scoped column selection/aliasing, and table filtering.

## LET Statement
The `LET` statement allows users to define new variables or computed fields for every row in the current working table. Each `LET` expression can reference all existing columns in scope, and the resulting variables are added as new columns. Aggregates are not allowed in `LET`.

**Implementation Details:** 
If `LET` directly follows a `MATCH` (optionally with `WHERE`), a single query part is formed that includes the match and where conditions, and the entire scope is projected along with the new variable. Otherwise, a `LET` is a single query part with an empty `MATCH` and `WHERE` that projects the new variables and current scope.

**Example:**
```cypher
MATCH (v:Vehicle)
LET isRed = v.color = 'Red'
RETURN v.color, isRed
```
_Equivalent Cypher:_
```cypher
MATCH (v:Vehicle)
WITH v, v.color = 'Red' AS isRed
RETURN v.color, isRed
```
## YIELD Clause
The `YIELD` clause projects and optionally renames columns from the working table, limiting the set of columns available to subsequent clauses. Only specified columns remain in scope after `YIELD`.

**Implementation Details:** 
If `YIELD` directly follows a `MATCH` (optionally with a `WHERE` clause), a single query part is formed that includes the match and where conditions, and the projection contains only the columns defined in `YIELD`. Otherwise, YIELD forms a separate query part with an empty MATCH and WHERE, projecting only the expressions defined in YIELD.

**Example:**
```cypher
MATCH (v:Vehicle)-[e:LOCATED_IN]->(z:Zone)
YIELD v.color AS vehicleColor, z.type AS location
RETURN vehicleColor, location
```
_Equivalent Cypher:_
```cypher
MATCH (v:Vehicle)-[e:LOCATED_IN]->(z:Zone)
WITH v.color AS vehicleColor, z.type AS location
RETURN vehicleColor, location
```

## FILTER Statement
`FILTER` is a standalone statement that filters the current working table after previous steps. It does not create a new table, and it updates the working table.

**Implementation Details:** 
If `FILTER` directly follows a `MATCH` (optionally with a `WHERE` clause), a single query part is formed that includes the match and where conditions and projects the current scope. Then, a separate query part is created where the `FILTER` condition is applied as a `WHERE` clause in the query part, and the current scope is projected. Regardless of its position, `FILTER` always produces a separate query part. 

`FILTER` is implemented by introducing a new query part where the filter expression is placed in the `WHERE` clause of the query part.

**Example:**
```cypher
MATCH (n:Person)
FILTER n.age > 30
RETURN n.name, n.age
```

### DIFF
--- a/query-gql/Cargo.toml
+++ b/query-gql/Cargo.toml
@@ -16,6 +16,3 @@ drasi-query-ast = { path = "../query-ast" }
 peg = "0.8.1"
 serde = "1.0.164"
 serde_json = "1.0.96"
-
-[dev-dependencies]
-drasi-query-cypher = { path = "../query-cypher" }
--- a/query-gql/Cargo.toml
+++ b/query-gql/Cargo.toml
@@ -16,3 +16,6 @@ drasi-query-ast = { path = "../query-ast" }
 peg = "0.8.1"
 serde = "1.0.164"
 serde_json = "1.0.96"
+
+[dev-dependencies]
+drasi-query-cypher = { path = "../query-cypher" }
--- a/query-gql/src/lib.rs
+++ b/query-gql/src/lib.rs
@@ -402,22 +402,24 @@ peg::parser! {
             = kw_group() __+ kw_by() __+ "(" __* ")" { Vec::new() }  // Handle GROUP BY ()
             / kw_group() __+ kw_by() __+ items:( expression() ++ (__* "," __*) ) { items }
 
-        rule with_or_return() -> Vec<Expression>
-            = r:return_clause() { r }
-
+        rule match_with_where() -> (Vec<MatchClause>, Vec<Expression>)
+            = ms:(match_clause() ++ (__+)) __*
+            w:where_clause()? { (ms.into_iter().flatten().collect(), w.into_iter().collect()) }
+        
         rule part(config: &dyn GQLConfiguration) -> Vec<QueryPart>
-            = match_clauses:( __* m:(match_clause() ** (__+) )? { m.unwrap_or_default().into_iter().flatten().collect() } )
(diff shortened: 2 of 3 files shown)


query-gql/Cargo.toml
```diff
@@ -16,3 +16,6 @@ drasi-query-ast = { path = "../query-ast" }
 peg = "0.8.1"
 serde = "1.0.164"
 serde_json = "1.0.96"
+
+[dev-dependencies]
+drasi-query-cypher = { path = "../query-cypher" }
```
danielgerlag: Why would the GQL crate depend on the Cypher crate?

query-gql/src/lib.rs
```diff
@@ -401,6 +433,12 @@ peg::parser! {
 
 }
 
+pub enum StatementClause {
```
danielgerlag: Is this internal to the GQL crate?

query-gql/src/tests.rs
```diff
@@ -16,6 +16,7 @@ use std::collections::HashSet;
 
 use super::*;
 use ast::*;
+use drasi_query_cypher::{parse, CypherConfiguration};
```
danielgerlag: Are we able to create a common configuration trait for both languages?

ruokun-niu (approved):

danielgerlag (approved):

PR #69, opened 2025-07-28
Separate Function Registry Implementation for Multiple Query Language Support

ama2369:
# Function Registry Implementation

## Summary

This PR introduces a function registry system that decouples the core crate from specific query languages. While function implementations remain in the core crate for reusability, their registration is now handled by language-specific crates. This architecture enables Drasi to better support multiple query languages.

### New Architecture
- **Function Registry**: A centralized registry for managing query language functions. The registry starts empty and functions are registered programmatically by language-specific crates or user code
- **Language-Agnostic Core**: Core now depends on `query-ast` instead of language-specific parsers, removing direct Cypher dependencies
- **Function Crates**: New `functions-cypher` and `functions-gql` crates that register language-specific functions and naming
- **QueryConfiguration Trait**: Enables language-agnostic configuration by providing a common interface for parsers to query available functions

### Dependency Changes
- Core no longer directly registers functions - registration is delegated to `functions-cypher` and `functions-gql` crates. Custom function sets can be created by directly registering functions on the `Arc<FunctionRegistry>`
- Function crates only depend on core, creating a clean dependency hierarchy
- Parser crates depend on the AST crate

### API Updates
QueryBuilder now requires an explicit parser with function registry:

```rust
// Old API
let query_builder = QueryBuilder::new(query_str);
let query = query_builder.build().await;

// New API  
let function_registry = Arc::new(FunctionRegistry::new()).with_cypher_function_set();
let parser = Arc::new(CypherParser::new(function_registry.clone()));
let query_builder = QueryBuilder::new(query_str, parser)
    .with_function_registry(function_registry);
let query = query_builder.build().await;
```
This new API explicitly registers a Cypher continuous query with all available Cypher functions. 

### QueryParser Trait
The `QueryParser` trait from `query-ast` provides a language-agnostic interface for parsing queries. Each language implementation (CypherParser, GQLParser) requires a function registry to properly identify aggregating functions during parsing.

### Migration Impact
- All tests and examples updated to use the new API pattern
- The `with_cypher_function_set()` extension method registers all standard Cypher functions
- Similar extension methods available: `with_gql_function_set()` for GQL functions
- Users can provide custom functions by registering directly on `Arc<FunctionRegistry>`


core/src/evaluation/functions/list.rs
```diff
@@ -16,22 +16,6 @@ mod range;
 mod reduce;
 mod tail;
```
danielgerlag, line 17: Do we still need this file / all these similar files?

core/src/evaluation/functions/drasi.rs
```diff
@@ -20,19 +20,4 @@ mod tests;
 
 pub use max::DrasiMax;
```
danielgerlag, line 21: Is there anything left in this file?

danielgerlag (approved):
LGTM, I think there are some hollowed out files that don't really contain anything anymore that we could just remove.

examples/vehicles.rs
```diff
@@ -17,22 +17,29 @@ mod process_monitor;
...
+
 #[allow(clippy::print_stdout, clippy::unwrap_used)]
 #[tokio::main]
 async fn main() {
     let query_str = "
     MATCH 
         (v:Vehicle)-[:LOCATED_IN]->(:Zone {type:'Parking Lot'}) 
     RETURN 
-        v.color AS color, 
+        toUpper(v.color) AS color,
```
ruokun-niu: Why are we adding `toUpper` in this test? Just curious

PR #70, opened 2025-07-31
Add GQL CAST Function Support

ama2369:
## Summary

This PR implements the `CAST` function for GQL, enabling type conversion operations in GQL queries. The implementation follows the GQL standard for scalar type casting and integrates with Drasi's existing type conversion infrastructure.

### Core Implementation
- Added `CAST` function in `functions-gql/src/gql_scalar/cast.rs`
- Supports casting to:
  - `BOOLEAN` - Converts values to boolean type
  - `INTEGER` - Converts values to integer type  
  - `FLOAT` - Converts values to floating-point type
  - `STRING` - Converts values to string type
- Leverages existing Drasi type conversion traits (`ToBoolean`, `ToInteger`, `ToFloat`, `ToString`)

### Testing
- Tests cover all supported type conversions and edge cases
- Validates proper error handling for invalid casts

### Integration
- Registered CAST function in the GQL function registry
- Updated module exports in `functions-gql/src/lib.rs` and `query-gql/src/lib.rs`

## Usage Example

```gql
CAST(42 AS STRING)
CAST(3.14 AS FLOAT)
CAST(1 AS BOOLEAN)
CAST(100 AS INTEGER)
```

## Technical Details

- The function accepts two arguments: the value to cast and the target type name
- Target type names are case-insensitive ("INTEGER", "integer", "Integer" all work)
- Returns appropriate errors for invalid argument counts or unsupported type names

## Limitations

- **Temporal type conversions are not supported**: The current implementation does not support casting to or from date, time, datetime, or duration types 
- Only basic scalar type conversions (BOOLEAN, INTEGER, FLOAT, STRING) are implemented in this PR

### DIFF
--- a/functions-gql/src/gql_scalar/cast.rs
+++ b/functions-gql/src/gql_scalar/cast.rs
@@ -1,0 +1,83 @@
+// Copyright 2024 The Drasi Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use drasi_core::evaluation::functions::ast;
+use drasi_core::evaluation::functions::async_trait;
+use drasi_core::evaluation::functions::{ScalarFunction, ToBoolean, ToFloat, ToInteger, ToString};
+use drasi_core::evaluation::variable_value::VariableValue;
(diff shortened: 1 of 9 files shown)


functions-gql/src/gql_scalar/cast.rs
```diff
@@ -0,0 +1,83 @@
...
+                return Err(FunctionError {
+                    function_name: expression.name.to_string(),
+                    error: FunctionEvaluationError::InvalidArgument(1),
+                })
+            }
+        };
+
+        match target_type.as_str() {
+            "STRING" => {
+                let to_string = ToString {};
```
ruokun-niu, line 52: I know that we suppor both ToString and ToStringOrNull in cypher. What is the expected behavior for cast() in gql?

AzureMike: It aligns with ToString. For example, when casting booleans to strings, ToString correctly returns the string representation, while ToStringOrNull returns an error.

ruokun-niu (approved):

danielgerlag (approved):

PR #71, opened 2025-08-05
GQL NEXT Keyword Implementation

ama2369:
## Overview
This pull request introduces support for the `NEXT` keyword from Graph Query Language (GQL), enabling linear composition of multiple query statements where each subsequent statement receives the working table from the previous statement.

## NEXT Statement
The `NEXT` statement allows users to chain multiple query statements in a pipeline. Each statement after `NEXT` operates on the output of the previous statement, with variables in scope carried forward to subsequent statements.

**Implementation Details:**
When `NEXT` is encountered, the parser captures the return projection from the previous statement and extracts the available scope using `get_scope_from_projection()`. This scope becomes the incoming working record for the next statement. Each `NEXT` segment can contain `FILTER`, `LET`, `YIELD` clauses followed by a `RETURN` with optional `GROUP BY`.

**Example:**
```gql
MATCH (v:Vehicle) 
RETURN v 
NEXT FILTER v.color = 'Red' 
RETURN v.color
```
1. `MATCH (v:Vehicle) RETURN v` - Initial query
2. `FILTER v.color = 'Red' RETURN v.color` - Operates on results from part 1

**Complex Example:**
```gql
RETURN v, z
NEXT YIELD v AS vehicle, z AS zone
LET loc_type = zone.type
RETURN vehicle.color, loc_type
```

### Post-Aggregation Filtering
In this PR, one of the most useful features enabled by `NEXT` is the ability to filter results after aggregation. This allows for SQL HAVING like functionality in GQL.

**Example:**
```gql
MATCH (v:Vehicle)
RETURN v.color AS color, count(v) AS vehicle_count
GROUP BY color
NEXT FILTER vehicle_count > 5
RETURN color, vehicle_count
```
_This filters to only show colors that have more than 5 vehicles, filtering AFTER the aggregation is complete._

## Scope Management Enhancements
This PR also improves scope validation for `YIELD` statements to comply with GQL specification requirements.

**YIELD Scope Validation:**
`YIELD` now properly validates that unaliased identifiers exist in the current scope. Complex expressions without aliases are rejected with appropriate error messages.

**New Error Types:**

1. **`UnaliasedComplexExpression`**: Thrown when complex expressions lack required aliases
   ```gql
   MATCH (v:Vehicle)
   YIELD v.color              // Error: "Complex expression must have an alias"
   
   // Correct usage:
   YIELD v.color AS color     // Valid
   ```
   
   This is similar to Cypher's restriction where you cannot write:
   ```cypher
   MATCH (v:Vehicle)
   WITH v.color               // Invalid in Cypher - requires alias
   RETURN v.color
   ```

2. **`IdentifierNotInScope`**: Thrown when referencing undefined identifiers
   ```gql
   MATCH (v:Vehicle)
   YIELD nonexistent          // Error: "Identifier 'nonexistent' not found in current scope"
   
   // Another example:
   MATCH (v:Vehicle) RETURN v.color AS color
   NEXT YIELD v               // Error: "Identifier 'v' not found in current scope"
   ```

## Future Enhancement: MATCH after NEXT
Currently, `MATCH` clauses cannot appear after `NEXT` due to query engine limitations. However, the parser architecture is designed to easily support this when the engine is ready.


ama2369:
While the query execution engine would catch these new errors (by throwing an UnknownIdentifier error), is early detection at parse time preferable to runtime detection?

query-gql/src/lib.rs
```diff
@@ -448,11 +486,63 @@ pub fn parse_expression(input: &str) -> Result<ast::Expression, ParseError<LineC
...
+        QueryParseError::MissingGroupByKey => {
+            "Non-grouped RETURN expressions must appear in GROUP BY clause"
+        }
+        QueryParseError::ParserError(_) => "Parser error",
+        QueryParseError::UnaliasedComplexExpression => "Complex expression must have an alias",
+        QueryParseError::IdentifierNotInScope => "Identifier not found in current scope",
+    }
+}
+
+fn get_scope_from_projection(
```
AzureMike, line 489: Is there a better way to avoid duplicating logic for the Item vs GroupBy ProjectionClause?

query-gql/src/lib.rs
```diff
@@ -448,11 +486,63 @@ pub fn parse_expression(input: &str) -> Result<ast::Expression, ParseError<LineC
     gql::expression(input)
 }
 
+fn map_query_parse_error(e: QueryParseError) -> &'static str {
```
danielgerlag: Could this be done by implementing the `Display` trait?

query-gql/src/lib.rs
```diff
@@ -528,12 +621,17 @@ fn build_parts_for_statements(
...
-                    } else {
-                        projection.push(expr.clone());
-                        if let Expression::UnaryExpression(UnaryExpression::Identifier(id)) = &expr
-                        {
-                            new_scope.push(UnaryExpression::ident(id));
+                    } else if let Expression::UnaryExpression(UnaryExpression::Identifier(id)) =
+                        &expr
+                    {
+                        if !scope_contains_identifier(&scope, id) {
+                            return Err(QueryParseError::IdentifierNotInScope);
```
danielgerlag: Should the error include details of the identifier that is not in scope, so that the user has more actionable feedback?

query-gql/src/lib.rs
AzureMike: The map_query_parse_error function returns &'static str because it's used within rust-peg's {? ... } conditional action blocks, which require Result<T, &str> where the error must be a string slice (https://docs.rs/peg/latest/peg/#combining). This peg constraint prevent using error types with Display implementations. Let me know if there's a better way.

danielgerlag: Implementing the `Display` trait will automatically implement the `ToString` trait. Could you use `.to_string()` and get the formatted message from within the error implementation?

AzureMike: Limitation added in the PEG Error Handling Caveats section of the description

danielgerlag (approved):

PR #74, opened 2025-08-14
GQL Temporal Function Naming Update

ama2369:
## Summary
This PR properly maps the temporal functions available in GQL according to the standard to our implementations

### DIFF
--- a/functions-gql/src/lib.rs
+++ b/functions-gql/src/lib.rs
@@ -41,7 +41,6 @@ pub fn register_default_gql_functions(registry: &FunctionRegistry) {
     register_list_functions(registry);
     register_metadata_functions(registry);
     register_drasi_functions(registry);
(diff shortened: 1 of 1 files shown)


ruokun-niu (approved):

functions-gql/src/lib.rs
```diff
@@ -94,13 +93,6 @@ fn register_drasi_functions(registry: &FunctionRegistry) {
     registry.register_function("drasi.stdevp", Function::Scalar(Arc::new(DrasiStdevP {})));
 }
 
-fn register_context_mutators(registry: &FunctionRegistry) {
```
danielgerlag: This was a drasi specific function

AzureMike: I've added it back

ruokun-niu (approved):

PR #79, opened 2025-10-09
JQ Middleware implementation

danielgerlag:
This pull request introduces a new JQ-based middleware for processing source changes using JQ queries, adds a utility conversion for property maps, and updates dependencies to support the new functionality. The main theme is enabling flexible transformation and mapping of elements via configurable JQ expressions.

https://jqlang.org/manual/

**JQ Middleware Implementation:**

* Added a new `middleware/src/jq/mod.rs` module implementing a `SourceMiddleware` that uses JQ queries to transform and map incoming `SourceChange` objects. This includes support for configurable operations (insert, update, delete), element types (node, relation), and error handling. The middleware is registered via a factory class.

**Utility and Conversion Improvements:**

* Added an `impl From<&serde_json::Value> for ElementPropertyMap` to allow conversion from a reference to a JSON value into an `ElementPropertyMap`, improving flexibility when handling JSON data.

**Dependency Updates:**

* Added the `jq-rs` crate to `middleware/Cargo.toml` to enable JQ query processing within the middleware.

### DIFF
--- a/middleware/src/relabel/README.md
+++ b/middleware/src/relabel/README.md
@@ -21,29 +21,25 @@ The **relabel** middleware processes `SourceChange` events (`Insert`, `Update`, 
 4. **Update Element Metadata**  
    Create a new element with the transformed labels while preserving all other element properties, metadata, and relationships.
 
-5. **Error Handling**  
-   Configuration and processing errors are handled according to the `on_error` setting.
+5. **Error Handling**
+   Invalid configurations (empty `labelMappings` or malformed JSON) will cause middleware creation to fail.
 
 ## Configuration Options
 
 | Field          | Type                          | Required | Default  | Description                                                                                      |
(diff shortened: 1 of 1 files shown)


middleware/src/jq/mod.rs
```diff
@@ -0,0 +1,400 @@
...
+    op: Option<MapOperation>,
+    label: Option<String>,
+    id: Option<String>,
+    element_type: Option<MapElementType>,
+
+    #[serde(default)]
+    query: String,
+
+    #[serde(default)]
+    halt_on_error: bool,
```
ruokun-niu, line 45: Is this value defined by the `on_error` field in the middleware config?

danielgerlag: There is no `on_error` field

ruokun-niu: Oh my bad, I mixed up two PRs. I was thinking about a different one.

PR #80, opened 2025-10-16
New Functions: previousValue / previousDistinctValue

danielgerlag:
### New Functions

* Added two new scalar functions: `drasi.previousValue` and `drasi.previousDistinctValue`, with corresponding implementations in `core/src/evaluation/functions/future/before.rs` and `core/src/evaluation/functions/future/previous_unique_value.rs`. These functions enable tracking and querying previous expression values from previous changes, useful for comparing what changed. [[1]](diffhunk://#diff-a21089954846a509ab63758253abff72706edc1072e18204c8f6397394bad8fcR1-R184) [[2]](diffhunk://#diff-e5b85a684956ebf5098d58a6b7feb246acbc3f4658fb90cfe897a99ca3d60ab6R1-R187) [[3]](diffhunk://#diff-caae83c25162c4b796a47abf7043fff0924b36d0225266414ed114c0c0049a7bR27-R29) [[4]](diffhunk://#diff-caae83c25162c4b796a47abf7043fff0924b36d0225266414ed114c0c0049a7bR89-R104)

#### Examples

If we are only interested in changes that caused the invoice amount to increase.
```cypher
MATCH 
  (i:Invoice)
WHERE i.amount > drasi.previousValue(i.amount)
RETURN
  i.id AS id,
  i.amount AS amount
```

If we are only interested in contracts in the active state, where the previous state was pending, but we also want other non-state related changes for contracts that are already in the result set.
```cypher
MATCH 
  (c:Contract WHERE c.status = 'active' AND drasi.previousDistinctValue(c.status) = 'pending')
RETURN
  c.id AS id,
  c.status AS status,
  c.tags AS tags
```

### Expression Evaluation Logic

* Modified the evaluation of logical AND and OR expressions to remove short-circuiting, ensuring both operands are always evaluated. This change is verified by new tests that check for non-short-circuit behavior using divide-by-zero errors. [[1]](diffhunk://#diff-26452b7180df22415aa6a7c5411c1c23c035f541b61895d8ca8a6f0c51928385L489-R493) [[2]](diffhunk://#diff-b2b233d21dbc0246a7d503cdb6a93b85da40e61bb579c5ee1d2e788496fa42c9R196-R271)

### Value Accumulation and Conversion

* Extended the `ValueAccumulator` enum to support accumulation of property maps, enabling more complex value tracking for aggregation and future functions.
* Added a `TryInto<ElementValue>` implementation for owned `VariableValue`, improving type conversion ergonomics and reliability when handling variable values in function calls.
* Updated imports to include `ElementPropertyMap` where necessary for the new accumulator variant.

### DIFF
--- a/core/src/evaluation/functions/future/previous_value.rs
+++ b/core/src/evaluation/functions/future/previous_value.rs
@@ -30,12 +30,12 @@ use crate::models::ElementValue;
 use async_trait::async_trait;
 use drasi_query_ast::ast;
 
-pub struct Before {
+pub struct PreviousValue {
     result_index: Arc<dyn ResultIndex>,
     expression_evaluator: Weak<ExpressionEvaluator>,
 }
 
-impl Before {
+impl PreviousValue {
     pub fn new(
         result_index: Arc<dyn ResultIndex>,
         expression_evaluator: Weak<ExpressionEvaluator>,
@@ -48,7 +48,7 @@ impl Before {
 }
 
 #[async_trait]
-impl ScalarFunction for Before {
+impl ScalarFunction for PreviousValue {
     async fn call(
         &self,
         context: &ExpressionEvaluationContext,
--- a/core/src/evaluation/functions/future/mod.rs
+++ b/core/src/evaluation/functions/future/mod.rs
@@ -24,9 +24,9 @@ use self::{future_element::FutureElement, true_for::TrueFor, true_until::TrueUnt
 use super::{Function, FunctionRegistry};
 
 mod awaiting;
-mod before;
 mod future_element;
 mod previous_distinct_value;
+mod previous_value;
 mod true_for;
 mod true_later;
 mod true_now_or_later;
@@ -88,8 +88,8 @@ impl RegisterFutureFunctions for FunctionRegistry {
         );
 
         self.register_function(
-            "drasi.beforeChange",
-            Function::Scalar(Arc::new(before::Before::new(
+            "drasi.previousValue",
+            Function::Scalar(Arc::new(previous_value::PreviousValue::new(
                 result_index.clone(),
                 expression_evaluator.clone(),
             ))),
--- a/shared-tests/src/use_cases/before/queries.rs
+++ b/shared-tests/src/use_cases/before/queries.rs
@@ -16,7 +16,7 @@ pub fn increasing_value_query() -> &'static str {
     "
 MATCH 
   (i:Invoice)
-WHERE i.amount > drasi.beforeChange(i.amount)
+WHERE i.amount > drasi.previousValue(i.amount)
 RETURN
   i.id AS id,
   i.amount AS amount
@@ -30,7 +30,7 @@ MATCH
 WITH
   i,
   sum(l.amount) AS total
-WHERE total > drasi.beforeChange(total, 0)
+WHERE total > drasi.previousValue(total, 0)
 RETURN
   i.id AS id
   "


core/src/evaluation/expressions/tests/logical.rs
```diff
@@ -193,3 +193,79 @@ async fn evaluate_or() {
         );
     }
 }
+
+#[tokio::test]
+async fn and_not_short_circuited() {
+    // Tests that AND evaluates both operands even when the first is false.
```
ruokun-niu, line 199: Good catch!

core/src/evaluation/functions/future/previous_value.rs
```diff
@@ -0,0 +1,184 @@
...
+            return Err(FunctionError {
+                function_name: expression.name.to_string(),
+                error: FunctionEvaluationError::InvalidArgumentCount,
+            });
+        }
+
+        let result_owner = ResultOwner::Function(expression.position_in_query);
+        let value = &args[0];
+        let default_value: ElementValue = {
+            if args.len() > 1 {
```
ruokun-niu, line 68: nit: maybe we should throw an error if the args.len() is an invalid number

core/src/evaluation/functions/future/previous_distinct_value.rs
```diff
@@ -0,0 +1,187 @@
...
+                expected: "ElementValue".to_string(),
+            },
+        })?;
+
+        let (mut prev_unique, before_value) =
+            match self.result_index.get(&result_key, &result_owner).await {
+                Ok(v) => match v {
+                    Some(v) => match v {
+                        ValueAccumulator::Map(m) => (
+                            m.get("0").cloned().unwrap_or_default(),
```
amansinghoriginal, line 131: **NIT**: I was comparing the state management here with `previousValue` and had a thought on the slot assignments.

In `previousValue`, slot "0" holds the T-1 value and slot "1" holds the T-2 value.

Here in `previousDistinctValue`, slot "0" holds the last distinct value, while slot "1" holds the T-1 value.

For the sake of consistency across these stateful functions, would it make sense to swap the roles here? We could store the T-1 value in slot "0" and the "last distinct value" in slot "1".

This would establish a convention where slot "0" is always the "immediately previous value".

A minor alignment for conceptual consistency?

core/src/evaluation/expressions/mod.rs
```diff
@@ -486,11 +486,11 @@ impl ExpressionEvaluator {
         let result = match expression {
             ast::BinaryExpression::And(c1, c2) => VariableValue::Bool(
                 self.evaluate_predicate(context, c1).await?
-                    && self.evaluate_predicate(context, c2).await?,
+                    & self.evaluate_predicate(context, c2).await?,
```
amansinghoriginal, line 489: By removing short-circuiting, won't we encounter issues with queries that have something like:

```
WHERE x != 0 AND y / x > 10
```

Would we get a null pointer in situations like this:

```
WHERE x IS NOT NULL AND x.property > 10
```

core/src/evaluation/expressions/mod.rs
```diff
@@ -486,11 +486,11 @@ impl ExpressionEvaluator {
...
             ),
             ast::BinaryExpression::Or(c1, c2) => VariableValue::Bool(
                 self.evaluate_predicate(context, c1).await?
-                    || self.evaluate_predicate(context, c2).await?,
+                    | self.evaluate_predicate(context, c2).await?,
```
amansinghoriginal, line 493: Does removal of short-circuiting lead to surprise for Drasi users?  (https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

Instead of this:

```
MATCH (c:Contract)
WHERE c.status = 'active' AND drasi.previousDistinctValue(c.status) = 'pending'
```

Can the users not write something like:

```
MATCH (c:Contract)
WITH c, drasi.previousDistinctValue(c.status) AS prevStatus
WHERE c.status = 'active' AND prevStatus = 'pending'
```

So using `WITH` clause they can still use this function?

core/src/evaluation/expressions/mod.rs
danielgerlag: The reference Neo4J implementation of Cypher logical AND / OR operators are not short-circuiting, so we were out of alignment with standard cypher. The bigger issue is that we use side effects in some of our functions, and short-circuiting logic can break their functionality depending on how the query is written.

eg.

```
WHERE x = 1 AND drasi.trueFor(....)
```

Would not invoke the trueFor function which is needed to enqueue on the future queue.

For the case of this PR, these 2 queries would produce different results with short circuiting and it would not be obvious to the query writer.

```
WHERE c.status = 'active'
AND drasi.previousDistinctValue(c.status) = 'pending'
```

```
WHERE drasi.previousDistinctValue(c.status) = 'pending'
AND c.status = 'active'
```


Even the SQL standard does not guarantee short-circuiting for logical operators, so query writers should not rely on this behaviour.

core/src/evaluation/expressions/mod.rs
danielgerlag: The Cypher specification does not enforce short circuiting for logical operators, so users would not expect it.

danielgerlag: I think users are more likely to get tripped up trying to write queries without the "WITH" statement and getting confused as to why it does not work because they would not have knowledge of the inner workings of the side effect functions.

amansinghoriginal (approved):
`code-cov` is failing, so might want to increase test coverage if possible. Can do it later as separate PR as well.
Also consider swapping the slots, if it makes sense.

Otherwise, LGTM!

PR #90, opened 2025-12-11
Fix non-events for DrasiLib

danielgerlag:
- Drasi core assumes the `effective_from` field in a source metadata is in milliseconds from the unix epoch. This PR adjusts the Lib sources to populate this field in ms instead of ns.
- The future queue consumer needed to be configured when a continuous query starts, which id basically a callback that defines what happens when an item becomes due on the future queue, in this case, add it to the priority queues that has the source changes for the query.

### DIFF
--- a/lib/src/sources/future_queue_source.rs
+++ b/lib/src/sources/future_queue_source.rs
@@ -276,7 +276,10 @@ impl FutureQueueSource {
                 // Preserve the original error and add context
                 Box::new(std::io::Error::new(
                     std::io::ErrorKind::Other,
-                    format!("Failed to create receiver for future queue subscription: {}", e)
+                    format!(
+                        "Failed to create receiver for future queue subscription: {}",
+                        e
+                    ),
                 ))
             },
         )?;


lib/src/queries/manager.rs
```diff
@@ -457,6 +470,83 @@ impl Query for DrasiQuery {
...
+        let mut receiver = dispatcher
+            .create_receiver()
+            .await
+            .map_err(|e| anyhow::anyhow!("Failed to create receiver for FutureQueueSource: {e}"))?;
+
+        if let Err(e) = future_queue_source.start(Box::new(dispatcher)).await {
+            error!(
+                "Query '{}' failed to start FutureQueueSource: {}",
+                self.base.config.id, e
+            );
```
agentofreality: clippy will complain about not using the {var} syntax.

lib/src/queries/manager.rs
```diff
@@ -457,6 +470,83 @@ impl Query for DrasiQuery {
...
+            return Err(anyhow::anyhow!("Failed to start FutureQueueSource: {e}"));
+        }
+
+        info!(
+            "Query '{}' FutureQueueSource started successfully",
+            self.base.config.id
+        );
+
+        // Spawn task to forward future queue events to priority queue
+        let priority_queue = self.priority_queue.clone();
```
agentofreality: I had imagined the mechanics of the future queue source would be encapsulated inside the future queue, not implemented in the manager.

agentofreality: Things needing to go to the future queue get pushed into the future queue source, and then at some point later they 'magially' appear in the p-queue for processing.

agentofreality: by that I mean why dont queries that need future queue subscribe to the future queue source?

danielgerlag: refactored accordingly

lib/Cargo.toml
```diff
@@ -64,4 +64,16 @@ test-case = "3.3"
...
+print_stdout = "warn"
+unwrap_used = "warn"
+module_inception = "allow"
+ptr_arg = "allow"
+type_complexity = "allow"
+uninlined_format_args = "allow"
+
+[lints.rust]
+unused = "allow"
+warnings = "deny"
```
agentofreality, line 79: @danielgerlag dont these effectively turn off these clippy checks/warnings for drasi-core?

We already have the ones we want disabled in the workflow builds. others should be handled individually.

I definitely think we should not be uniformly disabling:
- printing to stdout
- use of unwrap
- uninlined args (this stops a lot of changes due to cargo fmt)
- unused code

danielgerlag: That is not what these are doing. These are another way of expressing the exact same flags we have with the CI workflow, but now you can run it for an individual crate without remembering all long RUST_FLAGS string.

danielgerlag:
```
warnings = "deny"
	•	Rust lint group: warnings
	•	This is important:
	•	warnings is a lint group that includes all rustc warnings.
	Effect:
	•	Any rustc warning becomes a hard error.
	•	Compilation fails if any rustc warning is emitted.
```

components/sources/postgres/src/stream.rs
```diff
@@ -648,7 +648,7 @@ impl ReplicationStream {
             metadata: ElementMetadata {
                 reference: ElementReference::new(&self.source_id, &element_id),
                 labels: Arc::from([Arc::from(mapping.label.as_str())]),
-                effective_from: Utc::now().timestamp_nanos_opt().unwrap_or(0) as u64,
+                effective_from: Utc::now().timestamp_millis() as u64,
```
amansinghoriginal, line 651: Should we do this instead?
`Utc::now().timestamp_millis().try_into().unwrap_or(0)`

components/sources/application/src/time.rs
amansinghoriginal, line 1: Is this `time.rs` file identical to the ones in `http` and `mock` sources? Should we create a shared utility?

lib/src/sources/future_queue_source.rs
```diff
@@ -0,0 +1,552 @@
...
+                    Ok(None) => {
+                        // No items in queue, sleep for a bit and check again
+                        sleep(Duration::from_millis(100)).await;
+                        continue;
+                    }
+                    Err(e) => {
+                        error!(
+                            "FutureQueueSource failed to peek due time for query '{query_id}': {e}"
+                        );
+                        sleep(Duration::from_secs(1)).await;
```
amansinghoriginal, line 118: Should the `1 second` be a declared constant, so that we can document why it was chosen as that value in a single place?

amansinghoriginal (approved):
Approved with suggestions.

Issue #136, opened 2026-01-06
Add MS SQL Support to drasi-lib


agentofreality:
### Overview of feature request

Create drasi-lib Source and Bootstrap Provider components for MS SQL Server.

### Acceptance criteria

A Bootstrap Provider plugin that fetches data from MS SQL Server
A Source plugin that handles updates from MS SQL Server

Unit tests
Integration tests

User documentation / configuration guide.

Once complete, create issues to add support to Drasi Server for new source and bootstrap provider

### Would you like to support us?

- [ ] Yes, I would like to support you

PR #141, opened 2026-01-07
MS SQL Server Source

danielgerlag:
This pull request adds full support for Microsoft SQL Server (MSSQL) as both a Change Data Capture (CDC) source and as a bootstrap provider for the Drasi platform. It introduces new plugins, configuration, documentation, and connection management for MSSQL, enabling Drasi to ingest and bootstrap data from SQL Server databases.

Out of scope for this PR is managed identity authentication.

Resolves : #136


agentofreality:
@danielgerlag please look and see if you can use testcontainers trait to do tests against an active SQL Server. Look at the redis testing and Ruokun's reactions for examples of doing this.

Also, I am thinking it might be better to put examples specific to a source/reaction in the component's folder instead of in core/examples. Then the entire component is together.

PR #148, opened 2026-01-15
Source agent

danielgerlag:
This pull request introduces a comprehensive agent specification for the new `source-planner` role in the `.github/agents/source-planner.md` file. The agent is responsible for creating detailed, actionable implementation plans for new data source and bootstrap components in Drasi, with a strong emphasis on technical verification, evidence-based planning, and robust testing requirements.

For local usage with the Copilot CLI:

- Start copilot CLI
- Select the `source-planner` agent using `/agent`
- Select the `claude-sonnet-4.5` model using `/model` (CLI does not use the model defined in the frontmatter)
- Run a prompt like `Please write a plan for an XXXX source and save it to file`
- Once the plan is complete, review it and tweak it in the generated file
- Switch to the `gpt-5.2-codex` model using `/model`
- Switch to the `source-plan-executor` agent using `/agent`
- Run the prompt: `please implement the plan`

### DIFF
--- a/.github/agents/source-plan-executor.md
+++ b/.github/agents/source-plan-executor.md
@@ -767,6 +767,7 @@ Implementation is complete when ALL are true:
 - [ ] Code formatted: `cargo fmt`
 - [ ] Documentation complete (source, bootstrap, example READMEs)
 - [ ] Makefile targets tested and verified working
+- [ ] doctests RUN and PASS
 
 **If ANY checkbox is unchecked, implementation is INCOMPLETE.**
 


.github/agents/source-planner.md
```diff
@@ -0,0 +1,261 @@
...
+- Private fields that block access to needed data
+- Missing event types or change detection mechanisms
+- API limitations that require workarounds
+
+### 4. Determine Data Mapping Strategies
+
+- Decide how to map source data to Drasi's graph data model
+- If multiple strategies are possible, outline pros/cons of each
+- Consider data types, structures, and necessary transformations
+
```
amansinghoriginal, line 70:
```suggestion
- For event-stream based sources, default to Entity-Based mapping whenever possible along with  `SourceChange::Update` which has Upsert semantics.
    - Entity-based mapping means that every incoming event is treated as a state change to a persistent entity.
    - Because Update has upsert semantics, source can be stateless
    - This prevents unbounded graph growth and ensures correct query semantics
    - Decsision Tree:
        - Can you derive a stable Entity ID (pk, user_id, device_id) from the events?
        - YES -> Use Entity-Based mapping.
        - NO -> Consider Event-Centric Mapping.
    - In Event centric mapping, every incoming event is treated as a unique, historical fact.
        - If cosing this approach, you must document risk of unbounded graph growth
        - Unique IDs required to prevent phantom query results
        - If you send `SourceChange::Insert` with an existing Node ID, Drasi will overwrite the node in storage but will not retract the previous query results. This can causes phantom results in downstream reactions.
        - The consumer thinks the old event and new event both exist.
        
```

.github/agents/source-plan-executor.md
```diff
@@ -0,0 +1,510 @@
...
+#### Bootstrap Provider
+- Follow plan's data loading strategy
+- Implement all config fields
+- Handle errors gracefully
+
+#### Change Detection
+- Implement CDC mechanism from plan (no placeholders)
+- Use StateStore to persist cursor/position/offset
+- Handle reconnection and resume scenarios
+- Add debug logging for troubleshooting
```
amansinghoriginal, line 84:
```suggestion
- Add debug logging for troubleshooting
- For event-stream based sources, default to Entity-Based mapping whenever possible along with  `SourceChange::Update` which has Upsert semantics.
    - Entity-based mapping means that every incoming event is treated as a state change to a persistent entity.
        - Because Update has upsert semantics, source can be stateless
        - This prevents unbounded graph growth and ensures correct query semantics
    - Decsision Tree:
        - Can you derive a stable Entity ID (pk, user_id, device_id) from the events?
        - YES -> Use Entity-Based mapping.
        - NO -> Consider Event-Centric Mapping.
    - In event-centric mapping, every incoming event is treated as a unique, historical fact.
        - If cosing this approach, you must document risk of unbounded graph growth
        - Unique IDs required to prevent phantom query results
        - If you send `SourceChange::Insert` with an existing Node ID, Drasi will overwrite the node in storage but will not retract the previous query results. This can causes phantom results in downstream reactions.
        - The consumer thinks the old event and new event both exist.
```

amansinghoriginal (changes requested):
Should we add more explicit instructions about how to write sources?
Like how constructors / start() methods should be written etc.

.github/agents/source-planner.md
danielgerlag: I tested this with Kafka, and found the best workflow here is to review the mapping in the planning review stage. I then modified the plan to always use SourceChange::Insert. I think the user of the agents needs to be part of this planning workflow.

.github/agents/source-plan-executor.md
danielgerlag: This is not accurate?

```
If you send `SourceChange::Insert` with an existing Node ID, Drasi will overwrite the node in storage
```

Issue #149, opened 2026-01-15
Sync Qdrant Reaction for Drasi Lib


amansinghoriginal:
### Overview of feature request

Real-time RAG pipelines are increasingly common in AI applications. Currently, keeping vector stores synchronized with source data requires:
- Custom CDC pipelines
- Batch ETL jobs with stale data
- Complex event-driven architectures

Drasi's continuous queries are a natural fit for this problem: define what data to index via Cypher/GQL, and the reaction handles embedding generation and vector store synchronization automatically.

This reqeuest is to build a new reaction plugin for `drasi-lib` that synchronizes continuous query results to [Qdrant](https://qdrant.tech/) vector database in real-time. This enables building RAG (Retrieval-Augmented Generation) applications where the semantic search index stays automatically synchronized with source data.

### Must Have
- [ ] `sync-vectorstore-core` crate with shared utilities
  - [ ] `EmbeddingService` trait with OpenAI/Azure OpenAI implementation
  - [ ] Handlebars template rendering with caching
  - [ ] `SyncPointStore` trait for idempotency
  - [ ] `ChangeProcessor` for batching and embedding
- [ ] `sync-vectorstore-qdrant` crate implementing `Reaction` trait
  - [ ] Collection creation with configurable vector settings
  - [ ] String-to-UUID key conversion (deterministic)
  - [ ] Batch upsert for additions/updates
  - [ ] Batch delete for removals
  - [ ] Sync point persistence in Qdrant
  - [ ] Recovery from last sync point on startup
- [ ] Integration tests with Qdrant (via Docker)
- [ ] Example application demonstrating real-time RAG
- [ ] Documentation (README, inline docs, usage examples)
- [ ] Tutorial in [drasi-project/learning](https://github.com/drasi-project/learning) repository
  - [ ] Hands-on guide for building real-time RAG with Drasi Lib
  - [ ] Runnable in VS Code Dev Containers / GitHub Codespaces
  - [ ] Step-by-step walkthrough with explanations

### Nice to Have
- [ ] Configurable batch sizes for embedding API calls
- [ ] Retry logic with exponential backoff for API failures
- [ ] Metrics/logging for monitoring sync performance
- [ ] Support for Qdrant named vectors
- [ ] Title field stored as separate payload field for display

### Future Work

This architecture enables adding more vector store reactions in the future:
- `sync-vectorstore-lancedb` - Embedded/serverless mode
[...]

PR #154, opened 2026-01-20
Add Qdrant vector database reaction for drasi-lib

amansinghoriginal:
## Summary

This PR introduces a new reaction plugin (`drasi-reaction-qdrant`) that materializes Drasi query results as embeddings in a Qdrant vector database. It generates documents using handlebar-templates for the query results, and then generates embeddings for those documents using models, and finally stores them as searchable vector points in Qdrant.

Fixes - #149 

## Features

- **Embedder Trait**: Abstraction for embedding generation with two implementations:
  - `MockEmbedder`: SHA-256 based deterministic embeddings for testing
  - `AzureOpenAIEmbedder`: Real embeddings via Azure OpenAI API
  
- **Batch Optimization**: Configurable batch sizes for embedding API calls and Qdrant upserts
  - `embedding_batch_size` (default: 50): Max documents per embedding API call
  - `upsert_batch_size` (default: 100): Max points per Qdrant upsert call
  
- **Retry Logic**: Configurable retry with exponential backoff and jitter
  - `max_retries` (default: 3): Max retry attempts
  - `initial_delay_ms` (default: 100): Initial delay between retries
  - `max_delay_ms` (default: 5000): Maximum delay cap
  - Snapshot phase: fails startup on persistent error (prevents inconsistent state)
  - Stream phase: panics on persistent failure (zero data loss guarantee)

- **Materialized View Sync Protocol**: Loads existing query results on startup via `get_current_results()` before processing incremental updates

- **Deterministic UUID Generation**: RFC 4122 UUID v5 for consistent point IDs across upsert/delete operations (enables idempotent operations)

- **Handlebars Templating**: Flexible document content generation from query results

## Changes

| File | Description |
|------|-------------|
| `Cargo.toml` | Added `components/reactions/qdrant` to workspace |
| `components/reactions/qdrant/Cargo.toml` | Crate manifest with dependencies |
| `components/reactions/qdrant/README.md` | User guide and configuration reference |
| `components/reactions/qdrant/DESIGN.md` | Future architecture roadmap |
| `components/reactions/qdrant/src/lib.rs` | Public API re-exports |
| `components/reactions/qdrant/src/config.rs` | Configuration types |
| `components/reactions/qdrant/src/document.rs` | Document building and UUID generation |
| `components/reactions/qdrant/src/embedder/mod.rs` | Embedder trait definition |
| `components/reactions/qdrant/src/embedder/mock.rs` | MockEmbedder implementation |
| `components/reactions/qdrant/src/embedder/azure_openai.rs` | AzureOpenAIEmbedder implementation |
| `components/reactions/qdrant/src/qdrant.rs` | Main QdrantReaction implementation |
| `components/reactions/qdrant/tests/qdrant_helpers.rs` | Testcontainers utilities |
| `components/reactions/qdrant/tests/integration_test.rs` | Integration tests |
| `components/reactions/qdrant/tests/live_azure_test.rs` | Live Azure OpenAI integration test (ignored by default) |

## Test Plan

- Unit tests pass (40 tests)
- Integration tests pass with testcontainers (17 tests)
- Doc tests compile-checked (5 tests with `no_run`)
- Live Azure OpenAI test verified (1 ignored test)
- `cargo clippy` passes
- `cargo fmt` applied

### Running Tests

```bash
# Unit tests only
cargo test -p drasi-reaction-qdrant --lib

# Integration tests (requires Docker)
cargo test -p drasi-reaction-qdrant --test integration_test

# All tests
cargo test -p drasi-reaction-qdrant

# Live Azure OpenAI test (requires Azure credentials)
export TEST_AZURE_OPENAI_ENDPOINT="https://your-resource.openai.azure.com"
export TEST_AZURE_OPENAI_KEY="your-api-key"
export TEST_AZURE_OPENAI_MODEL="text-embedding-3-large"
cargo test -p drasi-reaction-qdrant --test live_azure_test -- --ignored
```

## Dependencies Added

| Crate | Version | Purpose |
|-------|---------|---------|
| `qdrant-client` | 1.12 | Qdrant gRPC client |
| `reqwest` | 0.12 | HTTP client for Azure OpenAI |
| `handlebars` | 5.1 | Template rendering |
| `sha2` | 0.10 | SHA-256 hashing for mock embeddings |
| `rand` | 0.8 | Retry jitter randomization |
| `testcontainers` | 0.23 | Integration testing (dev) |

### DIFF
--- a/components/reactions/qdrant/README.md
+++ b/components/reactions/qdrant/README.md
@@ -1,0 +1,620 @@
+# Qdrant Reaction
+
+A vector database reaction that synchronizes Drasi query results to Qdrant for semantic search and retrieval-augmented generation (RAG) applications.
+
+## Overview
+
+The Qdrant Reaction generates embeddings from query result data and stores them as searchable vector points in a Qdrant collection. It enables semantic search over your continuous query results, making it ideal for building AI-powered search, recommendation systems, and RAG pipelines.
+
+**Output Method**: Uses gRPC to communicate with Qdrant, supporting both local instances and Qdrant Cloud.
+
+### Key Capabilities
+
+- **Automatic Embedding Generation**: Generates embeddings using Azure OpenAI or mock embedders
+- **Template-Based Content**: Uses Handlebars templates to construct document content from query results
+- **Deterministic Point IDs**: RFC 4122 UUID v5 ensures consistent point IDs for upsert/delete operations
+- **Collection Management**: Optionally creates collections with correct vector dimensions
+- **Full Lifecycle Support**: Handles ADD, UPDATE, and DELETE operations from continuous queries
+
+### Use Cases
+
+- **Semantic Search**: Enable natural language search over structured data
+- **RAG Applications**: Build retrieval-augmented generation pipelines
+- **Recommendation Systems**: Find similar items based on content
+- **Knowledge Bases**: Create searchable document repositories from query results
+- **Content Discovery**: Surface related content based on semantic similarity
+
+**Best for**: Applications requiring semantic search over continuously updated data.
+
+**Not recommended for**: Scenarios where exact keyword matching is sufficient.
+
+### Synchronization Protocol
+
+The reaction follows the drasi-lib materialized view protocol:
+
+1. **Subscribe** - Subscribes to configured queries, buffering incoming events
+2. **Snapshot** - Loads existing query results via `get_current_results()`
+3. **Materialize** - Batch upserts snapshot data to Qdrant collection
+4. **Process** - Processes incremental updates from the event buffer
+
+This ensures the reaction starts with a consistent view of all existing data before processing incremental changes. Progress is logged for large datasets (> 1000 items).
+
+## Configuration
+
+### Builder Pattern (Recommended)
+
+The builder pattern provides a fluent, type-safe API for creating QdrantReaction instances.
+
+#### With MockEmbedder (Testing)
+
+```rust
+use drasi_reaction_qdrant::{
+    QdrantReaction, QdrantReactionConfig, QdrantConfig,
+    EmbeddingConfig, DocumentConfig, BatchConfig, RetryConfig, MockEmbedder,
+};
+use std::sync::Arc;
+
+let config = QdrantReactionConfig {
+    qdrant: QdrantConfig {
+        endpoint: "http://localhost:6334".to_string(),
+        api_key: None,
+        collection_name: "my_documents".to_string(),
+        create_collection: true,
+    },
+    embedding: EmbeddingConfig::Mock { dimensions: 1536 },
+    document: DocumentConfig {
+        document_template: "{{name}}: {{description}}".to_string(),
+        key_field: "id".to_string(),
+        ..Default::default()
+    },
+    batch: BatchConfig::default(),
+    retry: RetryConfig::default(),
+};
+
+let embedder = Arc::new(MockEmbedder::new(1536));
+let reaction = QdrantReaction::with_embedder(
+    "my-qdrant-reaction",
+    vec!["product-query".to_string()],
+    config,
+    embedder,
+)?;
+
+drasi.add_reaction(Arc::new(reaction)).await?;
+```
+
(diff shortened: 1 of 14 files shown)


components/reactions/qdrant/README.md
```diff
@@ -0,0 +1,629 @@
...
+        create_collection: true,
+    },
+    embedding: EmbeddingConfig {
+        service_type: EmbeddingServiceType::Mock,
+        dimensions: 1536,
+        ..Default::default()
+    },
+    document: DocumentConfig {
+        document_template: "{{name}}: {{description}}".to_string(),
+        key_field: "id".to_string(),
```
danielgerlag, line 67: Have we considered the use case of simply embedding the events, rather than entities with IDs?
eg. "As at {{timestamp}} there were {{count}} freezers above 32"

components/reactions/qdrant/src/config.rs
```diff
@@ -0,0 +1,542 @@
...
+/// Configuration for the embedding service.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct EmbeddingConfig {
+    /// Type of embedding service to use
+    pub service_type: EmbeddingServiceType,
+
+    /// Azure OpenAI endpoint (required for azure_openai service type)
+    /// e.g., "https://myresource.openai.azure.com"
+    #[serde(default)]
+    pub endpoint: Option<String>,
```
danielgerlag: I think these fields should be moved into the AzureOpenAI enum variant, as they are specific to the implementation. Then each implementation can have it's own set of config options rather than trying to fit them into a one size fits all. I imagine a future option might be a local model.
It might also make sense to extract the embedding functionality into it's own crate with abstractions, and concrete implementations in their own crates?

components/reactions/qdrant/src/config.rs
```diff
@@ -0,0 +1,542 @@
...
+/// Configuration for retry behavior on transient failures.
+///
+/// Uses exponential backoff with jitter to handle transient errors
+/// from Qdrant and embedding services.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RetryConfig {
+    /// Maximum number of retry attempts before failing.
+    /// Default: 3
+    #[serde(default = "default_max_retries")]
+    pub max_retries: u32,
```
danielgerlag, line 222: What happens after final failure?

components/reactions/qdrant/README.md
amansinghoriginal: Vector stores are optimized for semantic similarity search on content. Events with the same template but different values (timestamps, counts) would embed to nearly identical vectors, making similarity search meaningless.

The entity-based design with stable IDs fits both Drasi's model (structured data with updates/deletes) and vector store strengths (semantic search, idempotent upserts). Temporal event queries are better served by time-series databases.

components/reactions/qdrant/src/config.rs
amansinghoriginal: Good idea to do it now, instead of doing it later when we do common crates.
I've now added `EmbeddingConfig`:

```                                                                                                                                                                
  #[serde(tag = "service_type", rename_all = "snake_case")]
  pub enum EmbeddingConfig {
      Mock { dimensions },
      AzureOpenai { endpoint, model, api_key, dimensions, api_version },
  }
```
This makes it easy to add new providers (e.g., local models, other cloud services) with their own config shapes without affecting existing variants.

components/reactions/qdrant/src/config.rs
amansinghoriginal: Previously we would just panic (as opposed to default behavior of log-and-continue in other reactions). For an embedded library, that removes control from the process that embeds it.

So, I have now added an optional ErrorHandler callback that lets consumers decide how to handle failures after retry exhaustion:

```
pub type ErrorHandler = Arc<dyn Fn(anyhow::Error) + Send + Sync>; 
```

Default behavior (no handler set): errors are logged and processing continues. With handler: consumers can implement any policy. This replaces the previous panic!() approach, so it should give consumers full control over failure handling.

PR #170, opened 2026-01-23
Source planner/implementation agents

danielgerlag:
This pull request introduces a comprehensive agent specification for the new `source-planner` and `source-plan-executor`.

For local usage with the Copilot CLI:

- Start copilot CLI
- Select the `source-planner` agent using `/agent`
- Select the `claude-sonnet-4.5` or `claude-opus-4.5` model using `/model` (CLI does not use the model defined in the frontmatter)
- Run a prompt like `Please write a plan for an XXXX source and save it to file in my workspace`
- Once the plan is complete, review it and tweak it in the generated file
- Switch to the `gpt-5.2-codex` model using `/model`
- Switch to the `source-plan-executor` agent using `/agent`
- Run the prompt: `please implement the plan`


.github/agents/source-plan-executor.md
ruokun-niu, line 1: I think currently the components reference drasi-lib through the following way:
```toml
drasi-lib.workspace = true
```
I'm curious to hear your thoughts on if we should let the agent reference the latest version of lib on crates.io instead

PR #174, opened 2026-01-23
refactor(reaction-sse): remove dev dependency on drasi-source-applica…

agentofreality:
…tion

Make drasi-reaction-sse a fully independent, self-contained component by removing its dev dependency on drasi-source-application.

Changes:
- Create tests/mock_source.rs with MockSource implementing drasi-lib Source trait
- Update integration_tests.rs to use MockSource instead of ApplicationSource
- Remove drasi-source-application from Cargo.toml dev-dependencies
- Add drasi-core, ordered-float, chrono as dev-dependencies for mock source

The mock source provides the same functionality needed for testing:
- MockSourceHandle for sending events programmatically
- PropertyMapBuilder for constructing test data
- Full Source trait implementation using drasi-lib base helpers

This follows the design principle that components should only share types/traits from drasi-core or drasi-lib, not from other components.


components/reactions/sse/Cargo.toml
```diff
@@ -47,8 +47,10 @@ handlebars = "5.1"
 
 [dev-dependencies]
 reqwest = { version = "0.11", features = ["stream"] }
-drasi-source-application.workspace = true
+drasi-core.workspace = true
```
danielgerlag, line 50: This dependency was just a dev dependency for tests?

agentofreality: Yes, claude originally decided to use the application source when it created some tests, creating this cross plugin dependency.

danielgerlag: given, it's in `dev-dependencies`, I don't see this dependency as an issue.

PR #177, opened 2026-01-23
Fix stack overflow in broadcast receiver when lagging

WHOIM1205:
## Description

This PR fixes an unbounded recursive call in `BroadcastChangeReceiver::recv()` that could cause a stack overflow and crash the process under sustained load when a broadcast receiver falls behind the sender.

When a subscriber lags, `tokio::broadcast::Receiver::recv()` returns `RecvError::Lagged`. The previous implementation handled this by recursively calling `recv()` again, which results in unbounded stack growth if lag persists — a realistic scenario under high-throughput sources, slow consumers, CPU throttling, or pod restarts.

Since this logic is part of the core broadcast dispatch path, the failure can crash the entire Drasi process in production.

The fix replaces the recursive retry with an iterative loop, preserving existing behavior (dropping lagged messages and continuing) while ensuring constant stack usage.

---

## Type of change

- [x] Bug fix (non-breaking change which fixes a production issue)

---

## Verification

- Added unit tests that cover repeated lag scenarios which previously caused unbounded recursion
- Verified that the receiver continues operating correctly after severe lag
- Confirmed normal message delivery resumes after lag events

Verified locally with:
```bash
cargo test

### DIFF
--- a/lib/src/channels/dispatcher.rs
+++ b/lib/src/channels/dispatcher.rs
@@ -248,17 +248,33 @@ impl<T> ChangeReceiver<T> for BroadcastChangeReceiver<T>
 where
     T: Clone + Send + Sync + 'static,
 {
+    /// Receive the next change from the broadcast channel.
+    ///
+    /// # Lag Handling
+    ///
+    /// When the receiver falls behind the sender (lag), this method uses an
+    /// **iterative loop** instead of recursion to handle repeated lag events.
+    /// This prevents stack overflow under sustained high throughput where a slow
+    /// subscriber might experience continuous lag.
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(Arc<T>)` - The next available message
+    /// * `Err(...)` - If the broadcast channel is closed
     async fn recv(&mut self) -> Result<Arc<T>> {
-        match self.rx.recv().await {
-            Ok(change) => Ok(change),
-            Err(broadcast::error::RecvError::Closed) => {
-                Err(anyhow::anyhow!("Broadcast channel closed"))
-            }
-            Err(broadcast::error::RecvError::Lagged(n)) => {
-                // Log the lag but try to continue
-                log::warn!("Broadcast receiver lagged by {n} messages");
-                // Try to receive the next message
-                self.recv().await
+        loop {
+            match self.rx.recv().await {
+                Ok(change) => return Ok(change),
+                Err(broadcast::error::RecvError::Closed) => {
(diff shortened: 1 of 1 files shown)


amansinghoriginal:
Hi @WHOIM1205, thanks for your contribution!

This PR doesn't appear to have a linked issue. Per our [contributing guidelines](https://github.com/drasi-project/drasi-core/blob/main/CONTRIBUTING.md#current-status), please link the relevant issue in your PR description using "Fixes #<issue-number>".

Join our [Discord server](https://aka.ms/drasidiscord) to engage with maintainers and the community!

danielgerlag (commented):
Hi @WHOIM1205 👋

   Thank you for this contribution! The fix is clean and correct — replacing the recursive `self.recv().await` with an iterative `loop`/`continue` is exactly the
   right approach, and the tests you've added provide solid coverage for the lag scenarios.

   Before we can merge, there are just a couple of housekeeping items:

   1. **Linked issue**: Per our [contributing guidelines](https://github.com/drasi-project/drasi-core/blob/main/CONTRIBUTING.md#current-status), please create an
   issue for this bug and link it in the PR description using `Fixes #<issue-number>`.

   2. **Rebase onto `main`**: The branch is currently behind `main` and will need a rebase.

   3. **Optional simplification**: `test_broadcast_receiver_handles_multiple_consecutive_lags_without_stack_overflow` and
  `test_broadcast_receiver_handles_extreme_lag_scenario` cover essentially the same scenario (same pattern, different message counts). If you'd like to
  consolidate them into one test, that would be fine — but this is not a blocker.

   The core fix looks good and I'm happy to approve once the issue is linked and the branch is rebased. Thanks again!

agentofreality:
@danielgerlag please review this and determine if it is still relevant given recent changes. If so, resolve.

danielgerlag:
Thanks for this fix! Unfortunately this PR is now obsolete — the exact change it proposes has already landed on `main`.

`BroadcastChangeReceiver::recv()` was converted from the recursive `self.recv().await` retry to an iterative `loop` as part of #349 ("Runtime plugins", commit de6064a7, merged 2026-04-14). The current code in `lib/src/channels/dispatcher.rs` is functionally identical to what this PR adds, which is also why the PR now shows as conflicting against `main`.

The one thing not covered on `main` is the unit tests included here for the repeated-lag scenario. If those are valuable, feel free to open a follow-up PR with just the tests (rebased on current `main`).

Closing as already-resolved.

PR #184, opened 2026-01-23
MSSQL Source

danielgerlag:
This pull request adds full support for Microsoft SQL Server (MSSQL) as both a Change Data Capture (CDC) source and as a bootstrap provider for the Drasi platform. It introduces new plugins, configuration, documentation, and connection management for MSSQL, enabling Drasi to ingest and bootstrap data from SQL Server databases.

Out of scope for this PR is managed identity authentication.

Resolves : #136

### DIFF
--- a/components/sources/mssql/tests/integration_test.rs
+++ b/components/sources/mssql/tests/integration_test.rs
@@ -420,7 +420,7 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
         execute_sql(
             &mut client,
             "INSERT INTO dbo.TypesTest (Id, IntVal, BigIntVal, SmallIntVal, TinyIntVal, BitVal, FloatVal, RealVal, DecimalVal, VarcharVal, NVarcharVal) \
-             VALUES (1, 42, 9876543210, 256, 7, 1, 3.14, 2.5, 99.95, 'hello', N'world');",
+             VALUES (1, 42, 9876543210, 256, 7, 1, 3.15, 2.5, 99.95, 'hello', N'world');",
         )
         .await?;
 
@@ -479,8 +479,8 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
             );
             let float_val = data.get("float_val").unwrap().as_f64().unwrap();
             assert!(
-                (float_val - 3.14).abs() < 0.001,
-                "FLOAT value should be ~3.14, got: {float_val}"
+                (float_val - 3.15).abs() < 0.001,
+                "FLOAT value should be ~3.15, got: {float_val}"
             );
 
             assert!(


components/bootstrappers/mssql/src/mssql.rs
```diff
@@ -0,0 +1,525 @@
...
+        client: &mut tiberius::Client<tokio_util::compat::Compat<tokio::net::TcpStream>>,
+        label: &str,
+        table_name: &str,
+        context: &BootstrapContext,
+        event_tx: &tokio::sync::mpsc::Sender<BootstrapEvent>,
+    ) -> Result<usize> {
+        debug!("Starting bootstrap of table '{table_name}' with label '{label}'");
+
+        // Query all rows from the table
+        let query = format!("SELECT * FROM {table_name}");
```
ruokun-niu, line 313: Should we add a check for injection attack

danielgerlag: added

ruokun-niu (approved):

components/sources/mssql/src/stream.rs
```diff
@@ -0,0 +1,727 @@
...
+    validate_sql_identifier(table)?;
+
+    // MS SQL CDC function name: cdc.fn_cdc_get_all_changes_{capture_instance}
+    // Capture instance is usually: {schema}_{table}
+    // If table already contains schema (e.g., "dbo.Orders"), replace dot with underscore
+    let capture_instance = table.replace('.', "_");
+
+    // Tiberius uses @P1, @P2, etc. for positional parameters
+    let query = format!(
+        "SELECT * FROM cdc.fn_cdc_get_all_changes_{capture_instance}(@P1, @P2, 'all') ORDER BY __$start_lsn, __$seqval"
```
amansinghoriginal: Does `__$seqval`  guarantee correct ordering of DML statements within a single transaction?

components/sources/mssql/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    Ok(properties)
+}
+
+/// Extract a column value from a row by index and convert to ElementValue
+///
+/// This tries each type in sequence since tiberius requires knowing the exact type.
+pub fn extract_column_value(row: &Row, col_idx: usize) -> Result<ElementValue> {
+    // Try string types first (most common)
+    if let Ok(val) = row.try_get::<&str, _>(col_idx) {
```
amansinghoriginal: The bootstrap provider matches on column.column_type() to drive type selection. Should we use that same logic here?

amansinghoriginal: This tries row.try_get::<&str>() before numeric types.
Is it possible that value of 42 becomes ElementValue::String("42") instead of ElementValue::Integer(42) ?

components/bootstrappers/mssql/src/mssql.rs
```diff
@@ -0,0 +1,530 @@
...
+                .bootstrap_table(client, label, table_name, context, &event_tx)
+                .await?;
+
+            total_count += count;
+            info!("Bootstrapped {count} records from table '{table_name}'");
+        }
+
+        // Commit transaction
+        client
+            .simple_query("COMMIT TRANSACTION")
```
amansinghoriginal, line 240: In the PostgreSQL botstrap, we capture `pg_current_wal_lsn()` inside its snapshot transaction and store it for the source. Should we use `sys.fn_cdc_get_max_lsn()` and persist the result to the state store under the same `checkpoint.lsn` key the source reads from?

amansinghoriginal (changes requested):
Great work on this PR. The overall architecture is solid. Check the comments once on the usage of `__$seqval` and capturing LSN from bootstrap. 

Otherwise, the error classification system with exponential backoff, the custom LSN type, SQL identifier validation, and PrimaryKeyCache are all well-engineered. Clean module separation and good integration test coverage.  👍🏼

components/sources/mssql/src/stream.rs
danielgerlag: resolved in commit: https://github.com/drasi-project/drasi-core/pull/184/changes/a091cdfd2908c94e6dd089f5f1df8b030679376d

components/sources/mssql/src/types.rs
danielgerlag: resolved in 7ccfbc31794e5679297a1a0afcfb0f0626d410ef

components/bootstrappers/mssql/src/mssql.rs
danielgerlag: The Postgres bootstrapper captures  pg_current_wal_lsn() but doesn't actually persist it to the state store either, it only logs it. Neither bootstrapper currently coordinates with the source's checkpoint.

danielgerlag: It would not make sense, because if you fire a a second query after query 1 is already using the source, there will be second bootstrap, and we could skip updates for the first query.

Issue #168, opened 2026-01-22
Use test containers for Redis/Garnet Unit Tests


amansinghoriginal:
### Area for Improvement

In drasi-core some unit tests require a running Redis instance.
These could be updated to use testcontainers instead.

### Desired behavior

All unit tests related to redis/garnet index in Drasi-Core should use test containers and not rely on external deployment.

### Drasi Version

All

ahmed-kamal2004:
@amansinghoriginal  Great!,  I will work on this issue within the current time.

PR #185, opened 2026-01-23
components: use redis test container within garnet scenario tests

ahmed-kamal2004:
# Description
Instead of relying on external Redis instance, this PR utilizes the auto-managed test containers for Garnet storage unit tests 

## Type of change

Yes, issue link https://github.com/drasi-project/drasi-core/issues/168

Fixes: https://github.com/drasi-project/drasi-core/issues/168

## Additional
I have seen multiple approaches to tackle this issue, one of them is to use directly `testcontainers-module` within `scenario_tests.rs`, but I didn't prefer that, since there is already `redis_helpers.rs` that has the main blocks to use redis test containers.

I used the second option, better for maintaining.

example for the first unused option.
```
let redis_instance: testcontainers::ContainerAsync<Redis> = Redis::default().with_tag("7-alpine").start().await.unwrap();
let url = format!(
 "redis://{host}:{port}",
 host = redis_instance.get_host().await.unwrap(),
 port = redis_instance.get_host_port_ipv4(REDIS_PORT).await.unwrap(),
 );
```

- Another closed PR, because the branch is changed #181 

- I have done local tests for the whole project and drasi-index-garnet lonely, and checked that there no change in behavior.

### DIFF
--- a/components/indexes/garnet/tests/scenario_tests.rs
+++ b/components/indexes/garnet/tests/scenario_tests.rs
@@ -12,10 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{
-    env,
-    sync::{Arc, Mutex},
-};
+use std::sync::{Arc, Mutex};
 
 use shared_tests::redis_helpers::{setup_redis, RedisGuard};
 
@@ -122,6 +119,7 @@ mod building_comfort {
         let element_index = test_config.get_element_index();
         element_index.clear().await.unwrap();
         println!("Element Index Cleared");
+        test_config.redis_grd.cleanup().await;
     }
 
     // #[tokio::test]
@@ -141,6 +139,7 @@ mod curbside_pickup {
     async fn order_ready_then_vehicle_arrives() {
         let test_config = GarnetQueryConfig::new(false).await;
         curbside_pickup::order_ready_then_vehicle_arrives(&test_config).await;
+        test_config.redis_grd.cleanup().await;
     }
 
     #[tokio::test]
@@ -148,6 +147,7 @@ mod curbside_pickup {
     async fn vehicle_arrives_then_order_ready() {
         let test_config = GarnetQueryConfig::new(false).await;
         curbside_pickup::vehicle_arrives_then_order_ready(&test_config).await;
+        test_config.redis_grd.cleanup().await;
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


components/indexes/garnet/tests/scenario_tests.rs
```diff
@@ -17,6 +17,8 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use shared_tests::redis_helpers::{setup_redis, RedisGuard};
```
amansinghoriginal, line 17: The `env` import (line 16) is no longer needed since the `REDIS_URL` environment variable logic was removed. Please update the import to:

```rust
use std::sync::{Arc, Mutex};
```

components/indexes/garnet/tests/scenario_tests.rs
```diff
@@ -116,7 +117,7 @@ mod building_comfort {
 
     #[tokio::test]
     async fn building_comfort_use_case() {
-        let test_config = GarnetQueryConfig::new(false);
+        let test_config = GarnetQueryConfig::new(false).await;
```
amansinghoriginal, line 117: This test is missing the explicit `cleanup()` call that all other tests in this file have.

**Why explicit cleanup matters:**

The `RedisGuard` will clean up the container via its `Drop` implementation, but this relies on the tokio runtime still being available and can be less reliable in test scenarios. The `RedisGuard` documentation recommends calling `.cleanup().await` explicitly:

> **RECOMMENDED**: Call `.cleanup().await` explicitly before the test ends for the most reliable cleanup.

**The pattern used in other tests:**

```rust
#[tokio::test]
async fn some_test() {
    let test_config = GarnetQueryConfig::new(false).await;
    // ... test logic ...
    test_config.redis_grd.cleanup().await;
}
```

Please add `test_config.redis_grd.cleanup().await;` before the closing brace of this function.

components/indexes/garnet/tests/scenario_tests.rs
```diff
@@ -138,35 +139,40 @@ mod curbside_pickup {
...
         curbside_pickup::order_ready_then_vehicle_arrives(&test_config).await;
+        test_config.redis_grd.cleanup().await;
     }
 
     #[tokio::test]
     #[ignore]
     async fn vehicle_arrives_then_order_ready() {
-        let test_config = GarnetQueryConfig::new(false);
+        let test_config = GarnetQueryConfig::new(false).await;
         curbside_pickup::vehicle_arrives_then_order_ready(&test_config).await;
```
amansinghoriginal, line 149: Nit: Missing semicolon here and in most other `cleanup().await` calls.

**Codebase convention:**

Looking at the existing patterns in this repo:
- `redis_helpers.rs:486` uses `redis.cleanup().await;` (with semicolon)
- All tests in `rocksdb/tests/scenario_tests.rs` end with `.await;` (with semicolon)

**Why it matters in Rust:**

In Rust, semicolons have semantic meaning:
- `cleanup().await;` → Statement, value discarded, block returns `()`
- `cleanup().await` → Expression, value becomes the block's return value

Both compile here because `cleanup()` returns `()` and the test function expects `()`. But omitting the semicolon is "accidental correctness" - if `cleanup()` ever changed its return type, the no-semicolon version would break.

**What to fix:**

Line 144 has the correct style. Please add semicolons to all other `cleanup().await` calls.

amansinghoriginal (approved):
Thanks @ahmed-kamal2004 for a clean implementation addressing all feedback.
This looks good to merge! 🎉

ruokun-niu (approved):
Just tested it locally and everything works. Thanks @ahmed-kamal2004 !

PR #194, opened 2026-01-24
Fix panic in delete_source_join on missing join key

WHOIM1205:
## Description

This PR fixes a **critical panic** in `InMemoryElementIndex::delete_source_join` caused by an unchecked `.unwrap()` on a potentially missing `HashMap` key.

When processing DELETE events, the engine assumed that the corresponding `join_key` always existed in the in-memory `partial_joins` index. In real distributed CDC scenarios, this assumption does not hold and can result in a **hard panic** that crashes the engine.

This issue can occur in the following realistic situations:
- **Crash recovery / restart** where the in-memory index is cleared
- **Out-of-order event delivery** (DELETE processed before INSERT)
- **Race conditions** where `update_source_joins` skips insertion but deletion still occurs

Since DELETE handling happens inside the reconciliation/event-processing loop, a single invalid DELETE event could repeatedly crash the engine, leading to a **full service outage**.

The fix replaces the unchecked `.unwrap()` with safe handling that treats DELETEs for missing join keys as a no-op, preserving correct semantics while preventing crashes.

---

## Type of change

- **BUGFIX** (non-breaking change which fixes a critical runtime panic)

---

### Tests added

- Added regression tests to ensure `delete_source_join` does not panic when:
  - The join key was never registered
  - The element was already missing
- Added tests to validate correct behavior when the join key exists
- Ensured index integrity remains intact after no-op deletes

### How to run tests

```bash
cargo test --package drasi-core --lib in_memory_index::in_memory_element_index::tests

### DIFF
--- a/core/src/in_memory_index/in_memory_element_index.rs
+++ b/core/src/in_memory_index/in_memory_element_index.rs
@@ -371,7 +371,10 @@ impl InMemoryElementIndex {
             .entry((query_join.id.clone(), value_hash))
             .or_default();
 
-        let did_remove = partial_joins.get_mut(join_key).unwrap().remove(old_element);
+        let did_remove = partial_joins
+            .get_mut(join_key)
+            .map(|set| set.remove(old_element))
+            .unwrap_or(false);
 
         if did_remove {
             for qjk2 in &query_join.keys {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/src/in_memory_index/in_memory_element_index.rs [hunk 1]
```diff
@@ -371,7 +371,15 @@ impl InMemoryElementIndex {
             .entry((query_join.id.clone(), value_hash))
             .or_default();
 
-        let did_remove = partial_joins.get_mut(join_key).unwrap().remove(old_element);
+        // Safely handle the case where join_key was never registered in partial_joins.
+        // This can occur during:
+        // 1. Crash recovery/restart when in-memory index is lost
+        // 2. Out-of-order event delivery (DELETE before INSERT)
+        // 3. Race conditions in update_source_joins
```
danielgerlag:
```suggestion
```

core/src/in_memory_index/in_memory_element_index.rs
```diff
@@ -705,3 +713,143 @@ fn get_value_hash(value: &ElementValue) -> u64 {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::models::{ElementValue, QueryJoin, QueryJoinKey};
+
+    /// Test that delete_source_join does not panic when join_key was never registered.
+    /// This scenario occurs in production during:
+    /// 1. Crash recovery/restart when in-memory index is lost
+    /// 2. Out-of-order event delivery (DELETE before INSERT)
+    /// 3. Race conditions in update_source_joins
```
danielgerlag: Could you provide more detail on 1 and 3, not sure they are valid?

WHOIM1205: Sure, happy to clarify.

1. Crash recovery / restart: partial_joins is an in-memory structure and is rebuilt on startup. If the engine restarts after having processed INSERTs, it can still receive DELETE events (for example from replayed or delayed CDC events). In that case the corresponding join_key may never have been registered in this process, so get_mut(join_key) legitimately returns None.

3. update_source_joins: insertion into partial_joins is conditional. If an element does not meet the join criteria at insert/update time, no entry is created, but a later DELETE still attempts to remove it unconditionally. This can also lead to a missing join_key during deletion.

If you’d prefer, I’m happy to remove or simplify these comments and keep the detailed rationale in the PR description instead.

danielgerlag: These are not valid scenarios for the context within which this operates. I would remove the list of scenarios.

WHOIM1205: Thanks for the clarification 
I’ve removed the list of scenarios as suggested and pushed the update.

danielgerlag (approved):

amansinghoriginal:
Hi @WHOIM1205, thanks for your contribution!

This PR doesn't appear to have a linked issue. Per our [contributing guidelines](https://github.com/drasi-project/drasi-core/blob/main/CONTRIBUTING.md#current-status), please link the relevant issue in your PR description using "Fixes #<issue-number>".

Join our [Discord server](https://aka.ms/drasidiscord) to engage with maintainers and the community!

Issue #189, opened 2026-01-24
Clean up examples: remove unused imports and fix fragile path


amansinghoriginal:
## Overview of feature request

The `examples/` directory and some core crates have minor code quality issues that need cleanup:

1. **Unused imports/modules** - Several files have `mod` declarations or `use` imports that are never referenced
2. **Fragile path pattern** - One example uses `.parent().unwrap()` on `CARGO_MANIFEST_DIR` which breaks if the folder structure changes

These issues generate clippy warnings and represent technical debt.

### Verification

- [ ] `cargo clippy --all-targets` passes with no new warnings
- [ ] `cargo build --examples` succeeds
- [ ] `cargo run -p with-rocksdb-index` still works correctly
- [ ] All commits signed off (DCO requirement)

### Why this is a good first issue

- Self-contained task with clear scope
- Low risk - only affects examples and removes unused code
- Does not require deep knowledge of the Drasi codebase
- Teaches project structure and Rust module system
- Clear success criteria (clippy passes, examples run)

## Would you like to support us?

- [ ] Yes, I would like to support you

nitininhouse:
Hello @amansinghoriginal, can I work on this issue?

PR #198, opened 2026-01-24
fix: remove unused imports in examples

spline2hg:
# Description

Remove unused imports and module declarations

## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Drasi (#189 ).

Please update the following to link the associated issue. This is required for some kinds of changes (see above).

Fixes: #189

### DIFF
--- a/examples/lib/with-rocksdb-index/main.rs
+++ b/examples/lib/with-rocksdb-index/main.rs
@@ -87,9 +87,6 @@ async fn main() -> Result<()> {
     // =========================================================================
 
     let bootstrap_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-        .parent()
-        .expect("Cannot navigate to parent directory for bootstrap file")
-        .join("builder")
         .join("bootstrap_data.jsonl");
 
     println!("Loading bootstrap data from: {}", bootstrap_path.display());
--- a/examples/lib/with-rocksdb-index/bootstrap_data.jsonl
+++ b/examples/lib/with-rocksdb-index/bootstrap_data.jsonl
@@ -1,0 +1,5 @@
+{"kind": "Header", "start_time": "2024-01-15T09:30:00-05:00", "description": "Initial stock prices for DrasiLib Builder example"}
+{"kind": "Node", "id": "price_AAPL", "labels": ["stock_prices"], "properties": {"symbol": "AAPL", "price": 175.50, "previous_close": 173.00, "volume": 5000000}}
+{"kind": "Node", "id": "price_MSFT", "labels": ["stock_prices"], "properties": {"symbol": "MSFT", "price": 380.25, "previous_close": 382.00, "volume": 2500000}}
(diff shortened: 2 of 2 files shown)


spline2hg:
Hi @agentofreality, is there any more changes needed here?

examples/lib/with-rocksdb-index/main.rs
```diff
@@ -87,7 +87,8 @@ async fn main() -> Result<()> {
     // =========================================================================
 
     let bootstrap_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-        .parent().unwrap()
+        .parent()
+        .expect("Cannot navigate to parent directory for bootstrap file")
```
amansinghoriginal: Thanks for asking about this in the issue comments! Unfortunately, changing `.unwrap()` to `.expect()` doesn't actually fix the fragility - it just provides a better panic message when it fails.

The issue is that this path relies on the parent directory structure (`../builder/bootstrap_data.jsonl`). If someone reorganizes the examples folder, this will still break.

The fix is to make the example self-contained:
1. Copy `bootstrap_data.jsonl` into this example's directory (`examples/lib/with-rocksdb-index/`)
2. Simplify the path to:
   ```rust
   let bootstrap_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
       .join("bootstrap_data.jsonl");
   ```

This eliminates the `.parent()` call entirely and makes the example independent of sibling directories.

core/src/evaluation/functions/future/sliding_window.rs
```diff
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::result;
```
amansinghoriginal, line 15: This is a valid cleanup, but it's outside the scope of issue #189 which specifically targeted the `examples/` directory.

For future contributions, it's generally better to keep PRs focused on the issue they're addressing. If you spot other cleanups, consider opening a separate issue or PR for them.

That said, these core crate cleanups are harmless and can stay - just a note for next time!

amansinghoriginal (changes requested):
Thanks for your first contribution to drasi-core! 🎉

What's working well:
  - Part 1 (removing unused `mod process_monitor;` declarations) is done correctly
  - Commits are properly signed off
  - The bonus cleanups in `core/` are valid (though outside issue scope)

What needs revision:
  - **Part 2 is not fully addressed.** The `.expect()` change improves the error message but doesn't fix the underlying fragility. The path still depends on the parent directory structure.

Please see my inline comment on `main.rs` for the recommended fix - essentially, copy `bootstrap_data.jsonl` into the example directory and simplify the path to eliminate the `.parent()` call entirely.

After making the change, please verify:
  - `cargo build --examples` succeeds
  - `cd examples/lib/with-rocksdb-index && cargo build` succeeds
  - `cargo clippy --all-targets` passes

amansinghoriginal (approved):
Thanks for addressing the feedback! The changes look good.

PR #208, opened 2026-01-27
PostgreSQL Integration tests

danielgerlag:
This pull request adds comprehensive integration tests for the PostgreSQL source component, ensuring it correctly detects and processes logical replication events (insert, update, delete) using a real PostgreSQL instance via Docker.

The integration coverage now uses REPLICA IDENTITY DEFAULT (non-FULL) to allow updates while still exercising the “update without old tuple”
  replication path, matching real PostgreSQL constraints. This avoids disallowed updates under REPLICA IDENTITY NOTHING and keeps the test focused on ensuring
  updates are not downgraded to inserts.

### DIFF
--- a/components/sources/postgres/src/decoder.rs
+++ b/components/sources/postgres/src/decoder.rs
@@ -604,11 +604,6 @@ impl PgOutputDecoder {
     pub fn get_relation(&self, relation_id: u32) -> Option<&RelationInfo> {
         self.relations.get(&relation_id)
     }
-
-    #[cfg(test)]
-    pub fn insert_relation_for_test(&mut self, relation: RelationInfo) {
-        self.relations.insert(relation.id, relation);
-    }
 }
 
 fn read_cstring(cursor: &mut Cursor<&[u8]>) -> Result<String> {
--- a/components/sources/postgres/src/stream.rs
+++ b/components/sources/postgres/src/stream.rs
@@ -607,10 +607,6 @@ impl ReplicationStream {
             }
         }
 
-        // Note: SourceChange::Update only needs the after element
-        // The before state is tracked internally by drasi-core
-        // We still process old_tuple to ensure we have the correct element_id
-
         let after_element = Element::Node {
             metadata: ElementMetadata {
                 reference: ElementReference::new(&self.source_id, &element_id),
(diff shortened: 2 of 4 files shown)


components/sources/postgres/src/stream.rs
amansinghoriginal: Should we remove this comment now?

Cargo.toml
```diff
@@ -88,6 +88,8 @@ drasi-reaction-http = { version = "0.1.5", path = "components/reactions/http" }
 drasi-reaction-grpc = { version = "0.2.2", path = "components/reactions/grpc" }
 drasi-index-rocksdb = { version = "0.2.2", path = "components/indexes/rocksdb" }
 drasi-index-garnet = { version = "0.1.4", path = "components/indexes/garnet" }
+drasi-bootstrap-postgres = { path = "components/bootstrappers/postgres" }
```
amansinghoriginal, line 93: Do we need to add versions here following the convention?

components/sources/postgres/src/decoder.rs
```diff
@@ -604,6 +604,11 @@ impl PgOutputDecoder {
     pub fn get_relation(&self, relation_id: u32) -> Option<&RelationInfo> {
         self.relations.get(&relation_id)
     }
+
+    #[cfg(test)]
+    pub fn insert_relation_for_test(&mut self, relation: RelationInfo) {
```
amansinghoriginal: Where is this called?

components/sources/postgres/tests/postgres_helpers.rs
```diff
@@ -0,0 +1,272 @@
...
+    );
+    execute_sql(client, &create_sql).await?;
+
+    let replica_sql = format!("ALTER TABLE {table_name} REPLICA IDENTITY FULL");
+    execute_sql(client, &replica_sql).await?;
+
+    Ok(())
+}
+
+pub async fn create_test_table_replica_identity_nothing(
```
amansinghoriginal: I guess this is unused helper?

components/sources/postgres/tests/integration_tests.rs
amansinghoriginal, line 1: All tests are set `enable_bootstrap(true)`, but all tests seem to start with empty table. Maybe in a follow-up PR we need to add tests for the bootstrap path? Should we open an issue to track this?

amansinghoriginal (approved):
Approve with suggestions.

PR #225, opened 2026-02-01
fix: compilation errors in examples with with_default_template in LogReactionConfig

### LINKED ISSUE #223: Fix LogReaction builder with_default_template handler in examples/lib
The current examples/lib uses **LogReactionConfig**  with implementations of `added_template`, `deleted_template`, and `updated_template`. These should be replaced with the `with_default_template`. The build also returns Result which should be propagated.

BipulLamsal:
# Description
Primarily example crates: `three-query`, `constructor` were throwing compilation errors due to older api structure. 
I have also fixed the usage examples for documentation.

## Type of change

- This pull request fixes a bug in Drasi and has an approved issue.

Fixes: #223


components/reactions/log/src/config.rs
```diff
@@ -43,16 +43,10 @@ pub use common::{QueryConfig, TemplateSpec};
...
-///     updated: Some(TemplateSpec {
-///         template: "[CHG] {{after.id}}".to_string(),
-///     }),
-///     deleted: Some(TemplateSpec {
-///         template: "[DEL] {{before.id}}".to_string(),
-///     }),
-/// };
+///     added: Some(TemplateSpec::new("[NEW] {{after.id}}")),
+///     updated: Some(TemplateSpec::new("[CHG] {{after.id}}")),
+///     deleted: Some(TemplateSpec::new("[DEL] {{before.id}}"),
```
amansinghoriginal: I guess the library documentation here was already correct. Please revert this file; it's introducing a syntax error (missing `)`) and is outside the scope of fixing the examples.

components/reactions/log/src/log.rs
```diff
@@ -239,21 +239,15 @@ impl LogReactionBuilder {
...
-    ///     updated: Some(TemplateSpec {
-    ///         template: "[UPD] {{after.id}}".to_string(),
-    ///     }),
-    ///     deleted: Some(TemplateSpec {
-    ///         template: "[DEL] {{before.id}}".to_string(),
-    ///     }),
+    ///     added: Some(TemplateSpec::new("[ADD] {{after.id}}")),
+    ///     updated: Some(TemplateSpec::new("[UPD] {{after.id}}")),
+    ///     deleted: Some(TemplateSpec::new("[DEL] {{before.id}}"),
     /// };
```
amansinghoriginal, line 245: I think the original doc example was valid, and this change introduces a syntax error.
Do we need this change here?

examples/lib/constructor/main.rs
```diff
@@ -196,21 +200,25 @@ async fn main() -> Result<()> {
...
 
     // Create reaction config struct
+    //
+
+    let mut routes = HashMap::new();
+    routes.insert("all-prices".to_string(), LogQueryConfig {
+        added: Some(TemplateSpec::new("[+] {{after.symbol}}: ${{after.price}} (prev: ${{after.previous_close}}, vol: {{after.volume}})")),
+        updated: Some(TemplateSpec::new("[~] {{after.symbol}}: ${{before.price}} -> ${{after.price}} (vol: {{after.volume}})".to_string())),
+        deleted: Some(TemplateSpec::new("[-] {{before.symbol}} removed".to_string())),
+    });
```
amansinghoriginal, line 210: I think there is a bug here. The `routes` variable you created on lines 203-208 isn't used here. This passes an empty HashMap and `None` for default_template, so no templates will be applied. Should we use `default_template` instead:

```rust
let log_config = LogReactionConfig {
    routes: HashMap::new(),
    default_template: Some(LogQueryConfig {
        added: Some(TemplateSpec::new("[+] ...")),
        updated: Some(TemplateSpec::new("[~] ...")),
        deleted: Some(TemplateSpec::new("[-] ...")),
    }),
};
```

examples/lib/three-query/main.rs
```diff
@@ -47,17 +47,15 @@ use axum::{
...
-use drasi_bootstrap_scriptfile::ScriptFileBootstrapProvider;
-use drasi_reaction_log::LogReaction;
 
 #[tokio::main]
 async fn main() -> Result<()> {
     // Initialize logging - set RUST_LOG=debug for more detail
-    env_logger::Builder::from_env(
-        env_logger::Env::default().default_filter_or("info")
-    ).init();
+    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
```
amansinghoriginal, line 58: The fix here is good. A note for future PRs: keeping formatting changes minimal makes diffs easier to review.

examples/lib/three-query/main.rs
```diff
@@ -178,7 +191,7 @@ async fn main() -> Result<()> {
             .with_query(high_volume_query)
             .with_reaction(log_reaction)
             .build()
-            .await?
+            .await?,
```
amansinghoriginal, line 194: why is this change required?

amansinghoriginal (changes requested):
Thanks for contributing! The three-query/main.rs fix is correct.

Please address these comment's i've added.

components/reactions/log/src/log.rs
BipulLamsal: TemplateSpec has extension field aswell which is covered by TemplateSpec::new(string) method so i thought it was necessary.

examples/lib/three-query/main.rs
BipulLamsal: sure its because of formatter i am using; i will make sure to verify it on future prs

examples/lib/three-query/main.rs
BipulLamsal: that was rustfmt adding trailing commas

components/reactions/log/src/log.rs
amansinghoriginal: You're right, good call on the `extension` field.

components/reactions/log/src/config.rs
amansinghoriginal: Checked again, you're right about the `extension` field. Keeping this.

examples/lib/constructor/main.rs
amansinghoriginal: Correction: `routes,` is struct shorthand so the variable is being used. But `default_template` is still the better fit here. Addressed in the new review.

examples/lib/constructor/main.rs
```diff
@@ -196,21 +200,25 @@ async fn main() -> Result<()> {
...
     let log_config = LogReactionConfig {
-        added_template: Some("[+] {{after.symbol}}: ${{after.price}} (prev: ${{after.previous_close}}, vol: {{after.volume}})".to_string()),
-        updated_template: Some("[~] {{after.symbol}}: ${{before.price}} -> ${{after.price}} (vol: {{after.volume}})".to_string()),
-        deleted_template: Some("[-] {{before.symbol}} removed".to_string()),
+        routes,
+        default_template: None,
     };
```
amansinghoriginal, line 214: The templates here should go in `default_template` rather than `routes`. The `routes` field is for when you need different templates for different queries. Since this example has a single query, `default_template` is the right fit.

Please replace lines 205-214 with:

```rust
    let log_config = LogReactionConfig {
        routes: HashMap::new(),
        default_template: Some(LogQueryConfig {
            added: Some(TemplateSpec::new("[+] {{after.symbol}}: ${{after.price}} (prev: ${{after.previous_close}}, vol: {{after.volume}})")),
            updated: Some(TemplateSpec::new("[~] {{after.symbol}}: ${{before.price}} -> ${{after.price}} (vol: {{after.volume}})")),
            deleted: Some(TemplateSpec::new("[-] {{before.symbol}} removed")),
        }),
    };
```

Also: `TemplateSpec::new()` accepts `impl Into<String>`, so you can pass `&str` directly without calling `.to_string()` first (lines 208-209 had that).

examples/lib/three-query/main.rs
```diff
@@ -154,14 +157,24 @@ async fn main() -> Result<()> {
...
-        .from_query("all-prices_query")
-        .from_query("gainers_query")
-        .from_query("high-volume_query")
-        .with_added_template("[{{query_name}}] + {{after.symbol}}: ${{after.price}}")
-        .with_updated_template("[{{query_name}}] ~ {{after.symbol}}: ${{before.price}} -> ${{after.price}}")
-        .with_deleted_template("[{{query_name}}] - {{before.symbol}} removed")
-        .build();
+        .from_query("all-prices")
+        .from_query("gainers")
+        .from_query("high-volume")
```
amansinghoriginal, line 175: Good catch fixing these query IDs. The original code had `"all-prices_query"` etc., but the queries are created with `Query::cypher("all-prices")`, so this was a real bug beyond the original issue. Nice attention to detail.

amansinghoriginal (changes requested):
Thanks for working on this, @BipulLamsal! Great first contribution.

**Ready to go:**
- `three-query/main.rs` fix is correct. Nice catch on the query ID mismatch too (`all-prices_query` to `all-prices` etc.), that was a real bug beyond the original issue.
- `config.rs` and `log.rs` doc updates are welcome. Using `TemplateSpec::new(...)` is the right call since the struct literal form requires the `extension` field.
- The `SourceTrait` to `Source` rename and `SourceSubscriptionConfig` field additions in `constructor/main.rs` are needed for compilation, good that you caught those.

**One change needed in `constructor/main.rs`:** the template config should use `default_template` instead of `routes`. I left an inline comment with the details and exact code.

The `rustfmt` formatting changes are fine. For future PRs, running the formatter in a separate commit helps keep the functional diff clean, but not a blocker here.

ruokun-niu:
Hi @BipulLamsal  are you still working on this PR?

PR #232, opened 2026-02-04
WebHook routes for Http source

danielgerlag:
# Add Webhook Mode to HTTP Source

## Summary

This PR extends the HTTP source component to support **generic webhook ingestion** from external services like GitHub, Shopify, Stripe, or any custom application. The existing functionality is preserved as "standard mode" while the new "webhook mode" provides flexible payload transformation using Handlebars templates.

## Motivation

The current HTTP source accepts a fixed `HttpSourceChange` schema, which requires external services to conform to Drasi's format. Many real-world integrations involve receiving webhooks from third-party services that have their own payload formats. This feature enables users to:

- Receive webhooks from GitHub, Shopify, and other providers
- Transform arbitrary JSON, XML, YAML, or plain text payloads into graph elements
- Configure multiple routes with different mappings on a single source
- Secure endpoints with HMAC signature verification or bearer tokens

### New Modules

| Module | Description |
|--------|-------------|
| `auth.rs` | HMAC signature (SHA1/SHA256) and Bearer token verification |
| `content_parser.rs` | JSON, XML, YAML, and plain text payload parsing with Content-Type auto-detection |
| `template_engine.rs` | Handlebars template engine with custom helpers and type preservation |
| `route_matcher.rs` | Route pattern matching (`:param` syntax) and condition evaluation |

### Configuration

```yaml
webhooks:
  error_behavior: accept_and_log
  cors:
    enabled: true
    allow_origins: ["*"]
  routes:
    - path: "/github/events"
      methods: ["POST"]
      auth:
        signature:
          type: hmac-sha256
          secret_env: GITHUB_WEBHOOK_SECRET
          header: X-Hub-Signature-256
          prefix: "sha256="
      mappings:
        - when:
            header: X-GitHub-Event
            equals: push
          operation: insert
          element_type: node
          template:
            id: "commit-{{payload.head_commit.id}}"
            labels: ["Commit"]
            properties:
              message: "{{payload.head_commit.message}}"
              author: "{{payload.head_commit.author.name}}"
```

### Key Features

- **Mode Exclusivity**: Webhook mode and standard mode are mutually exclusive. When `webhooks` config is present, standard endpoints return 404.
- **Template Variables**: Access `payload`, `headers`, `route` (path params), `query`, `method`, `path`, and `content_type` in templates.
- **Template Helpers**: `lowercase`, `uppercase`, `now`, `concat`, `default`, `json`
- **Payload Spread**: Use `properties: "{{payload}}"` to map all payload fields as node properties
- **Type Preservation**: Objects and arrays from payloads are preserved as `ElementValue::Object` and `ElementValue::List`
- **Timestamp Handling**: `effective_from` auto-detects Unix seconds/millis/nanos and ISO 8601 formats
- **CORS Support**: Optional CORS configuration for browser-based clients
- **Error Handling**: Configurable per-route behavior (`reject`, `accept_and_log`, `accept_silent`)

### Authentication Options

**HMAC Signature** (GitHub, Shopify):
```yaml
auth:
  signature:
    type: hmac-sha256  # or hmac-sha1
    secret_env: WEBHOOK_SECRET
    header: X-Hub-Signature-256
    prefix: "sha256="
    encoding: hex  # or base64
```

**Bearer Token**:
```yaml
auth:
  bearer:
    token_env: API_TOKEN
```

## Testing

| Category | Tests |
|----------|-------|
| Unit tests | 101 |
| E2E tests (full DrasiLib pipeline) | 10 |
| HTTP integration tests | 13 |
| **Total** | **124** |

E2E tests verify data flows through the complete pipeline: HTTP source → Query → ApplicationReaction.

## Breaking Changes

None. The default behavior (no `webhooks` config) is unchanged.

## Documentation

- Updated module-level documentation in `lib.rs`
- Comprehensive README.md with configuration reference and examples
- Examples for GitHub, Shopify, and generic REST API webhooks

### DIFF
--- a/components/sources/http/src/lib.rs
+++ b/components/sources/http/src/lib.rs
@@ -756,13 +756,13 @@ impl HttpSource {
         debug!("[{source_id}] Webhook processing complete: {success_count} succeeded, {error_count} failed");
 
         if error_count > 0 && success_count == 0 {
-            return handle_error(
+            handle_error(
                 error_behavior,
-                &source_id,
+                source_id,
                 StatusCode::BAD_REQUEST,
                 &format!("All {error_count} mappings failed"),
                 last_error.as_deref(),
-            );
+            )
         } else if error_count > 0 {
             (
                 StatusCode::OK,


components/sources/http/src/lib.rs
```diff
@@ -513,10 +583,208 @@ impl HttpSource {
...
+                    warn!("[{source_id}] Failed to process mapping: {e}");
+                    error_count += 1;
+                    last_error = Some(e.to_string());
+                }
+            }
+        }
+
+        debug!("[{source_id}] Webhook processing complete: {success_count} succeeded, {error_count} failed");
+
+        if error_count > 0 && success_count == 0 {
```
amansinghoriginal, line 758: Should we respect the `error_behavior` setting here as well for consistency?
For instance if the setting is `accept_and_log`, should we return 200?

components/sources/http/src/auth.rs
```diff
@@ -0,0 +1,515 @@
...
+
+/// Compute HMAC-SHA256
+fn compute_hmac_sha256(key: &[u8], data: &[u8]) -> Result<Vec<u8>> {
+    let mut mac =
+        Hmac::<Sha256>::new_from_slice(key).map_err(|e| anyhow!("HMAC-SHA256 key error: {e}"))?;
+    mac.update(data);
+    Ok(mac.finalize().into_bytes().to_vec())
+}
+
+/// Constant-time comparison to prevent timing attacks
```
amansinghoriginal, line 156: Should we use `subtle::ConstantTimeEq` here? It'd part of Rust ecosystem just like `hmac` and `sha2` crates already used in the project.

The reason is that this XOR loop could be optimized into a short-circuiting comparison by the compiler right?

components/sources/http/src/lib.rs
danielgerlag: fixed in commit https://github.com/drasi-project/drasi-core/pull/232/changes/06440e272d756355b67f5587b6fc6aa9b7f6201a

components/sources/http/src/auth.rs
danielgerlag: fixed in commit https://github.com/drasi-project/drasi-core/pull/232/changes/06440e272d756355b67f5587b6fc6aa9b7f6201a

PR #242, opened 2026-02-06
clippy fix: Separates the jq feature in drasi-middleware into two distinct features to support both system-installed libjq and compile-from-source workflows

ruokun-niu:
# Description

Separates the `jq` feature in `drasi-middleware` into two distinct features to support both system-installed libjq and compile-from-source workflows:

- **`jq`** — links against the system's libjq via pkg-config / `JQ_LIB_DIR`
- **`bundled-jq`** — compiles jq from source (enables `jq-rs/bundled`), implies `jq`

This allows downstream consumers (e.g. `drasi-server`) to link against a pre-installed libjq at build time, avoiding the fragile autotools-based build that `jq-rs/bundled` triggers in CI — especially on cross-platform runners where autotools dependencies are difficult to manage.

### Changes

- **`middleware/Cargo.toml`** — Split `jq` from `bundled-jq`; removed `features = ["bundled"]` from the `jq-rs` dependency; updated `all` feature to use `bundled-jq` (preserving docs.rs builds)
- **`lib/Cargo.toml`** — Added `middleware-bundled-jq` passthrough feature so consumers can opt into bundled builds at the lib level
- **`.github/workflows/ci-lint.yml`** — Improved the bundled jq build workaround for CI lint jobs

### Migration

Consumers that previously depended on the `jq` feature and expected jq to be compiled from source should switch to `bundled-jq`. Consumers that install libjq as a system dependency can continue using the `jq` feature directly.

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/pr-first-approval-label-run.yml
+++ b/.github/workflows/pr-first-approval-label-run.yml
@@ -1,0 +1,106 @@
+# Copyright 2026 The Drasi Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Stage 2 of the PR approval-labeling workflow. Runs in the base-repo context
+# (triggered by workflow_run) so it has a write GITHUB_TOKEN, even when the
+# original review was on a fork PR.
+#
+# The PR number is read from an artifact uploaded by Stage 1 and treated as
+# untrusted — review state is always re-queried from the GitHub API.
+
+name: Label PR on First Approval
+
+on:
+  workflow_run:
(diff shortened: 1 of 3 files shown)


middleware/Cargo.toml
```diff
@@ -17,6 +17,7 @@ default = []
 
 # Individual middleware features
 jq = ["dep:jq-rs"]
+bundled-jq = ["jq", "jq-rs/bundled"]
```
amansinghoriginal, line 20: Should we also update middleware/README.md which says that we compile jq from source?

middleware/Cargo.toml
```diff
@@ -46,7 +47,7 @@ hex = "0.4.3"            # Used by decoder
 urlencoding = "2.1.2"    # Used by decoder
 
 # Optional: only compiled when jq feature is enabled (always uses bundled)
```
amansinghoriginal: Optional Nit: Should we say `# Optional: only compiled when jq feature is enabled; use bundled-jq to compile from source` instead in the comment?

middleware/Cargo.toml
ruokun-niu: Good point, I'll update it

ruokun-niu: Updated

middleware/Cargo.toml
ruokun-niu: Updated

danielgerlag (approved):

PR #251, opened 2026-02-09
feat(mongodb): implement MongoDB source with change stream support

### LINKED ISSUE #226: MongoDB Source Plugin for Drasi-Lib
## Task: MongoDB Source Plugin for Drasi-Lib

**Difficulty:** Intermediate  
**Estimated scope:** New source plugin in `components/sources/`  
**Skills:** Rust, MongoDB, async programming, understanding drasi-lib source architecture

---

### Background

Drasi-Lib currently supports PostgreSQL as a database source via logical replication. MongoDB is one of the most popular NoSQL databases, especially in IoT and real-time analytics scenarios where Drasi excels.

MongoDB 3.6+ provides **Change Streams** – a native CDC (Change Data Capture) mechanism that allows applications to subscribe to real-time data changes. This maps naturally to Drasi's source architecture.

---

### What You'll Build

```
components/sources/mongodb/
├── Cargo.toml
├── README.md
├── src/
│   ├── lib.rs           # MongoSource struct, Source trait impl, builder
│   ├── config.rs        # MongoSourceConfig
│   ├── stream.rs        # Change Stream handling, resume token management
│   └── conversion.rs    # BSON -> SourceChange conversion, dot-notation hydration
```

**Components:**
- **MongoSource** – Implements `Source` trait, manages Change Stream lifecycle
- **Change Stream Consumer** – Subscribes to MongoDB changes with resume token support
- **BSON Converter** – Transforms MongoDB documents/deltas to Drasi `SourceChange` events
- **Dot-Notation Hydrator** – Converts `{"a.b.c": value}` to nested `{"a": {"b": {"c": value}}}`

---

## To Get Assigned

We appreciate your interest in contributing! Every contribution to Drasi is a gift, and we truly value you taking the time to help improve the project.

To help us find the right match for this task, please complete the following steps. This isn't a test to trip you up – it's designed to ensure you'll have a smooth experience and set you up for success.

---

### Step 0: Explore MongoDB Change Streams

Before diving into the code, familiarize yourself with MongoDB Change Streams:

1. **Read:** [MongoDB Change Streams Documentation](https://www.mongodb.com/docs/manual/changeStreams/)
2. **Understand:** How `watch()` works, what resume tokens are, and the structure of change events
3. **Optional:** Set up a local MongoDB replica set and experiment with Change Streams

**Key insight:** MongoDB Change Streams return **partial updates by default** (delta mode), not full documents. This is actually ideal for Drasi's architecture.

---

### Step 1: Answer These Questions

Once you've explored Change Streams and the codebase, please reply with answers to the following. Take your time – we'll review responses **48-72 hours after posting** to give everyone a fair chance.

**Note:** Please don't comment with just "I'd like to work on this" without completing these steps, as we won't be able to consider your request.

---

#### Q1: Code Exploration (Source Architecture)

Look at `components/sources/postgres/` in this repository.

- What is the main struct that implements the PostgreSQL source?
- What trait does it implement, and where is that trait defined?
- What are the key lifecycle methods (`start`, `stop`, `subscribe`)?
- What is `SourceBase` and how does the PostgreSQL source use it?

---

#### Q2: Architecture Understanding

In 3-4 sentences, explain the data flow: How would a MongoDB Change Stream event (e.g., an `update` operation) flow through Drasi to eventually update a continuous query's results?

Include these components in your explanation:
- Change Stream event
- `SourceChange`
- Continuous Query engine
- Result diff

---

#### Q3: Element ID Design

Given this MongoDB document in collection `users`:
```json
{
  "_id": ObjectId("507f1f77bcf86cd799439011"),
  "name": "Alice",
  "email": "alice@example.com",
  "status": "active"
}
```

**Q3a:** What should the Drasi element ID be? Explain your format choice.

**Q3b:** What if `_id` is a string instead of ObjectId: `{"_id": "user-alice-123", ...}`? What would the element ID be?

**Q3c:** What should the node label(s) be?

---

#### Q4: Partial Update Handling (Important!)

MongoDB Change Streams return partial updates by default:
```json
{
  "operationType": "update",
  "documentKey": {"_id": ObjectId("507f1f77bcf86cd799439011")},
  "updateDescription": {
    "updatedFields": {"address.city": "Seattle", "address.zip": "98101"},
    "removedFields": ["tempFlag"]
  }
}
```

**Q4a:** Why is this **dot-notation** (`address.city`) problematic for Drasi's property model?

**Q4b:** Write pseudocode or a Rust function signature for a `hydrate_dot_notation` function that converts:
```
Input:  {"address.city": "Seattle", "address.zip": "98101", "name": "Alice"}
Output: {"address": {"city": "Seattle", "zip": "98101"}, "name": "Alice"}
```

**Q4c:** How should `removedFields` be handled in the `SourceChange::Update`?

---

#### Q5: Error Handling

Look at how the PostgreSQL source handles errors in `components/sources/postgres/src/stream.rs`.

**Q5a:** What happens when the connection to the database is lost mid-stream?

**Q5b:** How would you handle a similar scenario in the MongoDB source when the Change Stream connection drops? What MongoDB feature helps with recovery?

---

### What We're Looking For

This isn't about perfection. We're looking for:

| Criteria | What it shows |
|----------|---------------|
| Correctness | You can navigate and understand the codebase |
| Clarity | You can explain technical concepts clearly |
| Attention to detail | You noticed the dot-notation challenge |
| Design thinking | You can reason about data transformations |
| Curiosity | You explored Change Streams before answering |

Even if you're not selected for this particular issue, your response helps us understand your interests – we may reach out for other opportunities.

---

## Shared Contract

**Critical**: This source MUST follow the exact contract below for compatibility with the MongoDB Bootstrap Provider (separate issue). Both components must produce identical element representations for the same MongoDB document.

### Element ID Format

```
{collection}:{_id_serialized}
```

**Serialization rules for `_id`:**

| `_id` Type | Serialization | Example |
|------------|---------------|---------|
| `ObjectId` | Hex string via `oid.to_hex()` | `users:507f1f77bcf86cd799439011` |
| `String` | Direct | `users:user-alice-123` |
| `Int32` | Decimal string | `orders:12345` |
| `Int64` | Decimal string | `events:9876543210` |
| Other | `bson.to_string()` | Fallback |

### Node Labels

```
labels = [collection_name]
```

- Case-sensitive, preserve original collection name
- Single label per node (the collection name)
- Example: Collection `users` -> Label `"users"`

### Property Mapping

- Map **top-level BSON fields** to Element properties
- **Exclude** `_id` from properties (already encoded in element ID)
- **Include** all other fields including MongoDB internals (`__v`, etc.)
- Nested documents -> `ElementValue::Object` (preserve structure)
- Arrays -> `ElementValue::List`

### BSON Type Conversion

| BSON Type | Drasi ElementValue | Conversion |
|-----------|-------------------|------------|
| `ObjectId` | `String` | `oid.to_hex()` |
| `String` | `String` | Direct |
| `Int32` | `Integer` | Direct |
| `Int64` | `Integer` | Direct |
| `Double` | `Float` | Direct |
| `Boolean` | `Bool` | Direct |
| `DateTime` | `String` | ISO 8601 UTC: `dt.try_to_rfc3339_string()` |
| `Document` | `Object` | Recursive conversion |
| `Array` | `List` | Recursive conversion |
| `Null` | `Null` | Direct |
| `Binary` | `String` | Base64 encoded |
| `Decimal128` | `String` | `d.to_string()` (preserve precision) |
| `Undefined` | `Null` | Treat as null |

### Change Stream to SourceChange Mapping

| Change Stream `operationType` | Drasi `SourceChange` |
|-------------------------------|----------------------|
| `insert` | `SourceChange::Insert` (full document) |
| `update` | `SourceChange::Update` (partial – hydrated delta) |
| `replace` | `SourceChange::Update` (full document) |
| `delete` | `SourceChange::Delete` (element reference only) |

### Dot-Notation Hydration (Source-Only)

For `update` operations, convert dot-notation keys to nested structure:

```
Input:  {"user.profile.name": "Alice", "status": "active"}
Output: {"user": {"profile": {"name": "Alice"}}, "status": "active"}
```

### Removed Fields Handling

Set removed fields to `null` in the update properties:

```
removedFields: ["address.city"]
-> properties: {"address": {"city": null}}
```

---

## Requirements for Implementation

Once assigned, your PR should include:

### Functional Requirements

- [ ] `MongoSource` struct implementing `Source` trait
- [ ] `MongoSourceConfig` with connection string, database, collections, etc.
- [ ] `MongoSourceBuilder` with fluent API (`with_connection_string()`, `with_database()`, etc.)
- [ ] Change Stream subscription with resume token support
- [ ] Conversion logic per shared contract above
- [ ] Dot-notation hydration for partial updates
- [ ] Removed fields -> null handling
- [ ] Lifecycle methods: `start()`, `stop()`, `subscribe()`
- [ ] Integration with `SourceBase` for dispatcher management

### Code Quality

- [ ] No `.unwrap()` – use proper error handling with `?` or explicit handling
- [ ] No `println!` – use `log` crate macros (`info!`, `error!`, `debug!`)
- [ ] Apache 2.0 license header on all new `.rs` files
- [ ] Code formatted with `cargo fmt`
- [ ] No clippy warnings (`cargo clippy`)
- [ ] DCO sign-off on all commits (`git commit -s`)

### Testing

- [ ] Unit tests for BSON -> ElementValue conversion
- [ ] Unit tests for dot-notation hydration
- [ ] Unit tests for element ID generation
- [ ] Integration tests using testcontainers with MongoDB replica set

### Documentation

- [ ] `README.md` in `components/sources/mongodb/` explaining:
  - What this source does
  - MongoDB prerequisites (replica set requirement, etc.)
  - Configuration options
  - Example usage with `DrasiLib`
  - Supported BSON types

---

## Resources

- **PostgreSQL source (reference implementation):** `components/sources/postgres/`
- **Source trait definition:** `lib/src/sources/mod.rs`
- **SourceBase:** `lib/src/sources/base.rs`
- **MongoDB Rust driver:** https://docs.rs/mongodb/latest/mongodb/
- **Change Streams docs:** https://www.mongodb.com/docs/manual/changeStreams/
- **BSON crate:** https://docs.rs/bson/latest/bson/

---

## Related Issues

- **MongoDB Bootstrap Provider:** #227 – Provides initial data snapshot (separate, parallelizable work)

---

## Questions?

If anything is unclear about the requirements or shared contract, please ask in the comments before starting implementation. We're here to help.

You can also join our [Discord community](https://aka.ms/drasidiscord) to discuss this issue, ask questions, or connect with other contributors and maintainers.

sagar-h007:
This PR introduces a new MongoDB source plugin for Drasi, enabling ingestion of MongoDB Change Streams as SourceChange events.

The implementation follows the existing source architecture and mirrors the behavior of the PostgreSQL source where applicable, while handling MongoDB-specific concerns such as partial updates and resume tokens.

### What’s included
- New drasi-source-mongodb crate registered in the workspace
- MongoSource implementation of the Source trait
- Change Stream consumer with:
-- Resume token support
-- Automatic reconnect with backoff
-- FullDocument lookup for updates
- BSON → ElementValue conversion covering common MongoDB types
- Dot-notation hydration for partial updates and removed fields
- Config validation for required properties
- Unit tests for:
-- BSON conversion
-- Dot-notation hydration
-- Config validation
- End-to-end integration test using testcontainers (ignored by default due to Docker requirement)

### Testing
- Unit tests:
```bash
cargo test -p drasi-source-mongodb --lib
```
- Integration test (requires Docker):
```bash
cargo test -p drasi-source-mongodb --test mongo_integration
```
### Notes
- MongoDB must be running as a replica set or sharded cluster (Change Streams requirement).
- The integration test is marked #[ignore] to avoid failures in environments without a Docker socket.
- This PR focuses on correctness and parity with the shared MongoDB source contract; further performance tuning can follow.

Fixes: #226

### DIFF
--- a/components/sources/mongodb/README.md
+++ b/components/sources/mongodb/README.md
@@ -1,0 +1,25 @@
+# MongoDB Source Plugin
+
+This source plugin enables Drasi to consume change events from a MongoDB Replcia Set or Sharded Cluster.
+
+## Features
+
+- Captures `insert`, `update`, `replace`, `delete` events via Change Streams.
+- Handles resume tokens for reliable stream consumption.
+- Converts MongoDB BSON types to Drasi Element types.
+- Supports hydration of dot-notation updates (e.g. `{"a.b": 1}` -> `{"a": {"b": 1}}`).
+
+## Configuration
+
+```yaml
+source:
+  type: mongodb
+  properties:
+    connection_string: "mongodb://localhost:27017"
+    database: "mydb"
+    collection: "mycollection"
+```
+
+## Prerequisites
+
+- MongoDB must be running as a Replica Set or Sharded Cluster. Standalone instances do not support Change Streams.
--- a/components/sources/mongodb/README.md
+++ b/components/sources/mongodb/README.md
@@ -1,25 +1,280 @@
-# MongoDB Source Plugin
+# MongoDB Source for Drasi
 
-This source plugin enables Drasi to consume change events from a MongoDB Replcia Set or Sharded Cluster.
+A source component that captures change streams from a MongoDB Replica Set and forwards them as Drasi Source Changes.
+
+## Overview
+
+The MongoDB Source connects to a MongoDB Replica Set (or Sharded Cluster) and watches for data changes (Inserts, Updates, Replaces, Deletes) using [Change Streams](https://www.mongodb.com/docs/manual/changeStreams/). It converts these database events into standardized Drasi Source Events (Nodes), allowing you to build continuous queries and reactions based on realtime MongoDB activity.
+
+## Architecture Flow
+
+```mermaid
+graph LR
+    Mongo[MongoDB Replica Set] -->|Change Stream| Source[MongoDB Source]
+    Source -->|SourceChange| SourceBase[Drasi SourceBase]
+    SourceBase -->|Distribute| Queries[Drasi Queries]
+    Queries -->|Results| Reactions[Drasi Reactions]
+```
+
+1.  **Change Stream**: The source establishes a persistent connection to the MongoDB changestream.
+2.  **Conversion**: BSON documents and change events are converted into Drasi Elements (`Node`).
+3.  **Dispatch**: Changes are pushed to the Drasi Runtime for processing.
+
+---
 
 ## Features
 
-- Captures `insert`, `update`, `replace`, `delete` events via Change Streams.
-- Handles resume tokens for reliable stream consumption.
-- Converts MongoDB BSON types to Drasi Element types.
-- Supports hydration of dot-notation updates (e.g. `{"a.b": 1}` -> `{"a": {"b": 1}}`).
+-   **Realtime Change Capture**: Consumes the MongoDB `watch()` stream for low-latency updates.
+-   **Multi-Collection Support**: Can watch one or multiple collections within a database.
+-   **Resume Tokens**: Automatically persists resume tokens to the Drasi State Store. If the source restarts, it resumes consumption exactly where it left off, preventing data loss.
+-   **Partial Updates**: Efficiently handles MongoDB `update` operations. It hydrates "dot notation" fields (e.g., `user.address.city`) into structured property maps.
+-   **Full Document Lookup**: Configured to use `updateLookup` to ensure the full document state is available when needed (though partial updates are preferred for performance when available).
+-   **Structure Handling**: Converts complex BSON types (ObjectId, Date, Embedded Docs) into Drasi standard types.
+-   **Resilience**: Handles connection drops and temporary unavailability with automatic reconnection and backoff.
 
 ## Configuration
 
+The source is configured via a YAML file.
+
+### basic_config.yaml
 ```yaml
-source:
-  type: mongodb
+kind: Source
+name: my-mongo-source
+spec:
+  kind: mongodb
   properties:
-    connection_string: "mongodb://localhost:27017"
+    connection_string: "mongodb://user:pass@localhost:27017/mydb?replicaSet=rs0"
     database: "mydb"
-    collection: "mycollection"
+    collections: 
+      - "users"
+      - "orders"
 ```
 
-## Prerequisites
+### Configuration Fields
 
-- MongoDB must be running as a Replica Set or Sharded Cluster. Standalone instances do not support Change Streams.
+| Field | Type | Required | Default | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| `connection_string` | String | Yes | - | Standard MongoDB connection URI. |
+| `database` | String | Optional* | - | The database name to watch. *Required if not specified in connection string.* |
+| `collection` | String | Optional | - | Single collection to watch (legacy mode). |
+| `collections` | List\<String\> | Optional | `[]` | List of collections to watch. **Recommended.** |
+| `username` | String | Optional | - | Overrides username in connection string. |
+| `password` | String | Optional | - | Overrides password in connection string. |
+| `pipeline` | List\<Document\> | Optional | `[]` | Custom aggregation pipeline validation/filtering stages passed to the change stream. |
+
+### Precedence Rules
+1.  **Database**: If specified in both `connection_string` (e.g., `mongodb://.../mydb`) AND `database` config, they **MUST** match. If they differ, the source will fail to start with an `Ambiguous configuration` error.
+2.  **Collections**: You must specify at least one collection via `collection` or `collections`. If both are provided, they are merged.
+
+---
+
+## Authentication & Secrets
+
+Do not hardcode credentials in your YAML files. Use Drasi's environment variable injection or secret management.
+
+### Credentials Precedence
+1.  **Config Fields**: `username` / `password` in yaml (Lowest priority, discouraged for secrets).
+2.  **Connection String**: Credentials embedded in the URI.
+3.  **Environment Variables**: `MONGODB_USERNAME` and `MONGODB_PASSWORD` environment variables (Highest priority).
+
+**Best Practice:**
+Inject credentials via Kubernetes Secrets or environment variables at runtime:
+
+```yaml
+# In your deployment manifest
+env:
+  - name: MONGODB_USERNAME
+    valueFrom:
+      secretKeyRef:
+        name: mongo-secrets
+        key: username
+  - name: MONGODB_PASSWORD
+    valueFrom:
+      secretKeyRef:
+        name: mongo-secrets
+        key: password
+```
+
+The source log automatically redacts connection strings and credentials.
+
+---
+
+## Multi-Collection Support
+
+You can watch multiple collections in the same database.
+
+**Configuration:**
+```yaml
+collections:
+  - "users"
+  - "products"
+  - "orders"
+```
+
+**Behavior:**
+-   The source opens a single stream on the **Database** and filters events by `ns.coll` (namespace collection).
+-   Events are tagged with their origin collection (see Mapping Model).
+-   Efficient resource usage compared to running multiple sources.
+
+---
+
+## Update Handling Semantics
+
+Drasi needs to know the *current state* of a node to evaluate queries. MongoDB provides flexible update formats; here is how they are handled:
+
+1.  **Insert / Replace**: The `full_document` is always used to build the Element properties.
+2.  **Update (Partial)**:
+    -   If `full_document` is available in the event, it is used.
+    -   If only `update_description` is available:
+        -   **Updated Fields**: Dot-notation fields (e.g., `{"address.city": "London"}`) are "hydrated" into nested objects (`{address: {city: "London"}}`).
+        -   **Removed Fields**: Fields listed in `removed_fields` are explicitly set to `Null` in the element properties.
+3.  **Delete**: No properties are preserved; the element is marked for deletion.
+
+---
+
+## Resume Token Behavior
+
+The source ensures "at-least-once" delivery semantics using MongoDB Resume Tokens.
+
+-   **Storage**: Resume tokens are stored in the configured Drasi **State Store** (e.g., Redis or Kubernetes specific store) under the key `resume_token`.
+-   **Restart**: On startup, the source checks the state store.
+    -   **Found**: It resumes the change stream from that specific token.
+    -   **Not Found**: It starts consuming from the *current moment* (Live).
+-   **Persistence**: The token is updated and persisted atomically with event dispatch.
+
+---
+
+## Mapping Model
+
+How MongoDB BSON translates to the Drasi Property Graph Model:
+
+### Node Identity
+Drasi Elements must have a unique ID.
+Format: `CollectionName:ObjectIdHex`
+
+Examples:
+-   Collection: `users`, `_id`: `ObjectId("5f...")` -> ID: `users:5f...`
+-   Collection: `settings`, `_id`: `"global"` -> ID: `settings:global`
+
+### Property Mapping
+| MongoDB (BSON) | Drasi (ElementValue) |
+| :--- | :--- |
+| `String` | String |
+| `Int32`, `Int64` | Integer |
+| `Double` | Float |
+| `Boolean` | Bool |
+| `Null` | Null |
+| `Document` | Object (Map) |
+| `Array` | List |
+| `ObjectId` | String (Hex representation) |
+| `DateTime` | String (RFC3339 format) |
+
+### Deletes
+A MongoDB `delete` event generates a Drasi `Change::Delete` event. The Node with the corresponding ID is removed from the query graph.
+
+---
+
+## Integration Testing
+
+### Requirements
+-   **Replica Set**: MongoDB Change Streams **REQUIRE** a Replica Set. They do not work on a standalone single node unless it is configured as a single-node replica set.
+
+### Running Implementation Tests
+This repository includes integration tests that spin up a Docker container.
+
+```bash
+# Run the integration tests (requires Docker)
+cargo test --test mongo_integration -- --ignored
+```
+
+*Note: The `--ignored` flag is used because these tests are slow/heavy and require Docker.*
+
+---
+
+## MongoDB Setup Guide (Local Development)
+
+To run this source locally, you need a MongoDB Replica Set.
+
+1.  **Start MongoDB Container**:
+    ```bash
+    docker run -d --rm -p 27017:27017 \
+      --name mongo-repl \
+      mongo:5.0 mongod --replSet rs0 --bind_ip_all
+    ```
+
+2.  **Initiate Replica Set**:
+    Inside the container (or via a client):
+    ```bash
+    docker exec -it mongo-repl mongosh --eval "rs.initiate()"
+    ```
+
+3.  **Verify**:
+    ```bash
+    docker exec -it mongo-repl mongosh --eval "rs.status().ok"
+    # Should return 1
+    ```
+
+Now change streams are enabled.
+
+---
+
+## End-to-End Example
+
+### 1. MongoDB Data (Input)
+Command executed in MongoDB:
+```javascript
+db.users.insert({
+  "_id": ObjectId("60b8d295f1d21034907d32a1"),
+  "name": "Alice",
+  "age": 30,
+  "skills": ["Rust", "MongoDB"]
+})
+```
+
+### 2. Drasi Source Configuration
+```yaml
+collections: ["users"]
+```
+
+### 3. Source Change Event (Internal)
+The source generates this internal event:
+
+```rust
+SourceChange::Insert {
+    element: Element::Node {
+        metadata: {
+            reference: "users:60b8d295f1d21034907d32a1",
+            labels: ["users"]
+        },
+        properties: {
+            "name": "Alice",
+            "age": 30,
+            "skills": ["Rust", "MongoDB"]
+        }
+    }
+}
+```
+
+### 4. Drasi Query (Cypher)
+```cypher
+MATCH (u:users) WHERE u.age >= 30 RETURN u.name
+```
+
+### 5. Query Output
+```json
+{
+  "u.name": "Alice"
+}
+```
+
+---
+
+## Limitations
+
+-   **Transaction Support**: The source does not currently expose transaction markers to downstream queries; events are processed individually.
+-   **Schema Validation**: The source does not enforce a schema; downstream queries must handle optional/missing fields if the MongoDB schema is loose.
+
+## Future Work
+
+-   [ ] Support for Sharded Cluster specific configurations (though currently works via mongos).
+-   [ ] Advanced filtering pipeline configuration via external files for complex logic.
--- a/components/sources/mongodb/src/config.rs
+++ b/components/sources/mongodb/src/config.rs
@@ -1,0 +1,51 @@
+use serde::{Deserialize, Serialize};
+use anyhow::Result;
+use mongodb::bson::Document;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct MongoSourceConfig {
+    pub connection_string: String,
+    pub database: String,
+    pub collection: String,
+    #[serde(default)]
+    pub pipeline: Option<Vec<Document>>,
+}
+
+impl MongoSourceConfig {
+    pub fn validate(&self) -> Result<()> {
+        if self.connection_string.is_empty() {
+             return Err(anyhow::anyhow!("Validation error: connection_string cannot be empty"));
+        }
+        if self.database.is_empty() {
+            return Err(anyhow::anyhow!("Validation error: database cannot be empty"));
+        }
+        if self.collection.is_empty() {
+            return Err(anyhow::anyhow!("Validation error: collection cannot be empty"));
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_config_validation() {
+        let config = MongoSourceConfig {
+            connection_string: "mongodb://localhost".to_string(),
+            database: "db".to_string(),
+            collection: "col".to_string(),
+            pipeline: None,
+        };
+        assert!(config.validate().is_ok());
+
+        let invalid_config = MongoSourceConfig {
+            connection_string: "".to_string(),
+            database: "db".to_string(),
+            collection: "col".to_string(),
+            pipeline: None,
+        };
+        assert!(invalid_config.validate().is_err());
+    }
+}
--- a/components/sources/mongodb/src/config.rs
+++ b/components/sources/mongodb/src/config.rs
@@ -5,8 +5,7 @@ use mongodb::bson::Document;
 #[derive(Debug, Clone, Serialize, PartialEq)]
 pub struct MongoSourceConfig {
     pub connection_string: String,
-    #[serde(default)]
-    pub database: Option<String>,
+    pub database: String,
     
     /// DEPRECATED: Use `collections` instead. This field will be removed in a future version.
     /// For backward compatibility, this field is automatically migrated to `collections` during deserialization.
@@ -33,8 +32,7 @@ impl<'de> Deserialize<'de> for MongoSourceConfig {
         #[derive(Deserialize)]
         struct ConfigHelper {
             connection_string: String,
-            #[serde(default)]
-            database: Option<String>,
+            database: String,
             #[serde(default)]
             collection: Option<String>,
             #[serde(default)]
@@ -88,27 +86,10 @@ impl MongoSourceConfig {
              return Err(anyhow::anyhow!("Validation error: connection_string cannot be empty"));
         }
         
-        let mut connection_db = None;
-        if let Ok(url) = url::Url::parse(&self.connection_string) {
-            let path = url.path().trim_start_matches('/');
-            if !path.is_empty() {
-                connection_db = Some(path.to_string());
-            }
-        }
-
-        match (&self.database, &connection_db) {
-            (Some(config_db), Some(conn_db)) => {
-                if config_db != conn_db {
-                    return Err(anyhow::anyhow!(
-                        "Ambiguous configuration: database specified in both config ('{}') and connection string ('{}') but they differ",
-                        config_db, conn_db
-                    ));
-                }
-            }
-            (None, None) => {
-                 return Err(anyhow::anyhow!("Validation error: database must be specified in either config or connection string"));
-            }
-            _ => {}
+        if self.database.is_empty() {
+            return Err(anyhow::anyhow!(
+                "Validation error: database cannot be empty"
+            ));
         }
         
         // Warn if deprecated field is still set in memory (not from deserialization)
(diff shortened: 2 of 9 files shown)


components/sources/mongodb/src/config.rs
```diff
@@ -0,0 +1,51 @@
+use serde::{Deserialize, Serialize};
+use anyhow::Result;
+use mongodb::bson::Document;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct MongoSourceConfig {
+    pub connection_string: String,
```
danielgerlag, line 7: How would one map the password from a secret store or env variable, separate form the rest of the connection sting?

components/sources/mongodb/src/config.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub database: String,
```
danielgerlag, line 8: If we have a connection string, would it not contain the database?

components/sources/mongodb/src/config.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub collection: String,
```
danielgerlag: Are we limiting the source to a single collection?

components/sources/mongodb/src/config.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[serde(default)]
+    pub pipeline: Option<Vec<Document>>,
```
danielgerlag, line 12: What is this for?

components/sources/mongodb/src/stream.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    async fn stream_loop(&mut self) -> Result<()> {
+        let client_options = ClientOptions::parse(&self.config.connection_string).await?;
+        let client = Client::with_options(client_options)?;
+        let db = client.database(&self.config.database);
+        let collection = db.collection::<Document>(&self.config.collection);
+
+        let mut options = ChangeStreamOptions::builder()
+            .full_document(Some(FullDocumentType::UpdateLookup))
```
danielgerlag: Drasi can handle partial updates, we do not need to retrieve the full document

components/sources/mongodb/tests/mongo_integration.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    async fn create_receiver(&self) -> Result<Box<dyn ChangeReceiver<SourceEventWrapper>>> {
+        Err(anyhow::anyhow!("Not implemented for mock"))
+    }
+}
+
+#[tokio::test]
+#[ignore]
+async fn test_mongo_source_integration() -> Result<()> {
```
danielgerlag, line 53: Let's do a full e2e tests with a query and reaction part of the chain

components/sources/mongodb/README.md
```diff
@@ -0,0 +1,25 @@
+# MongoDB Source Plugin
+
+This source plugin enables Drasi to consume change events from a MongoDB Replcia Set or Sharded Cluster.
+
+## Features
+
+- Captures `insert`, `update`, `replace`, `delete` events via Change Streams.
+- Handles resume tokens for reliable stream consumption.
```
danielgerlag: I can't find how it handles the persistence of this token across restarts?

components/sources/mongodb/README.md
```diff
@@ -0,0 +1,25 @@
+# MongoDB Source Plugin
+
+This source plugin enables Drasi to consume change events from a MongoDB Replcia Set or Sharded Cluster.
+
+## Features
```
danielgerlag, line 25: There is no description of how mongo documents are mapped to the Drasi graph model

components/sources/mongodb/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Captures `insert`, `update`, `replace`, `delete` events via Change Streams.
+- Handles resume tokens for reliable stream consumption.
+- Converts MongoDB BSON types to Drasi Element types.
+- Supports hydration of dot-notation updates (e.g. `{"a.b": 1}` -> `{"a": {"b": 1}}`).
+
+## Configuration
```
danielgerlag, line 35: There is  no description of how to configure / setup Mongo to support this

danielgerlag (changes requested):

components/sources/mongodb/src/conversion.rs
```diff
@@ -0,0 +1,244 @@
...
+            // Hydrate dot notation and handle removed fields
+            let hydrated_update = hydrate_dot_notation(&updated_fields, &removed_fields)?;
+            let properties = document_to_properties(&hydrated_update)?;
+            
+            Ok(Some(SourceChange::Update {
+                element: Element::Node {
+                    metadata,
+                    properties,
+                },
+            }))
```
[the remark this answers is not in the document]
danielgerlag: We should cover these cases with e2e integration tests

components/sources/mongodb/src/config.rs
sagar-h007: Right now the source consumes a standard MongoDB connection string, which allows credentials to be injected via environment variables or secret stores at deployment time.

I kept credentials inside the connection string to stay aligned with the MongoDB driver’s expectations and keep the initial API small. If the project prefers explicit secret fields or runtime secret resolution, I can extend the config in a follow-up PR.

components/sources/mongodb/src/config.rs
sagar-h007: The connection string may include a default database, but it’s optional. I kept database explicit to avoid ambiguity and to align with other sources that separate connection details from logical scope.

Happy to switch to parsing it from the connection string if that’s preferred.

components/sources/mongodb/src/config.rs
sagar-h007: Yes, intentionally scoped to a single collection for now to keep IDs and labels unambiguous. Multi-collection support would require additional semantics and feels like a good follow-up once the base source is stable.

components/sources/mongodb/src/config.rs
sagar-h007: This is an optional Change Stream aggregation pipeline.

It allows users to filter or transform events at the MongoDB level (for example, $match on specific operation types or fields) before they reach Drasi. This reduces unnecessary traffic and processing when only a subset of changes is relevant.

When omitted, the source defaults to an empty pipeline and consumes all change events for the collection.

components/sources/mongodb/src/stream.rs
sagar-h007: Drasi can already handle partial updates via updateDescription, so fetching the full document isn’t necessary here. I’ll remove FullDocumentType::UpdateLookup and rely on the hydrated delta for Update events.

Inserts and replaces will still use fullDocument as expected.

components/sources/mongodb/tests/mongo_integration.rs
sagar-h007: For this PR I focused on validating the MongoDB source itself (change stream consumption, conversion, dispatch). Wiring a full end-to-end test with a continuous query and reaction would require standing up more of the Drasi runtime and reaction components.

I’m happy to follow up with a separate PR that adds a full e2e test (MongoDB → Source → Query → Reaction), or extend this one if you prefer — let me know what you’d like the scope to be here.

components/sources/mongodb/README.md
sagar-h007: at the moment the resume token is only held in-memory for reconnects within the same process. It is not persisted across restarts yet.

The intent here was to match the PostgreSQL source behavior, where durable position tracking is handled via the bootstrap/checkpoint mechanism rather than the source itself. For MongoDB, persisting the resume token would likely integrate with a MongoDB bootstrap provider or shared checkpoint store so both initial snapshot and change stream consumption agree on the resume position.

I’ll update the README to clarify the current behavior and am happy to follow up with persistence support once the bootstrap provider side is in place.

components/sources/mongodb/src/config.rs
danielgerlag: Let's validate this with an integration test

components/sources/mongodb/src/config.rs
danielgerlag: It is a bit confusing and ambiguous to have some fields of the connection string be in either the connection string or another field.

components/sources/mongodb/src/config.rs
danielgerlag: We should support multiple collections.

components/sources/mongodb/src/config.rs
danielgerlag: Could you provide some examples and documentation on this feature?

components/sources/mongodb/tests/mongo_integration.rs
danielgerlag: Let's do full e2e tests with queries and reactions in the chain to validate this source fully.

components/sources/mongodb/README.md
danielgerlag: If the state store component is available, it should store the token there, so the MS-SQL source PR for an example.

components/sources/mongodb/src/config.rs
```diff
@@ -5,22 +5,60 @@ use mongodb::bson::Document;
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct MongoSourceConfig {
     pub connection_string: String,
-    pub database: String,
-    pub collection: String,
+    #[serde(default)]
+    pub database: Option<String>,
+    #[serde(default)]
+    pub collection: Option<String>,
```
danielgerlag: Why do we have `collection` and `collections`?

components/sources/mongodb/src/config.rs
```diff
@@ -5,22 +5,60 @@ use mongodb::bson::Document;
...
     #[serde(default)]
     pub pipeline: Option<Vec<Document>>,
+    #[serde(default)]
+    pub username: Option<String>,
+    #[serde(default)]
+    pub password: Option<String>,
 }
 
 impl MongoSourceConfig {
+    pub fn get_collections(&self) -> Vec<String> {
```
danielgerlag, line 20: What is the goal of this?

components/sources/mongodb/src/config.rs
```diff
@@ -5,22 +5,60 @@ use mongodb::bson::Document;
...
+            }
         }
-        if self.collection.is_empty() {
-            return Err(anyhow::anyhow!("Validation error: collection cannot be empty"));
+
+        match (&self.database, &connection_db) {
+            (Some(config_db), Some(conn_db)) => {
+                if config_db != conn_db {
+                    return Err(anyhow::anyhow!(
+                        "Ambiguous configuration: database specified in both config ('{}') and connection string ('{}') but they differ",
```
danielgerlag: We should just eliminate the one option completely rather than trying to allow it to be specified in two places.

components/sources/mongodb/src/stream.rs
```diff
@@ -60,85 +81,166 @@ impl ReplicationStream {
...
         Ok(())
     }
 
     async fn stream_loop(&mut self) -> Result<()> {
-        let client_options = ClientOptions::parse(&self.config.connection_string).await?;
+        let mut client_options = ClientOptions::parse(&self.config.connection_string).await?;
+
+        // Handle credentials override
+        let username = self.config.username.clone()
+            .or_else(|| std::env::var("MONGODB_USERNAME").ok());
```
danielgerlag: If the user of the library wants to use a particular env variable, they would do it, we should not assume env variable names here, lets remove these.

components/sources/mongodb/src/config.rs
```diff
@@ -5,22 +5,60 @@ use mongodb::bson::Document;
...
     pub fn validate(&self) -> Result<()> {
         if self.connection_string.is_empty() {
              return Err(anyhow::anyhow!("Validation error: connection_string cannot be empty"));
         }
-        if self.database.is_empty() {
-            return Err(anyhow::anyhow!("Validation error: database cannot be empty"));
+        
+        let mut connection_db = None;
+        if let Ok(url) = url::Url::parse(&self.connection_string) {
+            let path = url.path().trim_start_matches('/');
```
danielgerlag: Could we just use https://docs.rs/mongodb/latest/mongodb/options/struct.ClientOptions.html#method.parse

components/sources/mongodb/tests/end_to_end.rs
```diff
@@ -0,0 +1,485 @@
...
+        database: Some(db_name.to_string()),
+        collection: Some(collection_name.to_string()),
+        collections: vec![],
+        pipeline: None,
+        username: None,
+        password: None,
+    };
+
+    let source = MongoSource::new("e2e-source", config)?;
+    source.base.dispatchers.write().await.push(Box::new(dispatcher.clone()));
```
danielgerlag: Why are we not using DrasiLib to manage this?

components/sources/mongodb/tests/end_to_end.rs
```diff
@@ -0,0 +1,485 @@
...
+            return Ok(());
+        }
+        tokio::time::sleep(Duration::from_millis(500)).await;
+    }
+    Err(anyhow::anyhow!("Timeout waiting for query results"))
+}
+
+#[tokio::test]
+#[ignore]
+async fn test_end_to_end_basic_pipeline() -> Result<()> {
```
danielgerlag, line 125: This is not a valid end to end test. It should setup a DrasiLib instance, with a Mongo source, a query and an application reaction, not hacking the internals of DrasiLib. Please see the usage examples of DrasiLib: https://github.com/drasi-project/drasi-core/blob/main/examples/lib/builder/main.rs

components/sources/mongodb/tests/end_to_end.rs
```diff
@@ -0,0 +1,485 @@
...
+        self.results.read().await.clone()
+    }
+    
+    async fn clear_results(&self) {
+        self.results.write().await.clear();
+    }
+}
+
+#[async_trait::async_trait]
+impl ChangeDispatcher<SourceEventWrapper> for QueryDispatcher {
```
danielgerlag: We do not need this if you use DrasiLib

components/sources/mongodb/tests/end_to_end.rs
```diff
@@ -0,0 +1,485 @@
...
+use testcontainers::{core::{ContainerPort, WaitFor}, runners::AsyncRunner, GenericImage, ImageExt};
+
+// Validates that the source sends events to the query
+#[derive(Clone)]
+struct QueryDispatcher {
+    query: Arc<RwLock<ContinuousQuery>>,
+    results: Arc<RwLock<Vec<QueryPartEvaluationContext>>>,
+}
+
+impl QueryDispatcher {
```
danielgerlag: We do not need this if you use DrasiLib

components/sources/mongodb/tests/end_to_end.rs
```diff
@@ -0,0 +1,485 @@
...
+        username: None,
+        password: None,
+    };
+
+    let source = MongoSource::new("multi-source", config)?;
+    source.base.dispatchers.write().await.push(Box::new(dispatcher.clone()));
+    
+    let state_store = Arc::new(drasi_lib::state_store::MemoryStateStoreProvider::new());
+    let (status_tx, _status_rx) = tokio::sync::mpsc::channel(100);
+    source.initialize(drasi_lib::context::SourceRuntimeContext::new("multi-source", status_tx, Some(state_store))).await;
```
danielgerlag: Lets use DrasiLib instead of reinventing it in the test.

danielgerlag (changes requested):
Thank you for your contribution. I think there would be value in you running through the examples for DrasiLib under /examples/lib, to better understand how DrasiLib is used by a developer.

components/sources/mongodb/src/config.rs
sagar-h007: Good question. collection is kept for backward compatibility and simple single-collection setups, while collections allows explicitly watching multiple collections. If both are set, collections is used and collection is ignored. I can mark collection as deprecated to avoid confusion going forward if that sounds reasonable.

components/sources/mongodb/src/config.rs
sagar-h007: This helper normalizes configuration so the source always works with a single resolved list of collections. It merges collection (legacy single value) into collections when present and avoids duplicates, so downstream logic doesn’t need to handle both cases separately. This keeps runtime code simpler while maintaining backward compatibility.

components/sources/mongodb/src/config.rs
sagar-h007: I can switch this to `ClientOptions::parse` so we rely on the driver’s parser instead of manually extracting parts from the URL. That should be more robust and consistent with MongoDB’s expected connection string handling. I’ll update this section accordingly.

components/sources/mongodb/src/config.rs
sagar-h007: That makes sense. I agree it’s cleaner to avoid supporting the same option in two places. I’ll remove one of the sources of truth and keep a single canonical way to specify the database so we don’t need this ambiguity check at all.

components/sources/mongodb/tests/end_to_end.rs
sagar-h007: Got it, thanks for the clarification. I see what you mean about this not exercising the system the way a real consumer would. I’ll refactor the test to spin up a proper DrasiLib instance with a Mongo source, query, and reaction wired together, following the builder example you shared, instead of interacting with internals directly.

components/sources/mongodb/src/stream.rs
sagar-h007: I’ll remove the hardcoded environment variable fallbacks and rely only on values provided through the config so the library doesn’t assume naming conventions.

components/sources/mongodb/src/config.rs
danielgerlag: Why do we need the legacy single value collection?

components/sources/mongodb/src/config.rs
```diff
@@ -2,15 +2,20 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Clone, Serialize, PartialEq)]
 pub struct MongoSourceConfig {
     pub connection_string: String,
-    #[serde(default)]
-    pub database: Option<String>,
-    #[serde(default)]
+    pub database: String,
+    
+    /// DEPRECATED: Use `collections` instead. This field will be removed in a future version.
+    /// For backward compatibility, this field is automatically migrated to `collections` during deserialization.
```
danielgerlag: Why do we need backward compatibility if this was never released?

sagar-h007: Good point. Since this hasn’t been released yet, there’s no real need to preserve backward compatibility here. I’ll remove the deprecated database field and its migration logic and keep only the collections-based configuration so we maintain a single clean source of truth going forward.

components/sources/mongodb/README.md
danielgerlag: @sagar-h007

components/sources/mongodb/README.md
danielgerlag: @sagar-h007

danielgerlag (changes requested):
Hi @sagar-h007, thanks for the effort on this — a MongoDB source is a great addition to Drasi and it's clear you've put a lot of work into it. The overall structure is solid and the e2e tests using `DrasiLib` are exactly what we need. I have some feedback below that I'd like addressed before we can move forward.

---

### Architecture Gap: Missing `SourcePluginDescriptor`

Every source plugin in the project (postgres, mssql, etc.) implements `SourcePluginDescriptor` in a `descriptor.rs` file. This is required for the platform's plugin discovery, OpenAPI schema generation, and factory-based source creation. Without it, the MongoDB source can only be used programmatically via `DrasiLib` but won't be loadable by the platform plugin system. Please add a descriptor implementation following the existing sources as a reference.

### Previously Requested Changes Still Outstanding

A few items from earlier review rounds haven't been addressed yet in the code:

1. **`FullDocumentType::UpdateLookup` should be removed** (`stream.rs`). As discussed, Drasi handles partial updates natively — there's no need to fetch the full document on every update. This adds unnecessary latency. Please remove the `UpdateLookup` option and rely on `update_description` for partial updates.

2. **Pipeline feature needs examples and documentation.** I asked for concrete examples of how to use the `pipeline` config option — what a pipeline document looks like, common filtering patterns, etc. The README mentions the field in a table but doesn't show any practical usage. Could you add a dedicated section with at least 2-3 examples?

3. **README describes features/config that no longer exist in the code:**
   - It still references a `collection` (singular) config field, but the struct only has `collections` now.
   - It documents environment variable precedence (`MONGODB_USERNAME` / `MONGODB_PASSWORD`), but that behavior was removed from the code per earlier feedback.
   - It describes database-matching validation between the connection string and the `database` field, but no such validation exists in the current code.
   
   Please do a pass through the README and make sure it accurately reflects what the code actually does today.

4. **No description of how to configure MongoDB to support change streams.** The setup guide covers starting a local replica set, which is helpful, but please also document the minimum MongoDB version requirements and any server-side configuration needed (e.g., replica set requirement, permissions needed for the connecting user).

### Bugs

- **`test_mongo_source_multi_collection`** in `mongo_integration.rs` starts MongoDB without the `--replSet rs0` flag (no `with_cmd` call), unlike every other test in the file. Change streams require a replica set, so this test will fail. Please add `.with_cmd(vec!["mongod", "--replSet", "rs0", "--bind_ip_all"])` to match the other tests.

- **Auth test timing** (`test_mongo_source_with_auth`): After `source.start()` + a 2-second sleep, the test asserts `status == Error` for a source without credentials. Since `start()` spawns a background task, the auth failure may not have propagated within 2 seconds, making this test flaky. Consider polling the status with a timeout instead.

### Cleanup

- **Unused dependencies in `Cargo.toml`**: `uuid` and the standalone `bson` crate don't appear to be used anywhere (the code uses `mongodb::bson` throughout). Please remove them to keep compile times and audit surface minimal.

- **`pub base` on `MongoSource`**: This exposes internal implementation details. The integration tests reach into `source.base.dispatchers` directly, which couples them to internals. Consider making `base` private and adding accessor methods if needed.

- **Test duplication**: The replica set initialization sequence (start container → `replSetInitiate` → wait for PRIMARY) is copy-pasted across 7+ tests. A shared helper function would make these much easier to maintain.

---

### One more ask

Could you please **actually build and run the tests on your side** before the next push? Specifically:
- `cargo test -p drasi-source-mongodb --lib` for the unit tests
- `cargo test -p drasi-source-mongodb --test mongo_integration -- --ignored` for the integration tests
- `cargo test -p drasi-source-mongodb --test end_to_end -- --ignored` for the e2e tests

Also, I'd encourage you to **build a simple app** (even just a small `main.rs` example) that connects to a local MongoDB replica set, inserts/updates/deletes some documents, and observes the query results flowing through via `DrasiLib`. This kind of hands-on testing catches issues that automated tests sometimes miss — things like how partial updates behave with nested documents, what happens on reconnection, etc. The examples under `examples/lib/` show the pattern.

---

Thanks again for your work on this. The core logic (BSON conversion, change stream consumption, resume tokens, multi-collection support) is heading in the right direction. Looking forward to the next iteration!

danielgerlag:
@sagar-h007 closing this as stale, please reopen if you wish to address the remaining issues.

Issue #263, opened 2026-02-12
Surface stable Result Row IDs from the query engine


amansinghoriginal:
## Overview of feature request

The core query engine (`drasi-core`) does not surface a stable identifier for each result row in the diffs it emits. Internally, the engine does compute a row identity (`SolutionSignature` for non-aggregating queries, `ResultKey::GroupBy` for aggregating queries) for its own delta computation, but this identity is discarded at the `QueryPartEvaluationContext` boundary and is not propagated to consumers.

This forces every downstream reaction that needs to maintain a materialized view to independently reinvent row identity from the data values. Whether it is reactions in drasi-lib or the view service that's part of `drasi-platform`.

Every such reimplementation inherits the following correctness and robustness problems.

### The problem in drasi-core's `lib` crate

The `DrasiQuery` struct in `lib/src/queries/manager.rs` maintains a `current_results: Vec<serde_json::Value>` that tracks the live result set. It updates this via JSON deep-equality matching:

```rust
ResultDiff::Delete { data } => {
    result_set.retain(|item| item != data);          // O(n) scan, removes ALL matches
}
ResultDiff::Update { before, after, .. } => {
    if let Some(pos) = result_set.iter().position(|item| item == before) {
        result_set[pos] = after.clone();
    } else {
        warn!("UPDATE: Could not find exact match for before state, treating as remove+add");
        result_set.retain(|item| item != before);
        result_set.push(after.clone());
    }
}
```

**Issues:**
- **Duplicate-row collapse on delete:** `retain(|item| item != data)` removes *all* rows matching the value, not just one. If a query legitimately produces two rows with identical projected values (different underlying graph solutions, same output), deleting one silently deletes both.
- **Fragile update matching:** If the `before` snapshot doesn't exactly equal any stored value (e.g., floating-point precision drift, numeric representation differences in JSON), the match fails and falls into the `warn!` path, which itself is lossy.
- **O(n) per diff:** Every update/delete requires a linear scan of the entire result set.

### The problem in drasi-platform's view service

The view service (`query-container/view-svc`) maintains a temporal materialized view in MongoDB. Since it receives no row ID, it computes a SipHash-128 of the row content and uses it as the document key:

```rust
fn hash_values(values: &Map<String, Value>) -> [u8; 16] {
    let mut h = siphasher::sip128::SipHasher::new();
    for (key, value) in values {
        key.hash(&mut h);
        hash_value(value, &mut h);
    }
    h.finish128().as_bytes()
[...]

PR #265, opened 2026-02-12
Surface ResultRowId on QueryPartEvaluationContext

amansinghoriginal:
# Description

Surface the already-computed result row identity on `QueryPartEvaluationContext` (core engine's public result type).

The core query engine internally computes stable row identity for every result: `SolutionSignature` (u64 SpookyHash of matched graph elements) for non-aggregating queries and grouping key value hashes for aggregating queries. This identity is used for delta computation but discarded before results reach consumers, forcing every downstream materialized-view consumer to reinvent row identity via fragile content-hashing.

This changeset adds a `row_id: u64` field to each data-carrying variant of `QueryPartEvaluationContext` (`Adding`, `Updating`, `Removing`, `Aggregation`) and makes the already-computed identity available through the evaluation pipeline in `project_solution`.

### Changes

**Core engine (`core/src/evaluation/context.rs`)**
- Added `row_id: u64` field to `Adding`, `Updating`, `Removing`, and `Aggregation` variants
- Added `row_id()` accessor method
- Added `data_eq()` method for comparing all fields except `row_id` (used by test helpers)

**Pipeline threading (`core/src/query/continuous_query.rs`)**
- `project_solution`: maps `SolutionSignature` → `row_id` for non-aggregating queries, `after_grouping_hash` → `row_id` for aggregating queries
- `build_solution_changes`: sets placeholder `row_id: 0` on intermediate contexts
- `CollapsedAggregationResults`: sets `row_id` from the grouping key hash

**Evaluation pipeline (`core/src/evaluation/parts/mod.rs`)**
- Added `row_id: 0` placeholder to all ~15 construction sites in `evaluate()`
- Added `..` to destructuring patterns to accommodate the new field

**Downstream pattern matches**
- Updated 5 downstream files (`lib/src/queries/manager.rs`, `components/reactions/platform/src/types.rs`, `drasi-platform/query-container/query-host/src/api.rs`, `examples/query/temperature.rs`, `examples/query/process_monitor.rs`) to add `..` to pattern matches

**Tests**
- New `core/src/query/tests/row_id_tests.rs` with 7 tests verifying row_id behavior (non-zero on insert, stable across update/delete, different for different solutions, correct for aggregations)
- Updated ~130 construction sites in `core/src/evaluation/parts/tests/` with `row_id: 0`
- Updated ~170 construction sites in `shared-tests/src/use_cases/` with `row_id: 0`
- Added `contains_data()` helper to `shared-tests/src/use_cases/mod.rs` that uses `data_eq()` to compare results ignoring `row_id`
- Changed `result.contains(...)` → `contains_data(&result, ...)` in shared-tests where needed

### Key design decisions

1. **Direct field, not a wrapper**: `row_id` is a plain `u64` field on each variant, not a separate struct. This is pre-release with no external consumers.
2. **Derived `PartialEq` includes `row_id`**: Tests use `data_eq()` when they need to ignore row_id.
3. **No downstream integration yet**: `ResultDiff`, `current_results`, and platform `ResultChangeEvent` are future work.

## Type of change

Fixes: #263

### DIFF
--- a/core/src/evaluation/context.rs
+++ b/core/src/evaluation/context.rs
@@ -37,13 +37,16 @@ pub enum SideEffects {
 pub enum QueryPartEvaluationContext {
     Adding {
         after: QueryVariables,
+        row_signature: u64,
     },
     Updating {
         before: QueryVariables,
         after: QueryVariables,
+        row_signature: u64,
     },
     Removing {
         before: QueryVariables,
+        row_signature: u64,
     },
     Aggregation {
         before: Option<QueryVariables>,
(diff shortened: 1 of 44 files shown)


core/src/evaluation/context.rs
```diff
@@ -37,25 +37,80 @@ pub enum SideEffects {
 pub enum QueryPartEvaluationContext {
     Adding {
         after: QueryVariables,
+        row_id: u64,
```
danielgerlag: The term `id` feels like a misnomer for what this is, since the id will change when any field changes, should we rename it to something like `row_hash` or `row_signature`?

shared-tests/src/use_cases/before/mod.rs
```diff
@@ -118,12 +119,16 @@ pub async fn before_value(config: &(impl QueryTestConfig + Send)) {
...
-            ),
-        }));
+        assert!(contains_data(
+            &result,
+            &QueryPartEvaluationContext::Adding {
+                after: variablemap!(
+                  "amount" => VariableValue::from(json!(105.0)),
+                  "id" => VariableValue::from(json!("i1"))
+                ),
+                row_id: 0,
```
danielgerlag: why would these all be coming out as zero?

core/src/evaluation/context.rs
amansinghoriginal: Good point. Renamed to `row_signature` which pairs well with the internal `SolutionSignature` type.

shared-tests/src/use_cases/before/mod.rs
amansinghoriginal: These are dummy placeholders. `data_eq()` and `contains_data()` ignore the `row_signature` field when comparing results, so any value works. The actual runtime values are non-zero (verifiedb y the dedicated tests in row_signature_tests.rs).

danielgerlag (approved):

Issue #266, opened 2026-02-12
Unify Per-Query Index Storage into a Single Backend Instance


amansinghoriginal:
## Area for Improvement

The core query engine maintains 4 index types per query (`ElementIndex`, `ElementArchiveIndex`, `ResultIndex`, `FutureQueue`). In the current implementation, each index type is created as an **independent storage instance** via separate factory methods on the `IndexBackendPlugin` trait. For RocksDB, this means **3 separate `OptimisticTransactionDB` instances per query**. For Garnet/Redis, it means 3 separate connections to the same server.

This proposal is to **unify all 4 indexes of a given query into a single shared storage instance** per backend.

## Current Architecture

The `IndexBackendPlugin` trait creates each index independently:

```rust
#[async_trait]
pub trait IndexBackendPlugin: Send + Sync {
    async fn create_element_index(&self, query_id: &str) -> Result<Arc<dyn ElementIndex>, IndexError>;
    async fn create_archive_index(&self, query_id: &str) -> Result<Arc<dyn ElementArchiveIndex>, IndexError>;
    async fn create_result_index(&self, query_id: &str) -> Result<Arc<dyn ResultIndex>, IndexError>;
    async fn create_future_queue(&self, query_id: &str) -> Result<Arc<dyn FutureQueue>, IndexError>;
    fn is_volatile(&self) -> bool;
}
```

### RocksDB Backend

Each call opens a separate `OptimisticTransactionDB`, resulting in the following on-disk layout per query:

```
{base_path}/{query_id}/
    elements/       ← separate RocksDB instance (5-6 column families, 64 MB write buffer)
    aggregations/   ← separate RocksDB instance (3 column families, 32 MB write buffer)
    fqi/            ← separate RocksDB instance (2 column families, 1 MB write buffer)
```

Each instance has its own WAL, background compaction threads, flush threads, and block cache management. For N queries, this creates **3×N RocksDB instances**.

Additionally, `create_element_index()` and `create_archive_index()` in `plugin.rs` each open **separate** `RocksDbElementIndex` instances pointing at the same `elements/` directory, creating two independent DB handles to the same path.

The 11 column families across all instances have **no name collisions**:
- Element index: `elements`, `slots`, `inbound`, `outbound`, `partial`, `archive`
- Result index: `values`, `sorted-sets`, `metadata`
- Future queue: `fqueue`, `findex`

[...]

PR #272, opened 2026-02-14
Unify query index storage into a single instance

amansinghoriginal:
## Description

Unify per-query index storage so that all four index types (`ElementIndex`, `ElementArchiveIndex`, `ResultIndex`, `FutureQueue`) share a single backend instance instead of each opening its own.

**Before:** Each query opened 3 separate RocksDB databases (at `elements/`, `aggregations/`, `fqi/` subdirectories) and 3 separate Redis connections - one per index type. This tripled file descriptors, WAL files, background threads,
and network connections per query.

**After:** Each query opens a single RocksDB `OptimisticTransactionDB` (with all 11 column families) or a single Redis `MultiplexedConnection`, shared across all index types via `Arc`.

## What changed

**Pre-existing bugs fixed**

- **Garnet `FutureQueue::clear()` missed the main sorted set key:** The old pattern `fqi:{query_id}:*` only matched secondary index keys (e.g. `fqi:query1:0:12345`) but not the main sorted set `fqi:{query_id}`. Added an explicit `DEL` of the main key before pattern-clearing secondary keys. A `clear_removes_all` test is added to `shared-tests` and wired into all three backends.
- **query-perf RocksDB indexes were silently unused:** Both the element and result index `RocksDB` match arms created and cleared the index but returned `builder` without calling `.with_element_index()` / `.with_result_index()`, so benchmarks with `--element-index rocksdb` fell back to in-memory defaults.

**Core trait (`core/src/interface/index_backend.rs`)**

- Added `IndexSet` struct that groups the four index trait objects into one unit
- Replaced the four individual factory methods on `IndexBackendPlugin` (`create_element_index`, `create_archive_index`, `create_result_index`, `create_future_queue`) with a single `create_index_set()` method
- `IndexSet` is re-exported from `core::interface` so downstream crates can use it without reaching into submodules

**RocksDB backend (`components/indexes/rocksdb/`)**

- Added `open_unified_db()` — a public function that aggregates all column family descriptors from element, result, and future-queue modules and opens one `OptimisticTransactionDB` with all 11 CFs
- Changed `RocksDbElementIndex`, `RocksDbResultIndex`, and `RocksDbFutureQueue` constructors to accept an `Arc<OptimisticTransactionDB>` instead of opening their own database
- Made `RocksIndexOptions` `Clone + Copy` so it can be passed to the unified opener and then to individual constructors
- `element_index` and `archive_index` now share the same `Arc`, eliminating the previous bug where both opened separate DB instances at the same path

**Garnet/Redis backend (`components/indexes/garnet/`)**

- Added synchronous `new()` constructors on `GarnetElementIndex`, `GarnetResultIndex`, and `GarnetFutureQueue` that accept a `MultiplexedConnection`; the old `connect()` convenience constructors and `enable_archive()` mutator are removed since the trait signature change already breaks downstream callers
- `GarnetIndexProvider::new()` now takes a third parameter `enable_archive: bool`, and archive writes are actually wired through to `GarnetElementIndex`. Previously, archive was always disabled through the plugin path regardless of configuration.
- Added Redis hash tags (`{query_id}`) to all key patterns so that all keys for a given query hash to the same Redis Cluster slot. This is a prerequisite for cross-index `MULTI`/`EXEC` transactions in clustered deployments
- The plugin now opens a single connection and passes clones to each index

**Lib crate (`lib/src/indexes/`)**

- Removed the local `IndexSet` struct from `factory.rs` — it is now imported from `drasi_core::interface::IndexSet`
- Simplified `build_from_plugin()` from ~40 lines of per-index construction down to a single `plugin.create_index_set(query_id)` call

**Tests and tools**

- Updated `rocksdb/tests/scenario_tests.rs` to use `open_unified_db()` with a shared `Arc<OptimisticTransactionDB>` handle
- Updated `garnet/tests/scenario_tests.rs` to open one `redis::Client` connection and share it across all index constructors
- Updated `query-perf/src/main.rs` to use the new constructor signatures
- Updated query-perf Redis paths to share a single `MultiplexedConnection` (mirroring the RocksDB shared-handle pattern)
- Added `redis` as a dev-dependency for the garnet crate (needed by integration tests that open connections directly)

## Why this matters

1. **Resource reduction:** 3x fewer DB instances (RocksDB) or connections (Redis) per query. Fewer WAL files, background compaction threads, and file descriptors.
2. **Correctness:** Eliminates the bug where `element_index` and `archive_index` opened separate RocksDB instances at the same path.
3. **Foundation for transactions:** A shared backend instance is a prerequisite for cross-index atomic operations (checkpointing, outbox pattern, result-set snapshots).
4. **Simpler plugin API:** Plugin authors implement one method instead of four, and resource lifecycle is centralized.
5. **Cluster readiness (Garnet):** Hash tags ensure all keys for a query land on the same Redis Cluster shard, enabling `MULTI`/`EXEC` transactions.

## Not in scope
- Garnet `ResultIndex` atomicity improvements (`pipe().atomic()`, `WATCH`-based optimistic locking) are deferred to a follow-up. This commit establishes the prerequisite (shared connection, hash-tagged keys) but does not add transactional guarantees to `increment_value_count()` or `apply_sequence()`.

## Type of change

Fixes: #266

### DIFF
--- a/components/indexes/garnet/src/element_index/archive_index.rs
+++ b/components/indexes/garnet/src/element_index/archive_index.rs
@@ -30,9 +30,9 @@ use crate::{
 
 use super::GarnetElementIndex;
 
-/// Redis key structure:
+/// Redis key structure (hash-tagged for cluster compatibility):
 ///
-/// archive:{query_id}:{source_id}:{element_id} -> [element (sorted by ts)]
+/// archive:{<query_id>}:{source_id}:{element_id} -> [element (sorted by ts)]
 #[async_trait]
 impl ElementArchiveIndex for GarnetElementIndex {
     async fn get_element_as_at(
@@ -130,7 +130,7 @@ impl ElementArchiveIndex for GarnetElementIndex {
 
     async fn clear(&self) -> Result<(), IndexError> {
         self.connection
-            .clear(format!("archive:{}:*", self.query_id))
+            .clear(format!("archive:{{{}}}:*", self.query_id))
             .await
     }
 }
--- a/components/indexes/garnet/src/result_index.rs
+++ b/components/indexes/garnet/src/result_index.rs
@@ -31,31 +31,22 @@ use redis::{aio::MultiplexedConnection, AsyncCommands};
 
 use crate::{storage_models::StoredValueAccumulator, ClearByPattern};
 
-/// Redis key structure:
-/// ari:{query_id}:{set_id} -> {value}
-/// ari:{query_id}:{set_id} -> [sorted value]
-/// ari:{query_id}:{set_id}:{value} -> count
+/// Redis key structure (hash-tagged for cluster compatibility):
+/// ari:{<query_id>}:{set_id} -> {value}
+/// ari:{<query_id>}:{set_id} -> [sorted value]
+/// ari:{<query_id>}:{set_id}:{value} -> count
 pub struct GarnetResultIndex {
     query_id: Arc<str>,
     connection: MultiplexedConnection,
 }
 
 impl GarnetResultIndex {
-    pub async fn connect(query_id: &str, url: &str) -> Result<Self, IndexError> {
-        let client = match redis::Client::open(url) {
-            Ok(client) => client,
-            Err(e) => return Err(IndexError::connection_failed(e)),
-        };
-
-        let connection = match client.get_multiplexed_async_connection().await {
-            Ok(con) => con,
-            Err(e) => return Err(IndexError::connection_failed(e)),
-        };
-
-        Ok(GarnetResultIndex {
+    /// Create a new GarnetResultIndex from a shared connection.
+    pub fn new(query_id: &str, connection: MultiplexedConnection) -> Self {
+        GarnetResultIndex {
             query_id: Arc::from(query_id),
             connection,
-        })
+        }
     }
 }
 
@@ -69,7 +60,7 @@ impl AccumulatorIndex for GarnetResultIndex {
     ) -> Result<Option<ValueAccumulator>, IndexError> {
         let mut con = self.connection.clone();
         let set_id = get_hash_key(owner, key);
-        let key = format!("ari:{}:{}", self.query_id, set_id);
+        let key = format!("ari:{{{}}}:{}", self.query_id, set_id);
         let result = match con.get::<String, Option<StoredValueAccumulator>>(key).await {
             Ok(v) => v,
             Err(e) => return Err(IndexError::other(e)),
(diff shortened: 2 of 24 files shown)


components/indexes/garnet/src/element_index/archive_index.rs
```diff
@@ -30,9 +30,9 @@ use crate::{
 
 use super::GarnetElementIndex;
 
-/// Redis key structure:
+/// Redis key structure (hash-tagged for cluster compatibility):
```
danielgerlag, line 33: Nice!

components/indexes/rocksdb/src/plugin.rs
```diff
@@ -31,21 +31,61 @@
...
+/// # Arguments
+///
+/// * `path` - Base directory for RocksDB data files
+/// * `query_id` - Unique identifier for the query
+/// * `options` - RocksDB index options (archive, direct I/O)
+///
+/// # Directory Structure
+///
+/// Data is stored at `{path}/{query_id}/` (single unified directory).
+pub fn open_unified_db(
```
danielgerlag, line 57: Would we want to expand this to include the queues between sources, queries and reactions?

components/indexes/garnet/src/result_index.rs
```diff
@@ -111,7 +102,7 @@ impl AccumulatorIndex for GarnetResultIndex {
 
     async fn clear(&self) -> Result<(), IndexError> {
         self.connection
-            .clear(format!("ari:{}:*", self.query_id))
+            .clear(format!("ari:{{{}}}:*", self.query_id))
```
danielgerlag, line 105: This PR hash-tags the metadata:{query_id}:sequence and metadata:{query_id}:source_change_id read/write paths, but clear() only deletes ari:{<query_id>}:* — the two metadata: keys are orphaned

danielgerlag (approved):
Looks really good, approving with suggestions. This would also break drasi-platform, so we'd need a follow-up PR in `drasi-platform` to make it compatible.

components/indexes/rocksdb/src/plugin.rs
amansinghoriginal: Yes, that's the plan. There are 2 categories, both of which would live as new column families in this same unified DB:

Source -> Query: The query processor will atomically checkpoint the sequence number of the last successfully processed SourceChange in its index. On restart, it can resume from that point instead of re-bootstrapping. The checkpoint write must be atomic with the element/result index writes for the same change.

Query-> Reaction: Two structures here — (1) a materialized view of the current result set, maintained atomically with result updates, and (2) a fixed-size outbox buffer of result diffs for at-least-once delivery to reactions. Both are written atomically with the result index update in the same transaction.

components/indexes/garnet/src/result_index.rs
amansinghoriginal: The metadata is not cleared for rocks, garnet or in-memory indexes.
In `drasi-platform` it tracks the sequence number the results.

If we clear it, then the position in the stream will be lost.

danielgerlag: What about when a query or source is deleted?

PR #273, opened 2026-02-14
feat(expr): support collect_list aggregation function

### LINKED ISSUE #217: Add `COLLECT_LIST` aggregation function for GQL queries
## Summary
Add support for the `COLLECT_LIST` aggregation function in GQL queries, as defined in the GQL ISO Standard (ISO/IEC 39075:2024).

## Background
Drasi currently supports the `collect()` aggregation function for Cypher queries, which collects values into a list. The GQL ISO Standard defines an equivalent function called `COLLECT_LIST` (part of Feature GF10: "Advanced aggregate functions: general set functions").

Currently:
- Cypher: `collect()` is supported
- GQL: `COLLECT_LIST` is **not** supported

## GQL ISO Standard Reference
From ISO/IEC 39075:2024, Section 20.9 `<aggregate function>`:

```
<general set function type> ::=
    AVG
  | COUNT
  | MAX
  | MIN
  | SUM
  | COLLECT_LIST
  | STDDEV_SAMP
  | STDDEV_POP
```

**Behavior per spec:**
- Returns a list containing all non-null values from the group
- If input is empty, returns an empty list `[]` (not NULL)
- Supports `DISTINCT` and `ALL` set quantifiers
- Result type is a list whose element type matches the input expression type

## Example Usage (after implementation)

```gql
MATCH (p:Product)<-[:REVIEWED]-(r:Review)
RETURN p.name, COLLECT_LIST(r.rating) AS all_ratings
```

## Acceptance Criteria
- [ ] `COLLECT_LIST` function is registered in the GQL function registry
- [ ] GQL queries using `COLLECT_LIST` work correctly
- [ ] Tests added for GQL `COLLECT_LIST` usage
- [ ] Documentation updated (if applicable)

## Hints
- Look at how Cypher registers its `collect` function
- Look at how other aggregation functions are registered in GQL
- The underlying implementation may already exist - find it and reuse it

Standing-Man:
# Description

## Summary
This PR introduces support for the `collect_list` aggregation function in GQL queries, in accordance with the GQL ISO Standard specification. The `collect_list` function aggregates values from multiple rows into a list, preserving duplicates and maintaining the input order where applicable.

## Implementation Details
- Reused the Cypher `collect` function implementation for the GQL `collect_list` function
- Handles `null` values correctly (ignores them)
- Supports collecting any variable value type
- Works seamlessly with all storage backends (`in-memory`, `Garnet`, `RocksDB`) without requiring backend-specific changes

## Test

- in-memory backend test
```
cargo test --package shared-tests --lib -- in_memory::collect_list_aggregation
```

- garnet backend test
```
cargo test --package drasi-index-garnet --test scenario_tests -- collect_list_aggregation
```

- RockDB backend test
```
cargo test --package drasi-index-rocksdb --test scenario_tests -- collect_list_aggregation
```

## Type of change

- Closes #217.


functions-gql/src/lib.rs
```diff
@@ -118,6 +118,10 @@ fn register_aggregation_functions(registry: &FunctionRegistry) {
     registry.register_function("count", Function::Aggregating(Arc::new(Count {})));
     registry.register_function("min", Function::Aggregating(Arc::new(Min {})));
     registry.register_function("max", Function::Aggregating(Arc::new(Max {})));
+    registry.register_function(
+        "collect_list",
+        Function::Aggregating(Arc::new(CollectList {})),
```
ruokun-niu: Since `collect_list` in GQL is equivalent to the `collect` function in Cypher (https://neo4j.com/docs/cypher-manual/current/appendix/gql-conformance/analogous-cypher/), is it possible to just use the existing mplementation for Collect instead of implementing the function again?

Standing-Man: Thank you for the helpful suggestions. I’ve reused the Cypher collect function implementation for the GQL collect_list function.

Standing-Man: Should we unify the error messages, or add a specific CollectList identifier to distinguish them? I’d be glad to hear your perspective.

ruokun-niu: I think it would be great if we can distinguish the error message. Thanks!

Standing-Man: I’ve implemented the change to distinguish the error messages on new commit. I’d appreciate it if you could take another look when you have time.

ruokun-niu (approved):

PR #277, opened 2026-02-15
fix(garnet): resolve race condition in increment_value_count

### LINKED ISSUE #276: Race Condition in GarnetResultIndex::increment_value_count
## Description
I identified a potential race condition in the `GarnetResultIndex::increment_value_count` method within [drasi-core/components/indexes/garnet/src/result_index.rs](file:///Users/avii/Documents/LFX/DRASI/drasi-core/components/indexes/garnet/src/result_index.rs).

The original implementation performed a `GET` followed by an `INCR` or `DEL` operation based on the retrieved value. Since these operations were executed separately without a transaction or locking mechanism (like `WATCH`), concurrent calls to [increment_value_count](file:///Users/avii/Documents/LFX/DRASI/drasi-core/components/indexes/garnet/src/result_index.rs#189-234) could lead to race conditions.

For example:
1. Two concurrent requests read the same counter value (e.g., 1).
2. Request A decrements it (1 -> 0) and issues a `DEL`.
3. Request B increments it (1 -> 2) and issues an `INCR`.
4. Depending on execution order, the key effectively disappears or has the wrong value, leading to lost updates.

There was a `TODO` comment in the code acknowledging this:
```rust
//todo: add WATCH to ensure that the value is not changed between the get and the del
```

## Proposed Solution
Replace the multi-step `GET` + logic + `INCR`/`DEL` approach with a Redis Lua script (`EVAL`). Lua scripts in Redis are executed atomically, ensuring that the read-modify-write cycle cannot be interrupted by other commands.

## Implementation Details
- Introduce a Lua script that:
    1. Gets the current value (defaulting to 0 if missing).
    2. Calculates the new value.
    3. If the new value is 0, deletes the key and removes it from the sorted set.
    4. Otherwise, updates the key and the sorted set.

Av1ralS1ngh:
# Description

This commit replaces the non-atomic check-and-set logic in `GarnetResultIndex::increment_value_count` with a Lua script (`EVAL`). This ensures atomic updates for counters and prevents race conditions where concurrent operations might overwrite data or incorrectly delete keys.

This specifically addresses the `TODO` comment in the code regarding missing concurrency control.

## Type of change

- [x] This pull request fixes a bug in Drasi and has an approved issue (issue link required).
- [ ] This pull request adds or changes features of Drasi and has an approved issue (issue link required).
- [ ] This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Drasi (issue link optional).

Fixes #276

### DIFF
--- a/components/indexes/garnet/src/result_index.rs
+++ b/components/indexes/garnet/src/result_index.rs
@@ -197,27 +197,37 @@ impl LazySortedSetStore for GarnetResultIndex {
         let set_key = format!("ari:{}:{}", self.query_id, set_id);
         let val_key = format!("{}:{}", set_key, value.0);
 
-        let current_count = match con.get::<&str, Option<isize>>(&val_key).await {
-            Ok(v) => v.unwrap_or(0),
-            Err(e) => return Err(IndexError::other(e)),
-        };
+        // Lua script to atomically update the value count and the sorted set
+        // If the new count is 0, we delete the value key and remove from the set
+        // Otherwise, we update the value key and ensure the value is in the set
+        let script = redis::Script::new(r#"
+            local val_key = KEYS[1]
+            local set_key = KEYS[2]
+            local delta = tonumber(ARGV[1])
+            local score = ARGV[2]
 
-        let mut pipeline = redis::pipe();
+            local current = redis.call("GET", val_key)
+            if not current then current = 0 end
+            current = tonumber(current)
+            local new_val = current + delta
 
-        if (current_count + delta) == 0 {
-            pipeline.del::<&str>(&val_key).ignore();
-            pipeline.zrem::<&str, f64>(&set_key, value.0).ignore();
-            //todo: add WATCH to ensure that the value is not changed between the get and the del
-        } else {
-            pipeline.incr::<String, isize>(val_key, delta).ignore();
-            pipeline
-                .zadd::<String, f64, f64>(set_key, value.0, value.0)
-                .ignore();
-        }
+            if new_val == 0 then
+                redis.call("DEL", val_key)
+                redis.call("ZREM", set_key, score)
+            else
+                redis.call("INCRBY", val_key, delta)
+                redis.call("ZADD", set_key, score, score)
+            end
+        "#);
 
-        if let Err(err) = pipeline.query_async::<_, ()>(&mut con).await {
-            return Err(IndexError::other(err));
-        }
+        script
+            .key(val_key)
+            .key(set_key)
+            .arg(delta)
+            .arg(value.0)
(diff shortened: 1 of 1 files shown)


danielgerlag (changes requested):
Hey @Av1ralS1ngh, thanks for taking the time to tackle this race condition — great catch on the non-atomic `GET` → check → `INCR`/`DEL` pattern, and the issue write-up was very clear. The Lua script approach is a solid solution in standard Redis, and the logic in your script is correct.

Unfortunately, our storage backend targets [Garnet](https://github.com/microsoft/garnet) (a Redis-compatible server from Microsoft), which does not support Lua scripting (`EVAL`/`EVALSHA`) out of the box. This means the Lua script would fail at runtime in production, even though it passes CI (which runs against vanilla Redis).

### Suggested Alternative

The good news is we can solve this without Lua by leaning on the fact that `INCRBY` is already atomic in Redis/Garnet — it increments and returns the new value in a single command. The idea is to do the increment unconditionally first, then clean up if the result hits zero:

```rust
let mut con = self.connection.clone();
let set_key = format!("ari:{}:{}", self.query_id, set_id);
let val_key = format!("{}:{}", set_key, value.0);

// INCRBY is atomic and returns the new value
let new_count: isize = con.incr(&val_key, delta)
    .await
    .map_err(IndexError::other)?;

if new_count == 0 {
    // Clean up: delete the counter key and remove from sorted set
    let mut pipeline = redis::pipe();
    pipeline.del::<&str>(&val_key).ignore();
    pipeline.zrem::<&str, f64>(&set_key, value.0).ignore();
    pipeline.query_async::<_, ()>(&mut con)
        .await
        .map_err(IndexError::other)?;
} else {
    // Ensure the value exists in the sorted set
    con.zadd::<&str, f64, f64>(&set_key, value.0, value.0)
        .await
        .map_err(IndexError::other)?;
}
```

**Why this eliminates the original race condition:** The core problem was that a `GET` followed by a separate `INCR` or `DEL` allowed concurrent callers to interleave and corrupt the count. With this approach, `INCRBY` atomically reads + increments + returns the new value in one shot, so the count itself is never racy.

There is a small residual window where another caller could increment the key between our `INCRBY` returning 0 and the `DEL`, but this is a benign edge case — it would leave a stale key that gets naturally cleaned up on the next operation.

### Summary of Requested Changes

1. Replace the Lua script with the `INCRBY`-first approach shown above
2. Keep the rest of your changes (key formatting, error handling, etc.)

Thanks again for the contribution — really appreciate you spotting the TODO and following through on it. Let me know if you have any questions!

PR #280, opened 2026-02-15
feat: Sqlite source

danielgerlag:
# SQLite Source

## Overview

`drasi-source-sqlite` is a protocol/local source that owns an embedded SQLite connection and emits Drasi `SourceChange` events for row-level `INSERT`, `UPDATE`, and `DELETE` activity.

Key capabilities:

- Real-time CDC using `rusqlite` `preupdate_hook`
- Transaction-aware dispatch (`commit_hook` flush, `rollback_hook` discard)
- Savepoint-aware buffering for partial rollback support
- Optional table filtering and explicit table key configuration
- Optional REST API for table CRUD and transactional batch operations
- Works with file-backed and in-memory SQLite databases

## Quick Start

```rust
use drasi_source_sqlite::{SqliteSource, TableKeyConfig, RestApiConfig};

let source = SqliteSource::builder("sqlite-source")
    .with_path("data/example.db")
    .with_table_keys(vec![TableKeyConfig {
        table: "sensors".to_string(),
        key_columns: vec!["id".to_string()],
    }])
    .with_rest_api(RestApiConfig {
        host: "127.0.0.1".to_string(),
        port: 9100,
    })
    .build()?;
```

## SQL Handle API

Use `source.handle()` to execute statements through the source-owned SQLite connection:

```rust
let handle = source.handle();
handle.execute("CREATE TABLE sensors(id INTEGER PRIMARY KEY, name TEXT, temp REAL)").await?;
handle.execute("INSERT INTO sensors(id, name, temp) VALUES (1, 'sensor-a', 31.5)").await?;

handle.transaction(|tx| async move {
    tx.execute("INSERT INTO sensors(id, name, temp) VALUES (2, 'sensor-b', 30.0)").await?;
    tx.execute("UPDATE sensors SET temp = 32.1 WHERE id = 1").await?;
    Ok(())
}).await?;
```

## REST API (Optional)

When `with_rest_api()` is configured:

- `GET /health`
- `GET /api/tables`
- `GET /api/tables/{table}`
- `GET /api/tables/{table}/{id}`
- `POST /api/tables/{table}`
- `PUT /api/tables/{table}/{id}`
- `DELETE /api/tables/{table}/{id}`
- `POST /api/batch` (single SQLite transaction for all operations)

## Configuration Notes

- `tables: Option<Vec<String>>` filters monitored tables. `None` means all user tables.
- `table_keys` defines element ID columns per table; if omitted, table PKs are detected from `PRAGMA table_info`.
- If no configured/detected keys exist, rowid is used as fallback for streaming events.

## Testing

From this crate directory:

```bash
cargo test
cargo test -- --ignored --nocapture
cargo clippy --all-targets -- -D warnings
```

The integration tests validate:

- handle-driven INSERT/UPDATE/DELETE flow
- REST CRUD and transactional batch behavior
- bootstrap delivery into query results
- multi-table query routing

## Limitations

- CDC only captures writes executed through the source-owned connection.
- DDL (`CREATE/ALTER/DROP TABLE`) is not emitted as `SourceChange`.
- BLOB values are emitted as base64-encoded strings.

### DIFF
--- a/components/bootstrappers/sqlite/src/lib.rs
+++ b/components/bootstrappers/sqlite/src/lib.rs
@@ -1,0 +1,290 @@
+// Copyright 2025 The Drasi Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use anyhow::Result;
+use async_trait::async_trait;
+use base64::Engine;
+use drasi_core::models::{
+    Element, ElementMetadata, ElementPropertyMap, ElementReference, ElementValue, SourceChange,
+};
+use drasi_lib::bootstrap::{BootstrapContext, BootstrapProvider, BootstrapRequest};
+use drasi_lib::channels::BootstrapEventSender;
+use log::{info, warn};
+use ordered_float::OrderedFloat;
+use rusqlite::types::ValueRef;
+use rusqlite::{Connection, OpenFlags};
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+/// Per-table key configuration for stable element IDs.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct TableKeyConfig {
+    pub table: String,
+    pub key_columns: Vec<String>,
+}
+
+/// SQLite bootstrap provider.
+#[derive(Clone)]
+pub struct SqliteBootstrapProvider {
+    path: Option<String>,
+    tables: Option<Vec<String>>,
+    table_keys: Vec<TableKeyConfig>,
+}
+
+impl SqliteBootstrapProvider {
+    pub fn builder() -> SqliteBootstrapBuilder {
+        SqliteBootstrapBuilder::new()
+    }
+
+    fn key_columns_for_table(&self, conn: &Connection, table: &str) -> Result<Vec<String>> {
+        if let Some(cfg) = self.table_keys.iter().find(|item| item.table == table) {
+            return Ok(cfg.key_columns.clone());
+        }
+        detect_primary_key(conn, table)
+    }
+}
+
+/// Builder for [`SqliteBootstrapProvider`].
+pub struct SqliteBootstrapBuilder {
+    path: Option<String>,
+    tables: Option<Vec<String>>,
+    table_keys: Vec<TableKeyConfig>,
+}
+
+impl SqliteBootstrapBuilder {
+    fn new() -> Self {
+        Self {
+            path: None,
+            tables: None,
+            table_keys: Vec::new(),
+        }
+    }
+
+    pub fn with_path(mut self, path: impl Into<String>) -> Self {
+        self.path = Some(path.into());
+        self
+    }
+
+    pub fn in_memory(mut self) -> Self {
+        self.path = None;
+        self
+    }
+
+    pub fn with_tables(mut self, tables: Vec<String>) -> Self {
+        self.tables = Some(tables);
+        self
+    }
+
+    pub fn with_table_keys(mut self, table_keys: Vec<TableKeyConfig>) -> Self {
+        self.table_keys = table_keys;
+        self
+    }
+
+    pub fn build(self) -> SqliteBootstrapProvider {
+        SqliteBootstrapProvider {
+            path: self.path,
+            tables: self.tables,
+            table_keys: self.table_keys,
+        }
+    }
+}
+
+#[async_trait]
+impl BootstrapProvider for SqliteBootstrapProvider {
+    async fn bootstrap(
+        &self,
+        request: BootstrapRequest,
+        context: &BootstrapContext,
+        event_tx: BootstrapEventSender,
+        _settings: Option<&drasi_lib::config::SourceSubscriptionSettings>,
+    ) -> Result<usize> {
+        info!("Starting SQLite bootstrap for query '{}'", request.query_id);
+
+        let changes = {
+            let Some(path) = &self.path else {
+                warn!("SQLite bootstrap skipped for in-memory database");
+                return Ok(0);
+            };
+
+            let conn = Connection::open_with_flags(path, OpenFlags::SQLITE_OPEN_READ_ONLY)?;
+            let tables = resolve_tables(&conn, self.tables.as_ref())?;
+            let mut changes = Vec::new();
+
+            for table in tables {
+                if !request.node_labels.is_empty() && !request.node_labels.contains(&table) {
+                    continue;
+                }
+
+                let key_columns = self.key_columns_for_table(&conn, &table)?;
+                let rows = read_table_rows(&conn, &table)?;
+
+                for row in rows {
+                    let element_id = generate_element_id(&table, &row, &key_columns);
+                    let mut properties = ElementPropertyMap::new();
+                    for (name, value) in row {
+                        properties.insert(&name, value);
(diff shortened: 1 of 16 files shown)


ruokun-niu (approved):

components/bootstrappers/sqlite/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+                    for (name, value) in row {
+                        properties.insert(&name, value);
+                    }
+
+                    let labels: Arc<[Arc<str>]> = vec![Arc::<str>::from(table.as_str())].into();
+                    let element = Element::Node {
+                        metadata: ElementMetadata {
+                            reference: ElementReference::new(&context.source_id, &element_id),
+                            labels,
+                            effective_from: chrono::Utc::now()
```
amansinghoriginal: `effective_from` uses nanoseconds here, but the source uses milliseconds

amansinghoriginal: I think postgres had this same issue, so likely the agent has cloned the behavior. We should fix that too, else agents will keep duplicating this pattern.

components/sources/sqlite/src/rest_api.rs
```diff
@@ -0,0 +1,469 @@
...
+        _ => return false,
+    }
+    chars.all(|c| c == '_' || c.is_ascii_alphanumeric())
+}
+
+fn quote_ident(value: &str) -> String {
+    format!("\"{}\"", value.replace('"', "\"\""))
+}
+
+fn value_to_sql_literal(value: &Value) -> String {
```
amansinghoriginal: Nit: The escaping here is technically correct for SQLite (no backslash escapes), but consider adding a parameterized execution variant to SqliteCommand using rusqlite's ? placeholders.

amansinghoriginal: Defense-in-depth, and avoids needing to reason about escaping correctness.

components/sources/sqlite/src/thread.rs
```diff
@@ -0,0 +1,678 @@
...
+        pk_columns,
+    })
+}
+
+fn column_name(schema: &TableSchema, index: i32) -> String {
+    schema
+        .columns
+        .get(index as usize)
+        .cloned()
+        .unwrap_or_else(|| format!("col_{index}"))
```
amansinghoriginal: Nit: If this fallback ever triggers, it would silently produce wrong property names in SourceChange events, leading to incorrect query results. Consider logging a warning here so it's not invisible

components/sources/sqlite/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        self.base
+            .set_status_with_event(
+                ComponentStatus::Starting,
+                Some("Starting SQLite source".to_string()),
+            )
+            .await?;
+
+        let (command_tx, command_rx) = mpsc::unbounded_channel::<SqliteCommand>();
+        *self.command_tx.write().await = Some(command_tx.clone());
+        let (event_tx, mut event_rx) = mpsc::unbounded_channel::<thread::ChangeEvent>();
```
amansinghoriginal, line 322: This event channel is unbounded. In Channel dispatch mode, downstream backpressure from SourceBase can stall the dispatcher task, letting this buffer grow.

amansinghoriginal: Hmm, but making it bounded with blocking_send on the OS thread would propagate backpressure to SQLite writes

amansinghoriginal: Worth a quick discussion in team sync if this is tradeoff is acceptable or current approach is fine.

components/bootstrappers/sqlite/src/lib.rs
amansinghoriginal, line 1: No unit tests for bootstrapper?

amansinghoriginal: I guess this pattern is followed from the postgres source.

components/bootstrappers/sqlite/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+        )),
+    }
+}
+
+fn generate_element_id(
+    table: &str,
+    values: &[(String, ElementValue)],
+    key_columns: &[String],
+) -> String {
+    if key_columns.is_empty() {
```
amansinghoriginal: This is a bug.

When key_columns is empty this concats all column values as the element ID.

But source falls back to using `rowid` from the preupdate hook.

amansinghoriginal: Does this not mean that bootstrap and streaming produce different element IDs for the same row?

Query engine will treat them as different elements, right?

amansinghoriginal (changes requested):
There might be a couple of bugs. Please review.

components/bootstrappers/sqlite/src/lib.rs
ruokun-niu: @amansinghoriginal Can you review this PR https://github.com/drasi-project/drasi-core/pull/284? It contains the fix to the Postgres Source

danielgerlag: good catch

danielgerlag: Fixed — changed `timestamp_nanos_opt()` to `timestamp_millis()` to match the source.

components/sources/sqlite/src/lib.rs
danielgerlag: Agreed this is worth discussing at team sync. Bounding the channel would back-pressure SQLite writes, which may or may not be desirable depending on the use case.

components/bootstrappers/sqlite/src/lib.rs
danielgerlag: Fixed — bootstrap now does `SELECT rowid, *` and falls back to rowid when no key columns are configured, matching the source behavior exactly. Added cross-verification unit tests that confirm bootstrap and source produce identical element IDs for the same row.

components/bootstrappers/sqlite/src/lib.rs
danielgerlag: Added 9 unit tests for the bootstrapper covering: element ID generation (with PKs, composite PKs, rowid fallback), row-to-node conversion, timestamp handling, and cross-verification that bootstrap and source produce identical element IDs.

components/sources/sqlite/src/rest_api.rs
danielgerlag: Done — refactored all REST API handlers to use rusqlite `?` placeholder parameterized queries. Removed `value_to_sql_literal` entirely. Added `SqliteParam` enum and `ExecuteParameterized`/`QueryRowsParameterized` command variants to thread the params safely across the channel.

components/sources/sqlite/src/thread.rs
danielgerlag: Added log::warn! when the column index is out of range and the fallback triggers.

components/sources/sqlite/src/config.rs
```diff
@@ -0,0 +1,261 @@
...
+    fn default() -> Self {
+        Self {
+            host: default_rest_host(),
+            port: default_rest_port(),
+        }
+    }
+}
+
+fn default_rest_host() -> String {
+    "0.0.0.0".to_string()
```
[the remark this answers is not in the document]
danielgerlag: Addressed — Changed default REST API bind address from 0.0.0.0 to 127.0.0.1.

components/sources/sqlite/src/rest_api.rs
```diff
@@ -0,0 +1,509 @@
...
+    let router = Router::new()
+        .route("/health", get(health_check))
+        .route("/api/tables", get(list_tables))
+        .route("/api/tables/:table", get(list_rows).post(insert_row))
+        .route(
+            "/api/tables/:table/:id",
+            get(get_row).put(update_row).delete(delete_row),
+        )
+        .route("/api/batch", post(batch_operations))
+        .with_state(state);
```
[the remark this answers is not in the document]
danielgerlag: Addressed — Changed default REST API bind address from 0.0.0.0 to 127.0.0.1.

agentofreality (approved):

PR #283, opened 2026-02-17
fix: Preserve datetime types from PostgreSQL

ruokun-niu:
# Description

### Problem

**PostgreSQL datetime values collapsed to strings**: PostgreSQL temporal values (`timestamp`, `timestamptz`) were being converted to `ElementValue::String` throughout the entire data pipeline — both the CDC (WAL replication) path and the bootstrap (initial snapshot) path. This caused Cypher aggregation functions like `max()` and `min()` to reject these values with `InvalidArgument` errors, since they received plain strings instead of comparable datetime objects.

#### 1. Added datetime variants to `ElementValue`

**File:** `core/src/models/element_value.rs`

- Added `LocalDateTime(NaiveDateTime)` and `ZonedDateTime(DateTime<FixedOffset>)` variants to the `ElementValue` enum
- Implemented all bidirectional conversions:
  - `ElementValue` → `VariableValue` (maps `LocalDateTime` → `VariableValue::LocalDateTime`, `ZonedDateTime` → `VariableValue::ZonedDateTime`)
  - `VariableValue` → `ElementValue` (both by-ref and by-value `TryInto` impls)
  - `ElementValue` → `serde_json::Value` (serializes as strings)
- Added 15 unit tests covering roundtrip conversions, JSON serialization, and type safety

#### 2. Fixed PostgreSQL CDC path

**Files:** `components/sources/postgres/src/types.rs`, `components/sources/postgres/src/stream.rs`

- Added `PostgresValue::to_element_value()` method that directly maps `Timestamp` → `ElementValue::LocalDateTime` and `TimestampTz` → `ElementValue::ZonedDateTime`, bypassing the lossy JSON intermediate step
- Added `PostgresValue::is_null()` helper
- Replaced the lossy `to_json()` + `convert_json_to_element_value()` pipeline in `convert_insert` and `convert_update` with direct `to_element_value()` calls
- Added 11 unit tests for all `PostgresValue` → `ElementValue` conversions

#### 3. Fixed PostgreSQL bootstrap path

**Files:** `components/bootstrappers/postgres/src/postgres.rs`, `components/sources/postgres/src/decoder.rs`

- Updated `decode_column_value_text()` for OIDs 1114 (`timestamp`) and 1184 (`timestamptz`) to parse timestamp strings into `LocalDateTime`/`ZonedDateTime` with multiple format attempts and a string fallback
- Updated bootstrapper row decoding to emit `ElementValue::LocalDateTime` / `ElementValue::ZonedDateTime` instead of `ElementValue::String`
- Added 13 unit tests covering various timestamp formats, edge cases (whitespace, missing timezone), and fallback behavior

#### 4. Updated storage backends

**Files:** `components/indexes/garnet/src/storage_models/stored_value.rs`, `components/indexes/rocksdb/src/storage_models/stored_value.rs`

- Added serialization arms for `LocalDateTime` and `ZonedDateTime` in both Garnet and RocksDB `StoredValueContainer` conversions (serialized as strings for protobuf compatibility)

#### 5. Updated middleware

**Files:** `middleware/src/decoder/mod.rs`, `middleware/src/parse_json/mod.rs`

- Added `"LocalDateTime"` and `"ZonedDateTime"` arms to `get_element_value_type_name()` to handle the new variants

### Testing

| Location | New Tests | Coverage |
|---|---|---|
| `core/src/models/element_value.rs` | 15 | Roundtrip conversions, JSON serialization, type safety |
| `components/sources/postgres/src/types.rs` | 11 | `to_element_value()` for all PostgresValue variants |
| `components/sources/postgres/src/decoder.rs` | 13 | `decode_column_value_text()` timestamp formats and fallbacks |

All existing tests continue to pass (663 core, 76 shared, 85 lib).

## Type of change

Fixes: #issue_number


components/indexes/garnet/src/storage_models/stored_value.rs
```diff
@@ -159,6 +159,12 @@ impl From<&ElementValue> for StoredValueContainer {
             ElementValue::Object(o) => StoredValueContainer {
                 value: Some(StoredValue::Object(o.into())),
             },
+            ElementValue::LocalDateTime(dt) => StoredValueContainer {
+                value: Some(StoredValue::String(dt.to_string())),
```
danielgerlag: We should expand the protobuf types to handle this data type natively.
Initially it will be a datetime, then serialize as a string, then when we deserialize it later, it will stay a string and we would have lost the datetime type info.

components/indexes/rocksdb/src/storage_models/stored_value.rs
```diff
@@ -159,6 +159,12 @@ impl From<&ElementValue> for StoredValueContainer {
             ElementValue::Object(o) => StoredValueContainer {
                 value: Some(StoredValue::Object(o.into())),
             },
+            ElementValue::LocalDateTime(dt) => StoredValueContainer {
+                value: Some(StoredValue::String(dt.to_string())),
```
danielgerlag: We should expand the protobuf types to handle this data type natively.
Initially it will be a datetime, then serialize as a string, then when we deserialize it later, it will stay a string and we would have lost the datetime type info.

We should add tests to cover this scenario.

core/src/models/element_value.rs
danielgerlag: So if ElementValue is serialized and then deserialized, datetimes will become strings?
Are there any scenarios (besides the index storage) where this can happen?
What if we swapped out the channels for persistent streams that need to serialize / deserialize source changes, how would they know to deserialize back to a datetime?

lib/src/queries/manager.rs
```diff
@@ -690,8 +690,21 @@ impl Query for DrasiQuery {
                                                         result_set.push(after_json);
                                                     }
                                                 }
-                                                QueryPartEvaluationContext::Aggregation { .. }
-                                                | QueryPartEvaluationContext::Noop => {}
+                                                QueryPartEvaluationContext::Aggregation { before, after, .. } => {
```
danielgerlag: The new aggregation handling uses position-based JSON equality to find and replace before values in result_set. This works correctly for the bootstrap path where 
  QueryPartEvaluationContext::Aggregation provides before/after as QueryVariables that get converted to JSON.

  However, grouping_keys from the Aggregation context are discarded — matching relies entirely on full JSON object equality. If floating-point rounding or serialization
  order changes between the stored before and the incoming before, the match will fail and hit the fallback retain + push path. This is defensive (no data loss) but
  could produce duplicate rows in edge cases.

components/sources/postgres/src/types.rs
```diff
@@ -193,3 +240,127 @@ mod base64 {
         result
     }
 }
+
+#[cfg(test)]
+mod tests {
```
danielgerlag, line 263: Could we add tests somewhere to assert that bootstrap data conversion matches CDC data conversions?

danielgerlag (changes requested):

components/indexes/garnet/src/storage_models/stored_value.rs
ruokun-niu: Good point. Thanks for pointing it out

lib/src/queries/manager.rs
ruokun-niu: Good point. I'll update to use `grouping_keys` for matching

components/sources/postgres/src/types.rs
ruokun-niu: added

components/indexes/garnet/src/storage_models/stored_value.rs
ruokun-niu: added

core/src/models/element_value.rs
ruokun-niu: Today this isn't an issue — `ElementValue` doesn't derive `Serialize`/`Deserialize`, the JSON `From`/`Into` impls are one-directional (import vs export, never chained), and channels are in-memory. Postgres is also the only source producing native datetime types; all others treat datetimes as strings.

If persistent channels are introduced, we could derive `Serialize`/`Deserialize` on `ElementValue` with serde's tagged enum representation (`#[serde(tag = "type", content = "value")]`). This would produce self-describing JSON like `{"type": "LocalDateTime", "value": "2024-06-15T10:30:45"}` that roundtrips correctly, and works with any serde-compatible format. No changes needed in this PR.

danielgerlag: ElementValue has a `Into<serde::Value>` implementation where this creates a lossy issue

danielgerlag, line 178: In `impl From<&ElementValue> for serde_json::Value`
 - We convert date/time values to string

In `impl From<&serde_json::Value> for ElementValue`
 - We convert string to string but don't know if it should be a date/time

This is an issue that we cannot ignore today, as anyone could end up serializing / deserializing via this path and loose the data type, and be completely unaware of this.

components/sources/postgres/src/decoder.rs
```diff
@@ -710,7 +710,19 @@ pub fn decode_column_value_text(
         }
         1114 | 1184 => {
             // timestamp, timestamptz
-            Ok(ElementValue::String(Arc::from(text)))
+            // Try parsing as NaiveDateTime first (timestamp without tz)
+            if let Ok(dt) = NaiveDateTime::parse_from_str(text.trim(), "%Y-%m-%d %H:%M:%S%.f") {
+                Ok(ElementValue::LocalDateTime(dt))
+            } else if let Ok(dt) = NaiveDateTime::parse_from_str(text.trim(), "%Y-%m-%d %H:%M:%S") {
+                Ok(ElementValue::LocalDateTime(dt))
```
danielgerlag, line 723: When PostgreSQL sends a timestamptz value like "2024-06-15 10:30:45" (no explicit offset, which can happen when the server timezone is UTC and it omits  +00), the NaiveDateTime parse succeeds first and produces ElementValue::LocalDateTime — even though the column is timestamptz (OID 1184).

 Why this matters — CDC/bootstrap parity breaks

 The CDC path (binary replication) in decoder.rs:544-562 handles OIDs separately:

```rust
1184 => {
       // timestamptz — always produces PostgresValue::TimestampTz(DateTime<Utc>)
       // which becomes ElementValue::ZonedDateTime
   }
```

  So the same column produces:

   - CDC: ElementValue::ZonedDateTime (always, because binary decoding knows it's timestamptz)
   - Bootstrap: ElementValue::LocalDateTime (when offset text is missing)

lib/src/queries/manager.rs
```diff
@@ -690,8 +718,20 @@ impl Query for DrasiQuery {
                                                         result_set.push(after_json);
                                                     }
                                                 }
-                                                QueryPartEvaluationContext::Aggregation { .. }
-                                                | QueryPartEvaluationContext::Noop => {}
+                                                QueryPartEvaluationContext::Aggregation { before, after, grouping_keys, .. } => {
+                                                    let after_json = convert_query_variables_to_json(after);
+                                                    if let Some(before) = before {
+                                                        let before_json = convert_query_variables_to_json(before);
+                                                        if let Some(pos) = find_by_grouping_keys(&result_set, &before_json, grouping_keys) {
```
danielgerlag: Would it be more efficient to leverage the row_signature instead of serializing and comparing json on every change?

danielgerlag (changes requested):

core/src/models/element_value.rs
ruokun-niu: Good point. We can serialize datetime values as a tagged JSON envelope (e.g. something like `{"__drasi_type": "ZonedDateTime", "value": "2024-06-15T10:30:45+01:00"}`). On deserialization, we check for the tag and reconstruct the correct variant.

Any thoughts @danielgerlag? I think an alternative is to do a try-parse on deserialization. In `impl From<&serde_json::Value> for ElementValue`,  we attempt `DateTime::parse_from_rfc3339(s)` and `NaiveDateTime::parse_from_str(s)` before falling back to `String`. Personally I am leaning towards the first option as the second options risks false positives

lib/src/queries/manager.rs
ruokun-niu: yep, good point. Will update this

components/sources/postgres/src/decoder.rs
ruokun-niu: updated

lib/src/queries/manager.rs
ruokun-niu: I refactored `current_results` from `Vec<serde_json::Value>` to `HashMap<u64, serde_json::Value>` keyed by row_signature. This removes find_by_grouping_keys entirely and gives us O(1) lookups in both the bootstrap and CDC paths.

danielgerlag (changes requested):
## Review Summary

Two issues to address before merging:

### 1. Bug: `convert_variable_value_to_json` missing datetime arms (regression)

In `lib/src/queries/manager.rs`, the `convert_variable_value_to_json` function has no explicit match arms for `VariableValue::LocalDateTime` or `VariableValue::ZonedDateTime`. They fall through to the `_ => serde_json::Value::String(format!("{value:?}"))` catch-all, which produces Debug-format output like `"LocalDateTime(2024-06-15 10:30:45)"` instead of a clean ISO-8601 string.

**Before this PR**, timestamps flowed as `ElementValue::String` → `VariableValue::String` → matched by the `String` arm → clean output.
**After this PR**, timestamps flow as `ElementValue::LocalDateTime` → `VariableValue::LocalDateTime` → catch-all → `format!("{value:?}")` → wrapped in variant name.

This is a regression in the JSON output sent to reactions and the REST API. Add explicit match arms:

```rust
VariableValue::LocalDateTime(dt) => serde_json::Value::String(dt.to_string()),
VariableValue::ZonedDateTime(dt) => serde_json::Value::String(dt.datetime().to_rfc3339()),
```

### 2. Consider the `__drasi_type` tag naming

The `__drasi_type` tagged envelope in `ElementValue` JSON serialization works, but it lives in user-data space (graph node properties come from user databases). A couple of considerations:

- If a user stores a JSON object with a `__drasi_type` key, deserialization will misinterpret it as a datetime value.
- The rest of the codebase uses serde's `#[serde(tag = "type")]` / `#[serde(tag = "kind")]` — this is the only place that embeds the product name into a data-format field.

An alternative worth considering: use a more generic key (e.g., `_type`) and namespace the *values* instead (e.g., `"drasi.LocalDateTime"`). Not a blocker, but worth discussing for long-term API surface consistency.

### Everything else looks good

- ✅ Aggregation result storage with `HashMap<u64, serde_json::Value>` keyed by `row_signature` — clean and efficient
- ✅ Storage backend protobuf datetime roundtripping (tags 7/8) — properly preserves type info
- ✅ Bootstrap/CDC parity — OID 1184 always produces `ZonedDateTime`, matching binary CDC path
- ✅ Good test coverage (39 new tests)

danielgerlag (changes requested):
## Review Summary

The datetime preservation work is solid — good protobuf roundtripping (tags 7/8), CDC/bootstrap parity, tagged JSON envelopes, and 39 new tests. Nice iteration on previous feedback.

### Blocking: Behavioral regression in JSON output

`lib/src/queries/manager.rs` is not changed in this PR, but `convert_variable_value_to_json` (line 78) has a catch-all arm:

```rust
_ => serde_json::Value::String(format!("{value:?}"))
```

Today this is harmless because no code produces `VariableValue::LocalDateTime`/`ZonedDateTime`. **After this PR merges**, the Postgres source will start producing those variants, and they will hit this catch-all, producing Debug-format output like `"LocalDateTime(2024-06-15 10:30:45)"` instead of clean ISO strings.

Please add explicit match arms (either in this PR or as an immediate companion PR):

```rust
VariableValue::LocalDateTime(dt) => serde_json::Value::String(dt.to_string()),
VariableValue::ZonedDateTime(dt) => serde_json::Value::String(dt.datetime().to_rfc3339()),
```

### Everything else looks good

- ✅ Protobuf storage roundtrip with native tags 7/8 in Garnet and RocksDB
- ✅ Bootstrap/CDC parity — OID 1184 always produces `ZonedDateTime`
- ✅ `_type` + `drasi.*` tagged JSON envelope for lossless serialization
- ✅ Direct `to_element_value()` in CDC path, bypassing lossy JSON intermediate
- ✅ Good test coverage across all layers
- ✅ Manager.rs aggregation changes correctly separated to PR #333

danielgerlag (approved):

components/indexes/garnet/src/storage_models/stored_value.rs
```diff
@@ -12,169 +12,354 @@
...
+            None => ElementValue::Null,
+            Some(StoredValue::Bool(b)) => ElementValue::Bool(b),
+            Some(StoredValue::Float(f)) => ElementValue::Float(f.into()),
+            Some(StoredValue::Integer(i)) => ElementValue::Integer(i),
+            Some(StoredValue::String(s)) => ElementValue::String(Arc::from(s.as_str())),
+            Some(StoredValue::List(l)) => {
+                ElementValue::List(l.values.into_iter().map(|v| v.into()).collect())
+            }
+            Some(StoredValue::Object(o)) => ElementValue::Object(o.into()),
+            Some(StoredValue::LocalDateTime(micros)) => {
```
agentofreality, line 215: This defaults to Unix Epoch if the from_timestamp_micros fails. Do we know that we will never get a value out if range? If not should error or at least log a warning.

components/indexes/rocksdb/src/storage_models/stored_value.rs
```diff
@@ -12,169 +12,354 @@
...
+            None => ElementValue::Null,
+            Some(StoredValue::Bool(b)) => ElementValue::Bool(b),
+            Some(StoredValue::Float(f)) => ElementValue::Float(f.into()),
+            Some(StoredValue::Integer(i)) => ElementValue::Integer(i),
+            Some(StoredValue::String(s)) => ElementValue::String(Arc::from(s.as_str())),
+            Some(StoredValue::List(l)) => {
+                ElementValue::List(l.values.into_iter().map(|v| v.into()).collect())
+            }
+            Some(StoredValue::Object(o)) => ElementValue::Object(o.into()),
+            Some(StoredValue::LocalDateTime(micros)) => {
```
agentofreality, line 215: This defaults to Unix Epoch if the from_timestamp_micros fails. Do we know that we will never get a value out if range? If not should error or at least log a warning.

agentofreality (changes requested):

components/indexes/rocksdb/src/storage_models/stored_value.rs
ruokun-niu: These micros values always come from a valid NaiveDateTime/DateTime] that we serialized via .timestamp_micros() on the write path, so the reverse conversion will always succeed.

components/indexes/garnet/src/storage_models/stored_value.rs
ruokun-niu: These micros values always come from a valid NaiveDateTime/DateTime] that we serialized via .timestamp_micros() on the write path, so the reverse conversion will always succeed.

agentofreality (approved):

Issue #191, opened 2026-01-24
Improve ci-lint.yml workflow: add caching, fix dependencies, align triggers


amansinghoriginal:
## Overview of feature request

The `ci-lint.yml` workflow has several issues that slow down CI and create inconsistencies with other workflows:

1. **No Cargo caching** - Every clippy/fmt run downloads and compiles all dependencies from scratch, wasting CI time
2. **Missing `libonig-dev` dependency** - The `test.yml` workflow installs this for the oniguruma regex library, but `ci-lint.yml` doesn't, which could cause clippy to fail on crates that depend on it
3. **Inconsistent trigger branches** - `push` triggers on all branches (`['*']`) but `pull_request` only triggers on specific branches, creating confusion about when linting runs

### Part 1: Add Cargo caching

- [ ] Add cache steps for Cargo registry, index, and build artifacts to the `clippy_check` job
- [ ] Use the same caching pattern as `test.yml` (lines 54-76) for consistency.

### Part 2: Add missing libonig-dev dependency

- [ ] Update line 25 in `ci-lint.yml` to include `libonig-dev`

### Part 3: Align trigger branches

- [ ] Update lines 4-10 to use consistent branch patterns for both `push` and `pull_request`:
  ```yaml
  # Before:
  on:
    push:
      branches: ['*']
    pull_request:
      branches:
        - main
        - feature/*
        - feature-lib

  # After (option A - match pull_request to be more selective):
  on:
    push:
      branches:
        - main
        - feature/*
        - feature-lib
        - release/*
    pull_request:
      branches:
        - main
        - feature/*
        - feature-lib
        - release/*
  ```

### Verification

- [ ] Workflow runs successfully on a test PR
- [ ] Cache is populated on first run and restored on subsequent runs (visible in workflow logs)
- [ ] `cargo clippy --all-targets --all-features` passes
- [ ] `cargo fmt -- --check` passes
- [ ] All commits signed off (DCO requirement)

### Why caching matters

Without caching, the clippy job must:
1. Download all crate dependencies from crates.io
2. Compile all dependencies from source
3. Then run clippy

This can add 5-10+ minutes to every CI run. With caching, subsequent runs only need to compile changed code.

### Why libonig-dev is needed

The `middleware` crate depends on the `jq-rs` crate which uses oniguruma for regex. The `test.yml` workflow correctly installs `libonig-dev` (line 49), but `ci-lint.yml` doesn't. This could cause clippy to fail when analyzing the middleware crate.

PR #284, opened 2026-02-18
Fix: Postgres bootstrap `effective_from` unit mismatch

ruokun-niu:
# Description

### Problem

In the [Drasi Server Getting Started tutorial](https://github.com/drasi-project/docs/issues/191), the inactive-senders query returns an empty result after bootstrapping when it should contain all senders (since none have sent a message in the last 20 seconds).

### Root Cause

`effective_from` in the Postgres bootstrap provider was set using `timestamp_nanos_opt()` (nanoseconds), but `ElementTimestamp` is milliseconds since epoch. When `drasi.changeDateTime(m)` calls `ZonedDateTime::from_epoch_millis()` on this value, it produces a date ~1,000,000x too far in the future (year 50,000+).

This caused:
- `LastMessageTimestamp <= datetime.realtime() - duration({ seconds: 20 })` → always `false`
- `drasi.trueLater(...)` scheduled at far-future + 20s → never fires

### Fix

Changed `timestamp_nanos_opt().unwrap_or(0)` → `timestamp_millis()` to match the CDC stream (`components/sources/postgres/src/stream.rs`) and the `ElementTimestamp` type contract used by all other sources.

### Prevention

Added a `validate_effective_from()` function in `drasi-core` that detects unit mismatches by value magnitude (ns timestamps are 10⁶× larger than ms, making a simple range check reliable with zero false positives). Added a `debug_assert!` in `drasi.changeDateTime()` that fires in test builds if a bad timestamp reaches the query engine. Added per-source/bootstrapper unit tests across all components that produce `effective_from`.

## Type of change

Fixes:[ #191 ](https://github.com/drasi-project/docs/issues/191)

### DIFF
--- a/query-perf/src/main.rs
+++ b/query-perf/src/main.rs
@@ -100,24 +100,24 @@ async fn main() {
 
         // Open shared RocksDB if either index type needs it (avoids LOCK conflict
         // from opening the same unified DB path twice)
-        let (rocks_db, rocks_session_state) =
-            if test_run_config.element_index_type == IndexType::RocksDB
-                || test_run_config.result_index_type == IndexType::RocksDB
-            {
-                let options = RocksIndexOptions {
-                    archive_enabled: false,
-                    direct_io: false,
-                };
-                let path = match env::var("ROCKS_PATH") {
-                    Ok(p) => p,
-                    Err(_) => "test-data".to_string(),
-                };
-                let db = open_unified_db(&path, &query_id, &options).unwrap();
-                let session_state = Arc::new(RocksDbSessionState::new(db.clone()));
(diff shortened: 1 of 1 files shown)


ruokun-niu:
> @ruokun-niu can you think of useful unit or integration test that would have captured this. and perhaps a way that this could be generalized across any source/bootstrap provider. Maybe the size of the value would be indicative of an erroneous value?

Added a validation checker for the effective_from field that detects unit mismatches by value magnitude, and added unit tests across all source/bootstrapper components. Fixed a couple of bugs that were caught

agentofreality (approved):

Issue #290, opened 2026-02-19
Session Scoped Transactions


amansinghoriginal:
### Overview of feature request

Add **session-scoped transactions** that are held open for the entire processing cycle of one source-change, committed atomically at the end.

### Motivation
Persistent indexes exist not only for datasets that exceed available memory, but also to **avoid cold-start re-bootstrap on restart**. When a process hosting Drasi crashes and restarts, in-memory indexes are lost entirely and building the entire state is a potentially lengthy operation during which no live changes are being processed. Persistent indexes eliminate this blackout by providing near-instant recovery: open the database and resume processing.

But this value proposition depends entirely on the persisted state being **consistent**.

The core query engine processes each `SourceChange` through a multi-step pipeline - element index writes, path solving, expression evaluation, aggregation updates, and future queue operations. Each of these steps writes to a different index (`ElementIndex`, `ResultIndex`, `FutureQueue`) independently, with **no transactional boundary spanning the full processing cycle**. Every individual write auto-commits immediately.

This means a crash, OOM kill, or propagated error midway through processing a single source-change leaves persistent indexes (`index-rocksdb`, `index-garnet`) in an **inconsistent state** that can only be recovered by a full re-bootstrap from the source.

### Lack of atomic writes
When the engine processes a single source-change (e.g., an INSERT), the following writes occur **sequentially but independently**:

```
change_lock.lock()                       ← serializes concurrent changes

1.  ElementIndex.set_element()           ← auto-committed to backend
       └─ update_source_joins()          ← more auto-committed writes (virtual join relations)

2.  Path solver reads from ElementIndex  ← reads committed state

3.  Expression evaluator processes query parts:
    3a. ResultIndex.get()                ← read current accumulator value
    3b. aggregate.apply()                ← in-memory modification
    3c. ResultIndex.set()                ← auto-committed to backend
    3d. LazySortedSet.commit()           ← auto-committed to backend (separate call)
    3e. FutureQueue.push()              ← auto-committed to backend (if temporal function)

4.  (For DELETE) ElementIndex.delete_element()  ← auto-committed to backend

change_lock.unlock()
```

The `change_lock` mutex ensures that only one source-change is processed at a time per query. However, it does not provide atomicity across the index writes within a single change. Each write is an independent, auto-committed operation.

### Bugs that can occur today after a crash

[...]

PR #296, opened 2026-02-26
Session Transactions Phase 4: Atomic Future Queue pop

amansinghoriginal:
## Description

If `FutureQueue::pop()` runs **outside** the session boundary, then a crash after `pop()` but before `commit_session()` permanently loses a future event, because the element would never be re-evaluated at its scheduled time. This PR moves `pop()` inside the session so it rolls back atomically with all other index writes on failure.

It also removes the auto-commit fallback from all RocksDB and Garnet write methods, now that every production entry point (`process_source_change()`, `process_due_futures()`) uses `SessionGuard`.

**Key decisions (rationale on the issue thread):**

- New `process_due_futures()` method on `ContinuousQuery`
  - Encapsulates lock -> begin session -> pop -> middleware -> process -> commit as a single atomic unit. Callers never touch the raw queue.

- `FutureQueueConsumer` trait simplified to `on_items_due()` (no params)
  - Consumer calls `process_due_futures()` internally; the polling loop only peeks, never pops.

- `AutoFutureQueueConsumer` holds `Weak<ContinuousQuery>` (was `Arc`)
  - Breaks reference cycle: CQ -> future_queue_task -> closure -> consumer -> CQ. If CQ is dropped, upgrade fails and the polling loop exits cleanly via the existing `Notify` shutdown. THis is not necessarily required as part of this PR, but noticed this issue so included the fix here.

- Garnet `pop()` session-aware via non-destructive read + buffer 
  - Inside a session: `ZRANGEBYSCORE LIMIT 0 1` to read head, then buffer `ZREM`/`SREM`. On commit, executes atomically via `MULTI/EXEC`. On rollback, buffer is dropped and item stays in Redis.

- `FutureQueueSource` as peek-only signaler
  - Polls `peek_due_time()` and dispatches lightweight `FuturesDue` control signals through the existing priority queue. 

- The processor handles the signal by calling `process_due_futures()` in a drain loop.
  - Signaler never pops. This keeps separation of concerns and avoids hot-path `peek_due_time()` overhead on every regular event.

- Extract `dispatch_query_results()` helper
  - Shared between the regular event path and the futures-due path; avoids duplicating result conversion, current_results update, and reaction dispatch logic.
 
- Remove auto-commit fallback from write methods
  - RocksDB `with_txn` returns error when no session is active. Garnet write methods remove the `else { direct redis }` branch. Read methods retain dual-path (buffer-then-Redis) for read-your-own-writes within sessions.

## Type of change

Part of: https://github.com/drasi-project/drasi-core/issues/290

### DIFF
--- a/components/indexes/rocksdb/src/element_index.rs
+++ b/components/indexes/rocksdb/src/element_index.rs
@@ -18,7 +18,6 @@ use std::{
     sync::{Arc, RwLock},
 };
 
-use async_stream::stream;
 use async_trait::async_trait;
 use bit_set::BitSet;
 use drasi_core::{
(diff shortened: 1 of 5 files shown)


amansinghoriginal:
Re-ordered the phases and covered this change in - https://github.com/drasi-project/drasi-core/pull/300

PR #304, opened 2026-03-03
feat: Add Microsoft Dataverse source and bootstrap provider

ruokun-niu:
# Description

Adds a new Dataverse source plugin and bootstrap provider that enable Drasi to continuously monitor Microsoft Dataverse (Power Platform) tables for data changes using the OData Web API change tracking mechanism — the REST equivalent of the platform's `RetrieveEntityChangesRequest`.

## New Crates

- **`drasi-source-dataverse`** (`components/sources/dataverse/`) — Polling-based change detection source using OData delta links
- **`drasi-bootstrap-dataverse`** (`components/bootstrappers/dataverse/`) — Initial data loader that fetches all existing records from configured entities

### Change Tracking
- Uses `Prefer: odata.track-changes` headers and `@odata.deltaLink` tokens (Web API equivalent of `RetrieveEntityChangesRequest` with `DataVersion`)
- Per-entity worker tasks with independent delta tokens
- Delta tokens persisted to StateStore for crash recovery

### Adaptive Backoff
- Matches the platform's `SyncWorker` adaptive polling pattern
- 500ms minimum → 30s maximum interval
- 1.2× multiplier under 5s, 1.5× above 5s
- Resets to minimum on detected changes

### Authentication
- **Client credentials** (OAuth2) — `tenant_id` + `client_id` + `client_secret`
- **Azure CLI** — `az account get-access-token` for local development without secrets

### Platform YAML Compatibility
- Serde aliases for camelCase field names (`endpoint`, `clientId`, `clientSecret`, `tenantId`, `useAzureCli`)
- `entities` field accepts a JSON array
- - 
## Configuration Example

```yaml
apiVersion: v1
kind: Source
name: leads
spec:
  kind: Dataverse
  properties:
    endpoint: https://orgXXXXX.crm.dynamics.com/
    entities: lead
    clientId: <app-client-id>
    clientSecret: <app-client-secret>
    tenantId: <azure-ad-tenant-id>
```

Or with Azure CLI auth (no secrets):

```yaml
apiVersion: v1
kind: Source
name: leads
spec:
  kind: Dataverse
  properties:
    endpoint: https://orgXXXXX.crm.dynamics.com/
    entities: lead
    useAzureCli: true
```

### New Files
| Path | Purpose |
|------|---------|
| `components/sources/dataverse/` | Source crate (config, auth, types, client, lib) |
| `components/bootstrappers/dataverse/` | Bootstrap crate (config, lib) |
| `components/sources/dataverse/AUTHENTICATION.md` | Auth method documentation |
| `examples/lib/dataverse-getting-started/` | Standalone example with auto-detected auth |

### Modified Files
| Path | Change |
|------|--------|
| `Cargo.toml` | Added both crates to workspace members |

## Testing

- **43 unit tests** across both crates covering config validation, serde aliases, entity set naming, change parsing, auth modes, builder patterns, and value conversion
- **1 integration test** (ignored, requires live Dataverse environment or wiremock)
- All tests pass, build clean

## Prerequisites for Dataverse Environment

1. Change tracking must be enabled on target entities (Settings → Customization → Entity → Enable Change Tracking)
2. For client credentials: Azure AD app registration with Dataverse API permissions + application user in the environment
3. For Azure CLI: `az login` with a user that has Dataverse access

## Type of change

Fixes: #issue_number


components/bootstrappers/dataverse/src/config.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Azure AD application (client) ID.
+    /// Required for client credentials flow, ignored when `use_azure_cli` is true.
+    #[serde(default, alias = "clientId")]
+    pub client_id: String,
+
+    /// Azure AD client secret.
+    /// Required for client credentials flow, ignored when `use_azure_cli` is true.
+    #[serde(default, alias = "clientSecret")]
+    pub client_secret: String,
```
danielgerlag, line 59: We should wrap these in ConfigValue<T> so that they can be mapped to env variables or secrets.

components/sources/dataverse/src/config.rs
```diff
@@ -0,0 +1,457 @@
...
+///   - `use_azure_cli = true` for Azure CLI authentication (local dev)
+///
+/// # Optional Configuration
+///
+/// - `entity_set_overrides`: Override computed entity set names for specific entities
+/// - `polling_interval_ms`: Polling interval in milliseconds (default: 5000)
+/// - `min_interval_ms`: Minimum adaptive polling interval (default: 500)
+/// - `max_interval_seconds`: Maximum adaptive polling interval (default: 30)
+/// - `api_version`: Dataverse Web API version (default: `v9.2`)
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
danielgerlag: These are internal domain objects, when do they need to be serialized / deserialized?

components/sources/dataverse/src/descriptor.rs
```diff
@@ -0,0 +1,181 @@
...
+            entities: dto.entities,
+            entity_set_overrides: dto.entity_set_overrides,
+            entity_columns: dto.entity_columns,
+            polling_interval_ms,
+            min_interval_ms,
+            max_interval_seconds,
+            api_version,
+        };
+
+        let source = crate::DataverseSourceBuilder::new(id)
```
danielgerlag: entity_set_overrides and entity_columns are being dropped here

components/sources/dataverse/src/descriptor.rs
```diff
@@ -0,0 +1,181 @@
...
+        let config = DataverseSourceConfig {
+            environment_url,
+            tenant_id,
+            client_id,
+            client_secret,
+            use_azure_cli: dto.use_azure_cli,
+            entities: dto.entities,
+            entity_set_overrides: dto.entity_set_overrides,
+            entity_columns: dto.entity_columns,
+            polling_interval_ms,
```
danielgerlag: polling interval is never actually used in the polling loop?

components/bootstrappers/dataverse/src/config.rs
ruokun-niu: These fields are already wrapped in `ConfigValue` in `descriptor.rs,` which is the platform/dynamic plugin entry point. DtoMapper resolves env vars and secrets into plain strings, which are then stored here in config.rs. This matches the pattern used by the postgres source and bootstrap components.

components/sources/dataverse/src/config.rs
ruokun-niu: Good catch. they dont need ot be serialized

components/sources/dataverse/src/descriptor.rs
ruokun-niu: removed

components/sources/dataverse/src/descriptor.rs
ruokun-niu: Good catch. I have added them back

components/bootstrappers/dataverse/src/descriptor.rs
```diff
@@ -0,0 +1,221 @@
...
+    #[serde(default)]
+    pub client_secret: Option<ConfigValue<String>>,
+
+    /// Use Azure CLI for authentication instead of client credentials.
+    #[serde(default)]
+    pub use_azure_cli: bool,
+
+    /// Entity logical names to bootstrap (e.g., `["account", "contact"]`).
+    pub entities: Vec<String>,
+
```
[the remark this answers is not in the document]
ruokun-niu: The DTOs and documentation are already consistent: [entities](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a YAML/JSON array, not a comma-separated string.

Cargo.toml
```diff
@@ -29,6 +30,7 @@ members = [
   "components/sources/application",
   "components/sources/mock",
   "components/sources/mssql",
+  "components/sources/dataverse",
```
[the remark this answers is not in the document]
ruokun-niu: Outdated. Resolved

components/bootstrappers/dataverse/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+        let obj = record.as_object()?;
+
+        // Extract entity ID - try common Dataverse ID patterns
+        let entity_id_key = format!("{entity_name}id");
+        let id = obj
+            .get(&entity_id_key)
+            .or_else(|| obj.get("id"))
+            .and_then(|v| v.as_str())
+            .unwrap_or("unknown");
+
```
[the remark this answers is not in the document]
ruokun-niu: Updated

components/sources/dataverse/src/config.rs
```diff
@@ -0,0 +1,295 @@
...
+            override_name.clone()
+        } else {
+            format!("{entity}s")
+        }
+    }
+
+    /// Get the `$select` clause for a given entity, or `None` for all columns.
+    pub fn select_columns(&self, entity: &str) -> Option<String> {
+        self.entity_columns.get(entity).map(|cols| cols.join(","))
+    }
```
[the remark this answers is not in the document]
ruokun-niu: Updated

ruokun-niu:
@copilot resolve the merge conflicts in this pull request

components/sources/dataverse/src/descriptor.rs
```diff
@@ -0,0 +1,273 @@
...
+    #[serde(default)]
+    pub client_secret: Option<ConfigValue<String>>,
+
+    /// Use Azure CLI for authentication instead of client credentials.
+    #[serde(default)]
+    pub use_azure_cli: bool,
+
+    /// Entity logical names to monitor (e.g., `["account", "contact"]`).
+    pub entities: Vec<String>,
+
```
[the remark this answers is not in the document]
ruokun-niu: PR description is outdated. I have updated this

.github/agents/source-planner.md
```diff
@@ -62,6 +62,24 @@ For **Protocol/Local** sources:
...
 - API limitations that require workarounds
 
+### 3.5. Authentication — Use Existing Identity Providers
+
+Before designing custom authentication logic, review the identity provider infrastructure in `lib/src/identity/`:
+
+- **`IdentityProvider` trait** (`lib/src/identity/mod.rs`) — the standard interface for credential acquisition. Returns `Credentials::Token`, `Credentials::UsernamePassword`, or `Credentials::Certificate`.
+- **`AzureIdentityProvider`** (`lib/src/identity/azure.rs`, behind `azure-identity` feature) — supports managed identity, client secrets, developer tools (CLI/azd/PS), and workload identity. Use `.with_scope()` to set the token audience.
+- **`AwsIdentityProvider`** (`lib/src/identity/aws.rs`, behind `aws-identity` feature) — for AWS-based sources.
+- **`PasswordIdentityProvider`** (`lib/src/identity/password.rs`) — simple username/password provider.
```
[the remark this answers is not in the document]
ruokun-niu: Updated

components/bootstrappers/dataverse/src/lib.rs
```diff
@@ -0,0 +1,872 @@
...
+        // Build properties from all non-annotation attributes
+        let mut properties = ElementPropertyMap::new();
+        for (key, value) in obj {
+            // Skip OData annotations (prefixed with @ or containing @)
+            if key.starts_with('@') || key.contains('@') {
+                continue;
+            }
+            // Skip navigation property annotations (_xxx_value)
+            // But keep the lookups themselves as they contain the foreign key GUID
```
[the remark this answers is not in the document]
ruokun-niu: Updated

danielgerlag (commented):
effective_from uses Utc::now() - The timestamp is when the source processes the change, not when it happened in Dataverse. Dataverse records have modifiedon, using that would give more accurate ordering, especially under latency or batched polls.
 
Worker death is permanent - If get_initial_delta_token() fails, the entity worker returns immediately with no retry. That entity is dead for the lifetime of the source. Should loop with backoff.

ruokun-niu:
> effective_from uses Utc::now() - The timestamp is when the source processes the change, not when it happened in Dataverse. Dataverse records have modifiedon, using that would give more accurate ordering, especially under latency or batched polls.
> 
> Worker death is permanent - If get_initial_delta_token() fails, the entity worker returns immediately with no retry. That entity is dead for the lifetime of the source. Should loop with backoff.

@danielgerlag I have addressed these comments

danielgerlag (approved):

agentofreality (changes requested):
All the copyright headers are dated '2025' they should be '2026'

ruokun-niu:
> All the copyright headers are dated '2025' they should be '2026'

Updated

ruokun-niu:
## 🧪 Testing Review

### Summary

The source crate (`drasi-source-dataverse`) has reasonable unit test coverage for construction, config, types parsing, and change conversion. However, several critical paths are untested. The bootstrapper crate has **zero tests**.

---

🔴 **Blocker — Integration test is `#[ignore]`; never runs in CI**

`tests/integration_test.rs::test_dataverse_change_detection_end_to_end` is the only end-to-end test covering the INSERT → UPDATE → DELETE lifecycle, but it is marked `#[ignore]`. This means CI never validates the core change-detection flow. The test also depends on types (`drasi_lib::DrasiLib`, `drasi_reaction_application::ApplicationReaction`) that appear to be high-level facade types — if they don't resolve, the test is dead code. Either make this test runnable in CI (even gated behind a feature flag or separate workflow), or replace it with a focused wiremock test that directly exercises `DataverseSource` without the full DrasiLib stack.

---

🔴 **Blocker — Bootstrapper crate has no tests**

`components/bootstrappers/dataverse/` contains no test modules or test files. The bootstrap provider is responsible for loading all initial data — a correctness failure here silently seeds the system with wrong or missing data. At minimum, test:
- Pagination: an OData response with `@odata.nextLink` should page through all records before completing.
- `SourceChange::Insert` is emitted for each fetched record with correct element ID and properties.
- `validate()` rejects missing `environment_url` or `entities`.

---

🟡 **Should-Fix — Adaptive backoff algorithm is untested**

The two-phase multiplicative backoff (1.2× under 5s, 1.5× above; reset to min on changes) is a key correctness concern. There are no tests verifying the backoff progression or the reset-on-change behavior. This logic lives in `run_entity_worker` and can be extracted and unit-tested by driving a mock that returns changes or no-changes responses in sequence.

---

🟡 **Should-Fix — Delta token persistence (StateStore) has no test coverage**

The PR description states delta tokens are persisted to `StateStore` for crash recovery, but there are no tests verifying that (a) a token is persisted after a successful poll, or (b) a previously persisted token is loaded on restart instead of triggering a full re-initialization. Suggest a unit test with a mock `StateStoreProvider` to verify read-on-start and write-on-poll.

---

🟡 **Should-Fix — HTTP error paths in `DataverseClient` are untested**

`client.rs` has no tests. Critical error conditions not covered:
- **401 Unauthorized** — should trigger token refresh, not a permanent failure.
- **429 / 503** — should back off and retry.
- **Network timeout / connection reset** — should not crash the worker task.
- **Malformed OData response** (missing `@odata.deltaLink`) — `poll_for_changes` should handle gracefully.

These are testable with wiremock without the full integration stack.

---

🟡 **Should-Fix — Azure CLI auth path tested only for construction**

`auth.rs` tests verify `TokenManager::azure_cli()` constructs correctly, but the actual `az account get-access-token` subprocess path is never exercised. Add a test that mocks the subprocess (or injects a fake CLI path) and verifies token parsing and error handling when `az` is not available.

---

🔵 **Nit — Timing-based assertions in the integration test are flaky-prone**

The integration test uses `tokio::time::sleep(Duration::from_secs(1))` to "wait for the source to start," then polls for up to 15 seconds with 2-second recv timeouts. This is sensitive to scheduler pressure in CI. If the test is ever made non-ignored, replace the sleep with a signal (e.g., a `tokio::sync::Notify` or status channel from the source) indicating readiness.

---

🔵 **Nit — `config.rs` `validate()` has no dedicated tests**

`DataverseSourceConfig::validate()` is indirectly tested through builder tests, but boundary cases like `environment_url` with no host, `use_azure_cli = true` with missing credentials (should succeed), and conflicting auth config are not explicitly verified.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> [!NOTE]
> <details>
> <summary>🔒 Integrity filter blocked 1 item</summary>
>
> The following item was blocked because it doesn't meet the GitHub integrity level.
>
> - [#304](https://github.com/drasi-project/drasi-core/pull/304) `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
>
> To allow these resources, lower `min-integrity` in your GitHub frontmatter:
>
> ```yaml
> tools:
>   github:
>     min-integrity: approved  # merged | approved | unapproved | none
> ```
>
> </details>


> Generated by [pr-testing-reviewer](https://github.com/drasi-project/.github/actions/runs/25465448674/agentic_workflow) · ● 275.9K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-testing-reviewer%22&type=issues)

ruokun-niu:
## 🔒 Security Review

PR: [feat: Add Microsoft Dataverse source and bootstrap provider](https://github.com/drasi-project/drasi-core/pull/304)

---

### 🟡 Finding 1 — `client_secret` exposed via `#[derive(Debug)]`

**Files:**
- `components/sources/dataverse/src/config.rs`, line 44 — `DataverseSourceConfig`
- `components/sources/dataverse/src/auth.rs`, line 33 — `AuthMethod` (variant `ClientSecret { client_secret }`)
- `components/bootstrappers/dataverse/src/config.rs`, line 44 — `DataverseBootstrapConfig`

**Risk:** All three structs carry `client_secret: String` and derive `Debug`. Any code path that formats these structs with `{:?}` — tracing spans, log macros, error display, assertion failures — will print the plaintext secret to logs. This is a realistic risk: a library consumer writes `tracing::debug!("{:?}", config)` or Rust panics with `{:?}` and the secret flows into structured log aggregators.

**Attack scenario:** An attacker with log-read access (e.g., cloud logging service) retrieves a valid Azure AD client secret and authenticates against the Dataverse environment directly.

**Fix:** Implement a manual `Debug` that redacts the secret field:

```rust
impl std::fmt::Debug for DataverseSourceConfig {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("DataverseSourceConfig")
            .field("environment_url", &self.environment_url)
            .field("tenant_id", &self.tenant_id)
            .field("client_id", &self.client_id)
            .field("client_secret", &"[REDACTED]")
            .field("use_azure_cli", &self.use_azure_cli)
            .field("entities", &self.entities)
            // ... remaining non-sensitive fields
            .finish()
    }
}
```

Apply the same pattern to `AuthMethod::ClientSecret` and the bootstrapper config. Alternatively, use the [`secrecy`]((crates.io/redacted) crate (`Secret<String>`) which prevents accidental `Debug` exposure at the type level.

---

### 🟡 Finding 2 — SSRF via unvalidated delta/next link URLs

**File:** `components/sources/dataverse/src/client.rs`
- `follow_delta_link(&self, delta_link: &str)` — line 149
- `follow_next_link(&self, next_link: &str)` — line 184

**Risk:** `delta_link` and `next_link` are taken directly from the API response body (`@odata.deltaLink`, `@odata.nextLink`) and used as HTTP targets without validating they remain within the expected Dataverse environment origin. If the Dataverse endpoint is compromised, if TLS validation is bypassed upstream, or if a misconfigured API gateway reflects attacker-supplied URLs, the client will make authenticated (bearer token) HTTP requests to arbitrary internal addresses — including cloud metadata services (`169.254.169.254`), internal APIs, or other services reachable from the runtime environment.

**Attack scenario:** Attacker controls or influences the Dataverse API response, injects `"`@odata`.deltaLink": "(169.254.169.254/redacted) The source makes a bearer-token-decorated request to the metadata service.

**Fix:** Validate that delta/next link URLs start with the known `self.base_url` before sending the request:

```rust
pub async fn follow_delta_link(&self, delta_link: &str) -> Result<ODataDeltaResponse> {
    if !delta_link.starts_with(&self.base_url) {
        anyhow::bail!(
            "delta_link host does not match expected base URL '{}': '{}'",
            self.base_url, delta_link
        );
    }
    // ... existing request logic
}
```

Apply the same check to `follow_next_link`.

---

### 🔵 Finding 3 — Full OAuth2 error response body propagated in error messages

**File:** `components/sources/dataverse/src/auth.rs`, `refresh_token_client_credentials`

```rust
anyhow::bail!("OAuth2 token request failed with status {status}: {body}");
```

**Risk:** Azure AD error responses are typically benign (`invalid_client`, `unauthorized_client`), but in abnormal cases (e.g., a proxied error from a misconfigured endpoint) the body could contain sensitive context. More practically, structured error logs that include this message may expose tenant/client configuration details to log viewers who should not have them.

**Fix:** Log the full body at `debug` level and surface only the status code in user-facing errors:

```rust
log::debug!("OAuth2 token error body: {body}");
anyhow::bail!("OAuth2 token request failed with status {status}");
```



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> [!NOTE]
> <details>
> <summary>🔒 Integrity filter blocked 1 item</summary>
>
> The following item was blocked because it doesn't meet the GitHub integrity level.
>
> - [#304](https://github.com/drasi-project/drasi-core/pull/304) `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
>
> To allow these resources, lower `min-integrity` in your GitHub frontmatter:
>
> ```yaml
> tools:
>   github:
>     min-integrity: approved  # merged | approved | unapproved | none
> ```
>
> </details>


> Generated by [pr-security-reviewer](https://github.com/drasi-project/.github/actions/runs/25465446114/agentic_workflow) · ● 507.2K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-security-reviewer%22&type=issues)

ruokun-niu:
## 🔍 Prior Art Review

### 🟡 Should-Fix — Custom OAuth2 token management duplicates existing codebase infrastructure

**File:** `components/sources/dataverse/src/auth.rs`

The new `TokenManager` struct manually implements:
1. **OAuth2 client credentials flow** — raw HTTP `POST` to `login.microsoftonline.com/.../oauth2/v2.0/token`, deserializing `TokenResponse`, and caching in `Arc<RwLock<Option<CachedToken>>>`.
2. **Azure CLI authentication** — spawning `tokio::process::Command::new("az") .args(["account", "get-access-token", ...])` and parsing the JSON output, including custom timestamp parsing for two different `az` CLI output formats.

**This functionality already exists in the codebase.** The `components/identity/azure/` crate (`drasi-identity-azure`) already depends on `azure_identity = "0.31"` and provides `AzureIdentityProvider`, which wraps:

| Custom `TokenManager` | Existing `AzureIdentityProvider` / `azure_identity` |
|---|---|
| Manual client credentials HTTP flow | `azure_identity::ClientSecretCredential` |
| `az account get-access-token` subprocess + timestamp parsing | `azure_identity::AzureCliCredential` / `DeveloperToolsCredential` |
| Manual `Arc<RwLock<Option<CachedToken>>>` | Token caching built into `azure_identity` credentials |
| Implements `IdentityProvider` trait | Already implements `IdentityProvider` trait |

`AzureIdentityProvider` also already supports custom scopes via `.with_scope("(myorg.crm.dynamics.com/redacted), which is exactly what Dataverse authentication needs.

**Recommendation:** Replace `auth.rs` with a thin wrapper around `AzureIdentityProvider` (or the `azure_identity` primitives directly):

```rust
// Client credentials
AzureIdentityProvider::with_client_secret("dataverse", tenant_id, client_id, client_secret)
    .with_scope(format!("{environment_url}/.default"))

// Azure CLI
AzureIdentityProvider::with_default_credentials("dataverse")
    .with_scope(format!("{environment_url}/.default"))
```

*(If `AzureIdentityProvider::with_client_secret` doesn't exist yet, it can be added to `drasi-identity-azure` — that crate is the right home for it.)*

**Trade-offs:** None significant. `azure_identity` is already a workspace dependency, handles token caching internally, and is maintained by the Azure SDK for Rust team. Removing `auth.rs` eliminates ~446 lines of auth code and the subprocess-based `az` CLI path parsing.

---

### ✅ Everything else is justified

- **Adaptive polling backoff** (two-phase 1.2×/1.5× multiplier, square-root entity count scaling): This matches specific Dataverse platform semantics from `ChangeMonitor.cs`/`SyncWorker`. Generic backoff crates (e.g., `backoff`, `tokio-retry`) don't support this two-phase pattern or entity-count scaling, so the custom implementation is appropriate.
- **Raw `reqwest` for OData delta links**: No well-maintained Rust OData client crate exists; `reqwest` is the right choice.
- **Custom `types.rs` / `descriptor.rs`**: Dataverse-specific data mapping with no generic alternative.



> [!WARNING]
> <details>
> <summary>Firewall blocked 2 domains</summary>
>
> The following domains were blocked by the firewall during workflow execution:
>
> - `crates.io`
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "crates.io"
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> [!NOTE]
> <details>
> <summary>🔒 Integrity filter blocked 1 item</summary>
>
> The following item was blocked because it doesn't meet the GitHub integrity level.
>
> - [#304](https://github.com/drasi-project/drasi-core/pull/304) `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
>
> To allow these resources, lower `min-integrity` in your GitHub frontmatter:
>
> ```yaml
> tools:
>   github:
>     min-integrity: approved  # merged | approved | unapproved | none
> ```
>
> </details>


> Generated by [pr-prior-art-reviewer](https://github.com/drasi-project/.github/actions/runs/25465449422/agentic_workflow) · ● 591.5K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-prior-art-reviewer%22&type=issues)

ruokun-niu:
## 📝 Documentation Review

**PR:** drasi-project/drasi-core#304 — Dataverse source + bootstrapper

---

### 🟡 Should-Fix

**`components/sources/dataverse/src/config.rs` — `DataverseSourceConfig` struct doc, lines ~39–41 (in diff)**

The `# Required Configuration` section lists only two authentication options, omitting identity provider — which the README presents as the **recommended** approach:

```rust
/// - Authentication (one of):
///   - `tenant_id` + `client_id` + `client_secret` for client credentials flow
///   - `use_azure_cli = true` for Azure CLI authentication (local dev)
```

Suggested replacement:

```rust
/// - Authentication (one of):
///   - Identity provider via [`DataverseSource::builder`] `.with_identity_provider()` (recommended)
///   - `tenant_id` + `client_id` + `client_secret` for client credentials flow
///   - `use_azure_cli = true` for Azure CLI authentication (local dev)
```

---

### 🔵 Nit

**`components/bootstrappers/dataverse/src/config.rs` — `use_azure_cli` field doc, lines ~66–67 (in diff)**

The bootstrapper's `use_azure_cli` field doc omits the `az login` prerequisite, unlike the equivalent field in `DataverseSourceConfig` (which includes "Requires `az login` to have been run beforehand."):

```rust
/// Use Azure CLI (`az account get-access-token`) for authentication.
/// When true, `tenant_id`, `client_id`, and `client_secret` are not required.
```

Suggested replacement:

```rust
/// Use Azure CLI (`az account get-access-token`) for authentication.
/// When true, `tenant_id`, `client_id`, and `client_secret` are not required.
/// Requires `az login` to have been run beforehand.
```

---

Everything else — READMEs, builder method docs, type docs, and the `source-planner.md` addition — is clear, complete, and consistent.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> [!NOTE]
> <details>
> <summary>🔒 Integrity filter blocked 2 items</summary>
>
> The following items were blocked because they don't meet the GitHub integrity level.
>
> - [#304](https://github.com/drasi-project/drasi-core/pull/304) `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - drasi-project/drasi-core#304 `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
>
> To allow these resources, lower `min-integrity` in your GitHub frontmatter:
>
> ```yaml
> tools:
>   github:
>     min-integrity: approved  # merged | approved | unapproved | none
> ```
>
> </details>


> Generated by [pr-docs-reviewer](https://github.com/drasi-project/.github/actions/runs/25465447804/agentic_workflow) · ● 1.1M · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-docs-reviewer%22&type=issues)

ruokun-niu:
## ✅ Correctness Review

Three findings across the new Dataverse source and bootstrapper components.

---

### 🟡 Should-Fix — Token refresh stampede under concurrency

**`components/sources/dataverse/src/auth.rs` — `TokenManager::get_token()`**

The read lock is dropped before entering the refresh path. If N tasks all call `get_token()` concurrently when the cache is stale, all N will race past the read-lock check and fire N independent token requests to Azure AD. Only the last write wins in the cache; the rest are wasted. Under high-concurrency source polling this can trigger Azure AD token-endpoint rate limiting.

```rust
// Current (races): read lock drops, then multiple callers refresh independently
pub async fn get_token(&self) -> Result<String> {
    {
        let cached = self.cached_token.read().await;
        if let Some(ref token) = *cached {
            if Instant::now() + Duration::from_secs(30) < token.expires_at {
                return Ok(token.access_token.clone());
            }
        }
    } // <-- read lock released; N concurrent callers all proceed to refresh

    match &self.auth_method { ... }
}
```

Fix — upgrade to a write lock and do a double-checked check inside it:

```rust
pub async fn get_token(&self) -> Result<String> {
    // Fast path: valid cached token under read lock
    {
        let cached = self.cached_token.read().await;
        if let Some(ref token) = *cached {
            if Instant::now() + Duration::from_secs(30) < token.expires_at {
                return Ok(token.access_token.clone());
            }
        }
    }

    // Slow path: acquire write lock and re-check before refreshing
    let mut cache = self.cached_token.write().await;
    if let Some(ref token) = *cache {
        if Instant::now() + Duration::from_secs(30) < token.expires_at {
            return Ok(token.access_token.clone()); // another task already refreshed
        }
    }

    let token = match &self.auth_method {
        AuthMethod::ClientSecret { .. } => self.fetch_token_client_credentials().await?,
        AuthMethod::AzureCli => self.fetch_token_azure_cli().await?,
    };
    *cache = Some(CachedToken { ... });
    Ok(token)
}
```

This requires extracting the HTTP fetch into helpers that don't acquire the cache lock (avoiding deadlock).

---

### 🟡 Should-Fix — Malformed scope when URL has no host

**`components/sources/dataverse/src/client.rs` — `DataverseClient::new()`** (and mirror in `components/bootstrappers/dataverse/src/lib.rs` ~line 951)

`host_str()` returns `None` for host-less URLs (e.g. `(redacted) `unwrap_or_default()` silently substitutes `""`, producing a malformed scope like `https:///.default`. `TokenManager::new()` correctly rejects such URLs with an explicit error, but `DataverseClient` swallows the problem and proceeds with a bad scope that will fail at token acquisition time with an opaque error.

```rust
// Current: silently uses "" as host
let scope = if let Ok(url) = url::Url::parse(&base_url) {
    let host = url.host_str().unwrap_or_default(); // "" if no host
    format!("{}://{}/.default", url.scheme(), host) // "https:///.default"
} else {
    format!("{base_url}/.default")
};
```

Fix — treat a missing host as an error (or at minimum fall back to the `else` branch):

```rust
let scope = match url::Url::parse(&base_url)
    .ok()
    .and_then(|u| u.host_str().map(|h| format!("{}://{}/.default", u.scheme(), h)))
{
    Some(s) => s,
    None => format!("{base_url}/.default"),
};
```

The same `unwrap_or_default()` pattern appears in the bootstrapper's `get_token()` at approximately the same call site.

---

### 🟡 Should-Fix — Lossy `i64 → u64` cast for `effective_from`

**`components/sources/dataverse/src/lib.rs` — `convert_to_source_change()`**

`timestamp_millis()` returns `i64`. Casting to `u64` with `as` wraps negative values (dates before 1970-01-01) to enormous positive numbers instead of producing an error. For the `Deleted` arm the timestamp is always `Utc::now()` so it's safe; for `NewOrUpdated` the `modifiedon` value from Dataverse is parsed from RFC 3339 — in practice always post-epoch, but defensive handling is cheap.

```rust
// Current: wraps negative timestamps silently
.map(|dt| dt.timestamp_millis() as u64)
.unwrap_or_else(|| chrono::Utc::now().timestamp_millis() as u64)
```

Fix — clamp negative values explicitly:

```rust
.map(|dt| dt.timestamp_millis().max(0) as u64)
.unwrap_or_else(|| chrono::Utc::now().timestamp_millis().max(0) as u64)
```



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> [!NOTE]
> <details>
> <summary>🔒 Integrity filter blocked 2 items</summary>
>
> The following items were blocked because they don't meet the GitHub integrity level.
>
> - [#304](https://github.com/drasi-project/drasi-core/pull/304) `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
> - drasi-project/drasi-core#304 `pull_request_read`: has lower integrity than agent requires. The agent cannot read data with integrity below "approved".
>
> To allow these resources, lower `min-integrity` in your GitHub frontmatter:
>
> ```yaml
> tools:
>   github:
>     min-integrity: approved  # merged | approved | unapproved | none
> ```
>
> </details>


> Generated by [pr-correctness-reviewer](https://github.com/drasi-project/.github/actions/runs/25465445146/agentic_workflow) · ● 1.1M · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-correctness-reviewer%22&type=issues)

agentofreality (approved):

PR #350, opened 2026-04-10
MQTT Source

### LINKED ISSUE #301: MQTT Source for Drasi
This issue will be related to adding MQTT Source within LFX mentorship
I am opening it so you can track my progress and discuss details if needed when you have the moment.

ahmed-kamal2004:
# Description

This PR introduces the MQTT Source to drasi-core, enabling real-time ingestion of IoT and industrial telemetry data directly into Drasi’s graph-based query engine. This implementation focuses on high-performance topic-to-graph mapping using a flexible DSL and handles the complexities of MQTT messaging (reconnections, protocol fallback, and security).

- Key  Features

    - Pattern-Based Mapping (DSL): Implements a template-driven engine to extract metadata from MQTT topics and  JSON payloads. It supports dynamic node and relation generation via {variable} and $.json_path extraction.

    - Protocol Flexibility: Built on rumqttc with support for MQTT v5 (default) and automated fallback to v3.1.1.

    - Ingestion Modes: Supports payload_as_field (for single-value ) and payload_spread (for full JSON object ingestion).
    - Security: Integrate with identity provider-based authentication and mutual TLS authentication.

- Design document
https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Mqtt-Source/mqtt-source-design.md

## Testing
Testing  is implemented covering almost all the plugin features (including integration and unit tests).

## Type of change

This pull request adds or changes features of Drasi and has an approved issue (#301 ).

Fixes: #301


ahmed-kamal2004:
I think the PR now is complete, PTAL when you have the moment @amansinghoriginal 

I will need to squash some commits into a more better commit

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,562 @@
...
+        params: &Params,
+        timestamp: u64,
+    ) -> Vec<MqttSourceChange> {
+        let mut result = Vec::new();
+        for relation in relations {
+            let relation_id = Self::map_template(&relation.id, params);
+            let element = MqttElement::Relation {
+                id: relation_id.clone(),
+                labels: vec![relation.label.clone()],
+                from: relation.from.clone(),
```
amansinghoriginal: Relations point to labels, not node IDs. The emitted graph will be broken.

amansinghoriginal: Why are we using the label string here directly as `element_id`?

With this we will get relations that point to ghost elements whose IDs don't match any real nodes.

The tests validate the broken behavior (pattern.rs asserts from="FLOOR" and to="ROOM" on the emitted relation). Relations should be from ID to ID instead of label to label.

amansinghoriginal: Resolve each relation.from/to label to the corresponding templated node id

examples/lib/mqtt-source-example/main.rs
```diff
@@ -0,0 +1,176 @@
...
+    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
+
+    println!("╔════════════════════════════════════════════╗");
+    println!("║     DrasiLib Temperature Monitor Example   ║");
+    println!("╚════════════════════════════════════════════╝\n");
+
+
+    // Add MqttSource configuration
+    let source_config_yaml = r#"
+broker_addr: "localhost"
```
amansinghoriginal: Is this a valid config field? I think the config has `host` and not `broker_addr`.

components/sources/mqtt/src/mqtt.rs
```diff
@@ -0,0 +1,492 @@
...
+                let _ = shutdown_rx.await;
+                let _ = connection_shutdown_tx.send(());
+            }
+            .instrument(span),
+        );
+
+        *self.base.task_handle.write().await = Some(server_handle);
+        *self.base.shutdown_tx.write().await = Some(shutdown_tx);
+
+        match timeout(Duration::from_millis(500), error_rx).await {
```
amansinghoriginal: `error_rx` is consumed by the 500ms timeout. If the MQTT connection fails at t > 500ms (common: TLS handshake, DNS, etc.), the spawned task calls  
  `let _ = error_tx.send(...)`.  But `error_rx` is dropped, so the send silently fails and status remains `Running` even though the source is dead.

Plus: no reconnection logic; one disconnect and the source is permanently broken but reports healthy.

components/sources/mqtt/src/processor.rs
```diff
@@ -0,0 +1,242 @@
...
+            debug!(
+                "[{source_id}] MQTT Batcher forwarding batch #{total_batches} with {batch_size} events to dispatchers"
+            );
+
+            let mut sent_count = 0;
+            let mut failed_count = 0;
+
+            for (idx, event) in batch.into_iter().enumerate() {
+                debug!(
+                    "[{source_id}] Batch #{total_batches}, dispatching event {{idx + 1}}/{batch_size}"
```
amansinghoriginal: `{{` in a format string produces a literal `{`, right?
The log output here will be literally `event {idx + 1}/5`.

amansinghoriginal (changes requested):
Several blocking issues. Will post more comments on some other issues I found later.

components/sources/mqtt/src/mqtt.rs
ahmed-kamal2004: I added more support for reconnection and resubscription.

Another assumption I am doing, when the broker is not reachable at the start, we try connecting for 10 retries (5 for v5 and then 5 for v3), and if no connection is established, the source doesn't try again, and the source state becomes `Error`,
I am thinking about adding `loop backoff` logic (the source will be still running and keeps trying to establish connection) wdyt about that ?

examples/lib/mqtt-source-example/main.rs
ahmed-kamal2004: oh, done (and added `serde` schema definitions to prevent this again).

components/sources/mqtt/src/pattern.rs
ahmed-kamal2004: oh, I missed that. it is fixed now.

ahmed-kamal2004:
I have found that there is mis-mapping in multiple places (sources - docs AFAIK) for `in_node` in the `Element:Relation` as it should be the source node `from`, but it is classified wrongly as target node `to`

Sample Ref:
https://github.com/drasi-project/drasi-core/blob/1a480b85c29f9001659e00fa32b20beda6c78ffc/components/sources/README.md?plain=1#L315

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,480 @@
...
+                        drasi_lib::identity::Credentials::Token { username, token } => {
+                            $options.set_credentials(username, token);
+                        }
+                        drasi_lib::identity::Credentials::Certificate { cert_pem, key_pem, .. } => { // not currently implemented in the identity provider.
+                            optional_mtls_client_auth = Some((cert_pem.clone(), key_pem.clone()));
+                        }
+                    }
+                }
+                Err(e) => {
+                    error!("Failed to retrieve credentials from identity provider for MQTT authentication: {e:?}");
```
amansinghoriginal, line 56: When the identity provider fails (expired token, network blip, etc.), this logs an error but falls through to an unauthenticated connection attempt. The broker then rejects with a misleading auth error.

This should either propagate the error (return early from the macro/function) or at minimum set `ComponentStatus::Error`. Falling through to anonymous connect would make debugging confusing for users.

components/sources/mqtt/src/processor.rs
```diff
@@ -0,0 +1,245 @@
...
+    ) {
+        let source_id = source_id.to_string();
+        for (idx, change) in source_changes.into_iter().enumerate() {
+            let timestamp = match change {
+                MqttSourceChange::Insert { timestamp, .. }
+                | MqttSourceChange::Update { timestamp, .. }
+                | MqttSourceChange::Delete { timestamp, .. } => timestamp.unwrap_or_else(|| {
+                    let now = std::time::SystemTime::now();
+                    now.duration_since(std::time::UNIX_EPOCH)
+                        .expect("System time before UNIX EPOCH!")
```
amansinghoriginal: `.expect("System time before UNIX EPOCH!")` would panic the entire source task. Same pattern exists in a few places in `schema.rs`.
 
This code also seems unreachable. In `pattern.rs` aren't we always setting `Some(packet.timestamp)` on every emitted change?

Either remove the dead fallback, or replace `.expect()` with a safe default (e.g., 0) + `error!()` log.

components/sources/mqtt/src/adaptive_batcher.rs
amansinghoriginal, line 1: This file is a verbatim copy of `components/sources/http/src/adaptive_batcher.rs`.
Can you re-base update your PR for GRPC or HTTP reaction where it was going to make the adaptive batcher a single implementation?

Also check that if both GRPC and HTTP have same implementation, how large is the duplication. Is it woth making a separate crate - let's make this decision once you figure out how much duplication are we doing.

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,616 @@
...
+        relations: &Vec<MappingRelation>,
+        params: &Params,
+        timestamp: u64,
+        nodes: &Vec<MqttSourceChange>,
+    ) -> Vec<MqttSourceChange> {
+        let mut result = Vec::new();
+        for relation in relations {
+            let relation_id = Self::map_template(&relation.id, params);
+
+            let from_node = Self::get_node_with_label(&relation.from, nodes);
```
amansinghoriginal: Relationship `from`/`to` currently only resolve via label lookup in already-emitted nodes. Did we not agree to add JSONPath-style extractors (e.g., $.gateway_id) to dynamically link nodes based on payload contents?

Example:
```
to: { label: "Gateway", id: "$.gateway_id" }
```

 without it, relationships can only be built from topic variables, not payload data.

components/sources/mqtt/Cargo.toml
amansinghoriginal, line 1: This source can't be loaded as a dynamic plugin by drasi-server. It needs:

  1. crate-type = ["lib", "cdylib"] in Cargo.toml
  2. A descriptor.rs implementing SourcePluginDescriptor (kind "mqtt")
  3. export_plugin! macro invocation in lib.rs

 See the latest `components/sources/http/` for the canonical pattern.

components/sources/mqtt/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+rumqttc = "0.25.1"
+ordered-float = "3.0"
+chrono = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+serde_yaml = "0.9"
+tracing = "0.1"
+bytes = "1.5"
+matchit = "0.9.2"
+serde_repr = "0.1.20"
```
amansinghoriginal, line 50: All 55 tests are unit tests. There are no integration tests against a real MQTT broker.

The v5 to v3.1.1 fallback, TLS handshake, subscription flow, reconnect re-subscription, and retained message behavior are entirely unverified end-to-end.

Please add at minimum a testcontainers-based Mosquitto test. We can also consider gating it behind a feature-flag later.

components/sources/mqtt/src/config.rs
```diff
@@ -0,0 +1,1223 @@
...
+        for (idx, node) in mapping.nodes.iter().enumerate() {
+            Self::validate_template_placeholders(
+                &node.id,
+                &allowed,
+                &format!("nodes[{idx}].id"),
+                &mapping.pattern,
+            )?;
+        }
+
+        for (idx, relation) in mapping.relations.iter().enumerate() {
```
amansinghoriginal, line 641: This loop validates relation.id template placeholders, but doesn't check that relation.from and relation.to reference a declared nodes[].label or the entity.label. A typo in from/to would pass validation but fail at runtime with "Node with label 'X' not found".

Should we check here that each relation.from and relation.to exists in the set of {entity.label} UNION {nodes[].label}.

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,616 @@
...
+        let mut properties = match &mapping.properties.mode {
+            MappingMode::PayloadAsField => {
+                let field_name = Self::map_template(
+                    mapping.properties.field_name.as_ref().ok_or_else(|| {
+                        anyhow::anyhow!("Field name is required for PayloadAsField mode")
+                    })?,
+                    params,
+                );
+                let mut props = serde_json::Map::new();
+                let val = serde_json::from_slice(&packet.payload).unwrap_or_else(|_| {
```
amansinghoriginal: `payload_as_field` is useful for raw scalar payloads" but this code happily stores an entire JSON object (e.g., `{"lux":90}`) as the field value. The test at line 573 locks in this behavior.
 
Should we reject non-scalar payloads with an error?

components/sources/mqtt/src/schema.rs
```diff
@@ -0,0 +1,506 @@
...
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+use std::vec::Vec;
+
+/// Data schema for MQTT source events
+///
+/// This schema closely mirrors drasi_core::models::SourceChange for efficient conversion
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "operation", rename_all = "lowercase")]
+pub enum MqttSourceChange {
```
amansinghoriginal, line 27: `MqttSourceChange::Insert` and `::Delete` are defined and have conversion logic + tests, but nothing ever constructs them.  We only emit `::Update` (per the upsert requirement).

Should we remove the dead variants and their tests?

components/sources/mqtt/src/config.rs
```diff
@@ -0,0 +1,1223 @@
...
+            }
+        }
+
+        if self.max_incoming_packet_size.is_none() ^ self.max_outgoing_packet_size.is_none() {
+            return Err(anyhow::anyhow!(
+                "Both max_incoming_packet_size and max_outgoing_packet_size must be set together"
+            ));
+        }
+
+        for topic_mapping in &self.topic_mappings {
```
amansinghoriginal, line 522: There's no validation that the topics[] subscriptions can produce topics matching any topic_mappings[].pattern. 

Is it possible to emit a warn!() at startup if no subscription could produce topics matching a mapping pattern?

For example, a user could subscribe to `sensors/#` but only define mappings for `devices/{id}/telemetry`. Then every message arrives but silently fails to match.

amansinghoriginal: This an optional nice-to-have.

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,480 @@
...
+                            Ok(event) => {
+                                let packet = event.to_mqtt_packet();
+                                if let Some(packet) = packet {
+                                    if let Err(e) = processer_tx.send(packet).await {
+                                            error!("Failed to send MQTT packet to processor: {e:?}");
+                                    }
+                                }
+                            },
+                            Err(e) => {
+                                error!("MQTT event loop error: {e:?}");
```
amansinghoriginal: Thanks for logging Event loop errors.

However, the source will just contineu with no status update or backpressure, right?

If rumqttc hits persistent errors (repeated auth failures, broker gone), will the source silently spins logging errors while reporting `Running` ?

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,480 @@
...
+        // v5 specific options
+        if let Some(conn_timeout) = config.conn_timeout {
+            options.set_connection_timeout(conn_timeout);
+        }
+        if let Some(connect_properties) = config.connect_properties.as_ref() {
+            let connection_props = connect_properties.to_connection_properties();
+            options.set_connect_properties(connection_props);
+        }
+
+        if let Some(connect_properties) = config.connect_properties.as_ref() {
```
amansinghoriginal: This code block is duplicated here.

components/sources/mqtt/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
```
amansinghoriginal, line 15: Please add a README.md for the MQTT source crate. See `components/sources/http/README.md` for the expected structure.

It should cover configuration, mapping modes, TLS, plugin descriptor, and usage examples. Please keep it concise.

amansinghoriginal (changes requested):
Nice work - Thanks for working on the previous review comments. 

There are several issues that I've added inline comments for. Please take a look.
Please also provide at least 2-3 lines of PR description - the template was not filled. You can add link to your design doc instead of copying over details here.

Please follow the best practices you did in your GRPC Source PR earlier.
Also, let's try to add at least one integration test using a real broker - maybe using a testcontainer.

components/sources/mqtt/Cargo.toml
ahmed-kamal2004: Done, I think I added a full suite for integration tests now.

components/sources/mqtt/src/adaptive_batcher.rs
ahmed-kamal2004: Screenshot from 2026-04-21 00-47-28



Yes, they are identical
Actually, I have done something similar before in my mqtt reaction PR, and I think making a separate crate will make things easier (especially before the files diverge over time)

components/sources/mqtt/src/connection.rs
ahmed-kamal2004: Done, I added a two config fields `max_retries` and `base_retry_delay_secs` with exponential increase of the retry delay on errors till it hits the limit `max_retries`, if the retries exceeds the max number, it exits the loop and changes the source status to `Error`
Also added an integration testing for it `test_mqtt_reconnection_and_loopbackoff`

components/sources/mqtt/src/pattern.rs
ahmed-kamal2004: I was thinking about giving the user the full flexibility and making the config the only source of truth regardless of the data shape, wdyt ?

components/sources/mqtt/Cargo.toml
ahmed-kamal2004: Great, I have done a draft, ~but couldn't find a way to test it~
as I have found this part is a bit outdated for dynamic plugins, ~or I am missing something ?~

https://github.com/ahmed-kamal2004/drasi-server#dynamic-plugin-build

ahmed-kamal2004: Done, and verified locally

Screenshot from 2026-04-22 04-26-43

Screenshot from 2026-04-22 04-28-40

components/sources/mqtt/src/config.rs
ahmed-kamal2004: > This an optional nice-to-have.

great, let me delay this a bit so I can work on the design docs.

components/sources/mqtt/src/pattern.rs
ahmed-kamal2004: Done, I removed the `label`, as I think it doesn't help.

components/sources/mqtt/src/config.rs
ahmed-kamal2004: I removed `label` entirely, so I am now doing another type of checks on the `id` only.

ahmed-kamal2004:
@amansinghoriginal I think all the comments are resolved now, 

can you please also assign me an issue, so I can mention it in this PR (the template needs an issue)

ahmed-kamal2004:
I updated the PR to follow the latest code base changes, I think the PR can pass the tests now

cc. @amansinghoriginal

danielgerlag (changes requested):
## Review: MQTT Source Plugin (PR #350)

Hey @ahmed-kamal2004 — thanks for putting in the work on this MQTT source. The topic-to-graph mapping DSL is a thoughtful design, and the protocol negotiation (v5 → v3.1.1 fallback) is really nicely done. This is a solid foundation.

I've gone through the code in detail and have some findings I'd like to share. I've grouped them by severity to help prioritize — nothing here should be too painful to address, and many are small fixes.

---

### 🔴 High Priority

**1. Initial subscription failure is silently swallowed** (`connection.rs`)
If `subscribe_to_topics()` fails, the error is logged but the source reports `ComponentStatus::Running` with zero active subscriptions. Users would have no signal that nothing is being received. This should either abort startup or retry.

**2. No token refresh on reconnection** (`connection.rs`)
Identity provider credentials are fetched once during `MqttConnection::new()` and baked into `MqttOptions`. For token-based auth (OAuth, etc.), when the broker disconnects after token expiry, the reconnection logic reuses the same stale token — causing an infinite reconnect-fail cycle until `max_retries` is exhausted. The identity provider should be re-consulted on each reconnection attempt.

**3. Spawned processor/batcher tasks aren't explicitly cancellable** (`mqtt.rs`)
The `JoinHandle`s from `start_processing_loop()` and `start_adaptive_batcher_loop()` are dropped (detached) at the end of `start()`. Shutdown relies on the implicit channel-close cascade, but if the processor is blocked on a full `batch_tx` channel (backpressure), `stop()` can stall past its 5-second timeout. Consider storing the handles or using a shared `CancellationToken`.

**4. `event_channel_capacity: 0` causes runtime panic** (`config.rs`)
`mpsc::channel(0)` is invalid in tokio. There's no validation to prevent this — a simple `if capacity == 0 { return Err(...) }` in config validation would catch it early.

**5. Overlapping `topic_mapping` patterns are silently dropped** (`pattern.rs`)
When `matchit::Router` rejects a duplicate pattern, the error is logged but the source continues with only the first mapping. This should be a startup-time error — users won't realize their second mapping is being ignored.

---

### 🟡 Medium Priority

**6. `MqttSourceConfig` derives `Serialize` — credentials could leak** (`config.rs`)
The custom `Debug` impl correctly redacts sensitive fields (nice!), but the derived `Serialize` will include `password` in any `serde_json::to_string()` call. Consider `#[serde(skip_serializing)]` on the `password` field.

**7. Duplicate error logging in event loops** (`connection.rs`)
In both the v3 and v5 branches of `run_subscription_loop`, the `Err` arm logs the error twice — once inside the if/else and again unconditionally at the end of the arm.

**8. `adaptive_batcher.rs` is a verbatim copy from the HTTP source**
This was flagged in the earlier review and is still unresolved. Extracting to a shared crate would prevent the two copies from diverging.

**9. Unmatched topics generate `error!()` logs** (`processor.rs`)
If you subscribe to `building/#` but only define mappings for `building/{floor}/{room}/{device}`, every message on an unmapped sub-topic produces an `error!()` log. At high message rates this floods logs. `warn!` or `debug!` would be more appropriate.

**10. Unused `event_loop_handle` field** (`connection.rs`)
`MqttConnection::event_loop_handle` is always `None` — never set or read. Can be removed.

---

### 🟢 Low Priority / Suggestions

**11. `with_topic()` replaces vs `with_topic_config()` appends** (`lib.rs`)
`with_topic()` sets `self.topics = vec![...]` (replaces all), while `with_topic_config()` calls `push()` (appends). This inconsistency could surprise users — consider making both append, or documenting `with_topic()` as "set single topic, replacing all others."

**12. `InjectId` uses string enum instead of bool** (`config.rs`)
Users write `inject_id: "true"` instead of `inject_id: true`. A plain `bool` would be more natural in YAML/JSON configs.

**13. Unreachable match arms**
- `convert_mqtt_to_source_change` in `schema.rs`: `MqttSourceChange` only has `Update`, so the `_ =>` arm is dead code.
- `generate_packet_properties` in `pattern.rs`: The `PayloadAsField` fallback matches on a `Value::String` it just constructed — the non-`String` arm is unreachable.

**14. Copyright year** — All files say 2026. Should this be 2025?

**15. Inject-over-payload precedence** — In `PayloadSpread` mode, injected topic variables silently overwrite same-named payload fields. The test confirms this behavior, which is fine, but worth documenting in the DSL reference so users aren't surprised.

---

### 📋 Integration Test Coverage

The integration tests do a great job covering connectivity scenarios (auth, mTLS, reconnection, protocol fallback). Two tests (`test_mqtt_query_results` and `test_mqtt_query_results_with_using_input_data`) verify the full source → query pipeline, which is great to see.

A couple of gaps worth noting:
- **No test verifies reaction output** — the `ApplicationReaction` is wired in but never asserted on. It would strengthen confidence to verify the reaction actually receives the expected change notifications.
- **9 of 11 tests only check `ComponentStatus::Running`** — they don't publish messages or verify processing. Two even have TODO comments acknowledging this.

---

### ✅ What's Working Well

I want to call out some things that are done really well:
- **Protocol negotiation** — the v5 → v3.1.1 fallback with early break on `UnsupportedProtocolVersion` is clean
- **TLS security** — no insecure defaults, CA cert required, server cert validation enforced
- **Config validation** — placeholder syntax checking, cross-field consistency, `deny_unknown_fields` on serde structs
- **`matchit::Router`** — great choice for topic pattern matching
- **Topic-to-graph DSL** — the `PayloadAsField`/`PayloadSpread` modes cover the core IoT use cases well
- **Testcontainers setup** — `MosquittoGuard` with cert generation is solid test infrastructure

Thanks again for the effort here — looking forward to seeing this land! 🙏

ahmed-kamal2004:
@amansinghoriginal @danielgerlag Hi, all the issues should be resolved now.

I added too a new integration test that uses the application reaction to verify the full pipeline of the packet.

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,572 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use core::error;
+use std::time::Duration;
+
+use crate::config::{
+    self, default_base_retry_delay_secs, default_max_retries, MqttQoS, MqttSourceConfig,
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `core::error` and `config::self` imports.

components/sources/mqtt/src/processor.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl MqttProcessor {
+    //...... public methods
+
+    pub fn new(config: &MqttSourceConfig) -> Self {
+        Self {
+            mapper: Arc::new(PatternMatcher::new(&config.topic_mappings)),
+            adaptive: config.adaptive_enabled.unwrap_or(false),
+        }
+    }
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `adaptive` field.

components/sources/mqtt/src/schema.rs
```diff
@@ -0,0 +1,355 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use anyhow::Result;
+use core::time;
+use drasi_core::models::{ElementMetadata, ElementReference, SourceChange};
+use log::error;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+use std::vec::Vec;
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `core::time`, `ElementMetadata`, `ElementReference` imports.

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,866 @@
...
+use matchit::{Params, Router};
+
+use crate::schema::MqttSourceChange;
+use crate::utils::MqttPacket;
+use crate::FROM_TO_STARTS_WITH;
+use crate::{
+    config::{InjectId, MappingMode, MappingNode, MappingRelation, TopicMapping},
+    MqttElement,
+};
+use log::{error, info};
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `HashMap` and `FROM_TO_STARTS_WITH` imports.

components/sources/mqtt/src/mqtt.rs
```diff
@@ -0,0 +1,572 @@
...
+use drasi_batching_common::AdaptiveBatchConfig;
+use log::{error, info};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::{mpsc, RwLock};
+use tokio::task::JoinHandle;
+use tokio::time::timeout;
+
+use drasi_lib::channels::{ComponentType, *};
+use drasi_lib::SourceRuntimeContext;
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `identity` and `ComponentType` imports.

components/sources/mqtt/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+
+use anyhow::Result;
+use drasi_batching_common::AdaptiveBatchConfig;
+use log::error;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::mpsc;
+
+use drasi_lib::channels::{ComponentType, *};
+use drasi_lib::{Source, SourceRuntimeContext};
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `SourceSubscriptionSettings`, `ComponentType`, `mpsc`, `Source`, `SourceRuntimeContext`, `error`, `Arc` imports.

components/sources/mqtt/tests/integration_tests.rs
```diff
@@ -0,0 +1,1435 @@
...
+    MqttSourceBuilder,
+};
+use log::info;
+use rumqttc::QoS;
+use serial_test::serial;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::{thread::sleep, time::Duration};
+use tokio::time::Instant;
+
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `identity::{self}`, `IdentityProvider`, `Source`, `MqttSourceBuilder`, `QoS`, `sleep` imports.

components/sources/mqtt/tests/mosquitto_helpers.rs
```diff
@@ -0,0 +1,409 @@
...
+//! Mosquitto test helpers for MQTT integration tests.
+//!
+//! Provides utilities to start and manage a Mosquitto MQTT broker using testcontainers.
+
+use anyhow::{anyhow, Result};
+use drasi_source_mqtt::config::{MqttQoS, MqttSourceConfig, MqttTopicConfig, MqttTransportMode};
+use log::{debug, info};
+use rcgen::{BasicConstraints, CertificateParams, DnType, IsCa, KeyPair, SanType};
+use rumqttc::v5::{AsyncClient as AsyncClientV5, MqttOptions as MqttOptionsV5};
+use rumqttc::{AsyncClient, MqttOptions, QoS};
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed unused `MqttQoS`, `MqttSourceConfig`, `MqttTopicConfig`, `MqttTransportMode`, `QoS` imports.

examples/lib/mqtt-source-example/main.rs
```diff
@@ -0,0 +1,196 @@
...
+    });
+
+    println!("\n┌────────────────────────────────────────────┐");
+    println!("│ Temperature Monitor Started!               │");
+    println!("├────────────────────────────────────────────┤");
+    println!("│ MQTT Source: localhost:9001                │");
+    println!("│   Topic: sensors/temperature               │");
+    println!("├────────────────────────────────────────────┤");
+    println!("│ Results API: http://localhost:8080         │");
+    println!("│   GET /queries/all-readings/results        │");
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — banner now shows correct port 8883 and query ID `floor_readings`.

examples/lib/mqtt-source-example/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "mqtt-source-example"
+version = "0.1.0"
+edition = "2021"
+publish = false
+description = "DrasiLib example demonstrating stock price monitoring with 3 queries"
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — updated description to match MQTT example.

components/sources/mqtt/README.md
```diff
@@ -0,0 +1,400 @@
...
+      mode: payload_as_field
+      field_name: payload
+```
+
+### Structured JSON Payload Ingestion
+
+```yaml
+topic_mappings:
+  - pattern: metrics/{site}/{device}
+    entity:
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — added missing closing code fence.

components/sources/mqtt/README.md
```diff
@@ -0,0 +1,400 @@
...
+      - label: LOCATED_IN_FLOOR
+        id: "{room}_located_in_floor_{floor}"
+        from: { label: Room, id: "{room}" }
+        to: { label: Floor, id: "{floor}" }
+      - label: LOCATED_IN_GATEWAY
+        id: "{room}:{device}_located_in_gateway_$.gateway_id"
+        from: { label: Device, id: "{room}:{device}" }
+        to: { label: Gateway, id: "$.gateway_id" }
+```
+```
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed — removed stray code fence.

amansinghoriginal (approved):
looks good to me.

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+                                error_count = 0;
+                                doubler = 1;
+                            },
+                            Err(e) => {
+                                error_count += 1;
+                                if error_count >= max_retries {
+                                    error!("MQTT event loop error: {e:?}. Encountered {error_count} consecutive errors, exceeding the maximum of {max_retries}. Stopping the event loop.");
+                                    return Err(anyhow::anyhow!("MQTT event loop has encountered {error_count} consecutive errors, exceeding the maximum of {max_retries}. Stopping the event loop."));
+                                } else {
+                                    let delay = base_retry_delay_secs * doubler;
```
ruokun-niu, line 297: It looks like the exponential backoff here has no limit?

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+        if let Some(max_inflight) = config.max_inflight {
+            options.set_outgoing_inflight_upper_limit(max_inflight);
+        }
+        if let Some(clean_start) = config.clean_start {
+            options.set_clean_start(clean_start);
+        }
+
+        // v5 specific options
+        if let Some(conn_timeout) = config.conn_timeout {
+            options.set_connection_timeout(conn_timeout);
```
ruokun-niu, line 563: The config field says:
```rust
pub conn_timeout: Option<u64>, // Connection timeout in milliseconds for MQTT v5
```

But in connection.rs, it's passed to options.set_connection_timeout(conn_timeout) which expects seconds (based on the rumqttc v5 API). The default is Some(5000), which would mean 5000 seconds, not 5 seconds.

components/sources/mqtt/src/config.rs
```diff
@@ -0,0 +1,1895 @@
...
+    PayloadSpread,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub enum InjectId {
+    #[serde(rename = "true")]
+    True,
+    #[serde(rename = "false")]
+    False,
```
ruokun-niu: nit: Is there a reason why we use `"false"` instead of`false` (plain bool)

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,865 @@
...
+                let start = i + 2;
+                let mut end = start;
+                while end < bytes.len() {
+                    let c = bytes[end];
+                    if c.is_ascii_alphanumeric() || c == b'_' || c == b'.' {
+                        end += 1;
+                    } else {
+                        break;
+                    }
+                }
```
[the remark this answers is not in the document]
ahmed-kamal2004: Nested traversal is not to be implemented according to the design document.

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,865 @@
...
+                if end == start {
+                    out.push_str("$.");
+                    i += 2;
+                    continue;
+                }
+                let path = &s[start..end];
+                let value = properties.get(path).ok_or_else(|| {
+                    anyhow::anyhow!("Property '{path}' not found for payload reference")
+                })?;
```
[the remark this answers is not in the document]
ahmed-kamal2004: Nested traversal is not to be implemented according to the design document
and the (first < ) `.` is treated as a raw char without any special handling

components/sources/mqtt/src/connection.rs
ahmed-kamal2004: Oh, thanks, I missed it

components/sources/mqtt/src/config.rs
ahmed-kamal2004: from prev changes, updated it now, no change in API

components/sources/mqtt/src/connection.rs
ahmed-kamal2004: I added a new limit over the `doubler` to be at most `256` (`2^8`)

So the maximum backoff time is `base_retry_delay_secs` * `256`

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+            None
+        }
+    }
+}
+
+impl ToMqttPacket for rumqttc::v5::Event {
+    fn to_mqtt_packet(self) -> Option<MqttPacket> {
+        if let rumqttc::v5::Event::Incoming(rumqttc::v5::mqttbytes::v5::Packet::Publish(p)) = self {
+            Some(MqttPacket {
+                topic: str::from_utf8(&p.topic).unwrap_or_default().to_string(),
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

components/sources/mqtt/src/schema.rs
```diff
@@ -0,0 +1,354 @@
...
+    match value {
+        serde_json::Value::Null => Ok(ElementValue::Null),
+        serde_json::Value::Bool(b) => Ok(ElementValue::Bool(*b)),
+        serde_json::Value::Number(n) => {
+            if let Some(i) = n.as_i64() {
+                Ok(ElementValue::Integer(i))
+            } else if let Some(f) = n.as_f64() {
+                Ok(ElementValue::Float(OrderedFloat(f)))
+            } else {
+                Err(anyhow::anyhow!("Unsupported number type"))
```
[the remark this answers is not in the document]
ahmed-kamal2004: Updated it in another way

components/sources/mqtt/src/processor.rs
```diff
@@ -0,0 +1,303 @@
...
+                        // generate source changes from the packet and topic name
+                        let source_changes = Self::process(&matcher, &packet);
+
+                        // send to the batcher
+                        Self::send_to_batcher(
+                            &source_id.clone(),
+                            &batch_tx,
+                            source_changes,
+                            &cancellation_token,
+                        )
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

examples/lib/mqtt-source-example/main.rs
```diff
@@ -0,0 +1,196 @@
...
+    });
+
+    println!("\n┌────────────────────────────────────────────┐");
+    println!("│ Temperature Monitor Started!               │");
+    println!("├────────────────────────────────────────────┤");
+    println!("│ MQTT Source: localhost:8883               │");
+    println!("│   Topics: sensors/temperature, building/#  │");
+    println!("├────────────────────────────────────────────┤");
+    println!("│ Results API: http://localhost:8080         │");
+    println!("│   GET /queries/floor_readings/results      │");
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

examples/lib/mqtt-source-example/main.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let source_config: MqttSourceConfig = serde_yaml::from_str(source_config_yaml)?;
+
+    // Build MqttSource with the configuration
+    let mqtt_source = match MqttSource::builder("mqtt-source")
+        .with_config(source_config)
+        .with_identity_provider(PasswordIdentityProvider::new(
+            "drasi".to_string(),
+            "drasi".to_string(),
+        ))
```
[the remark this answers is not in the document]
ahmed-kamal2004: hmm, left as it is

examples/lib/mqtt-source-example/main.rs
```diff
@@ -0,0 +1,206 @@
...
+    let api_core = core.clone();
+    let results_api = Router::new()
+        .route("/queries/:id/results", get(get_query_results))
+        .with_state(api_core);
+
+    println!(">>> Starting Results API server on port 8080...");
+    let api_handle = tokio::spawn(async move {
+        let listener = tokio::net::TcpListener::bind("0.0.0.0:8080").await.unwrap();
+        println!(">>> Results API listening on http://0.0.0.0:8080");
+        axum::serve(listener, results_api).await.unwrap();
```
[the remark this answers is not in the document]
ahmed-kamal2004: No need

components/sources/mqtt/README.md
```diff
@@ -0,0 +1,400 @@
...
+- Two payload mapping modes for structured or raw payload ingestion.
+- Optional TCP or TLS transport.
+- Static credentials or identity-provider-based authentication
+- Adaptive batching for higher-throughput streams
+- Optional bootstrap provider when building the source programmatically
+
+## Configuration
+
+### Builder Pattern
+
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,865 @@
...
+        let mut router = Router::new();
+        for mapping in topic_mappings {
+            if let Err(e) = router.insert(&mapping.pattern, mapping.clone()) {
+                error!(
+                    "Failed to insert topic mapping pattern '{}': {}",
+                    mapping.pattern, e
+                );
+            } else {
+                info!("Inserted topic mapping pattern: {}", mapping.pattern);
+            }
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+
+        Err(anyhow::anyhow!("Failed to create MQTT client with both v5 and v3 options. Check configuration for errors."))
+    }
+
+    pub async fn run_subscription_loop(
+        &mut self,
+        mut shutdown_rx: tokio::sync::oneshot::Receiver<()>,
+        mut event_loop: MqttEventLoopWrapper,
+        config: &MqttSourceConfig,
+        mut processer_tx: tokio::sync::mpsc::Sender<MqttPacket>,
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,865 @@
...
+                }
+                if end == start {
+                    out.push_str("$.");
+                    i += 2;
+                    continue;
+                }
+                let path = &s[start..end];
+                let value = properties.get(path).ok_or_else(|| {
+                    anyhow::anyhow!("Property '{path}' not found for payload reference")
+                })?;
```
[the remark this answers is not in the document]
ahmed-kamal2004: Same conversation happened above, `.` without `$` is treated as raw char without any meaning

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+                                    error!("Failed to subscribe to topics: {e:?}");
+                                });
+                                error_count = 0;
+                                doubler = 1;
+                            },
+                            Ok(event) => {
+                                let packet = event.to_mqtt_packet();
+                                if let Some(packet) = packet {
+                                    if let Err(e) = processor_tx.send(packet).await {
+                                            error!("Failed to send MQTT packet to processor: {e:?}");
```
danielgerlag: 🔴 **Critical: Silent data loss on channel send failure**

When `processor_tx.send(packet).await` fails (e.g., processor task panicked or was aborted during shutdown), the MQTT packet is logged and silently dropped. The event loop continues processing subsequent messages as if nothing happened.

This violates QoS guarantees — even QoS 2 (exactly-once) MQTT messages can be permanently lost with no status change or operator alerting.

**Suggestion:** At minimum, transition the component to `Error` status and break out of the loop. Continuing to consume from the broker while the processor is dead means all subsequent messages are also silently lost:
```rust
if let Err(e) = processor_tx.send(packet).await {
    error!("Failed to send MQTT packet to processor: {e:?}");
    return Err(anyhow::anyhow!("Processor channel closed: {e}"));
}
```
The same issue exists in the v3 event loop below (line ~333).

components/sources/mqtt/src/mqtt.rs
```diff
@@ -0,0 +1,570 @@
...
+            )
+            .await;
+
+        // shutdown signal to the connection task
+        if let Some(tx) = self.base.shutdown_tx.write().await.take() {
+            let _ = tx.send(());
+        }
+
+        // stop the processor and adaptive batcher tasks by triggering the cancellation token
+        self.shutdown_token.cancel();
```
danielgerlag: 🔴 **Critical: `CancellationToken` not reset after `stop()` — source broken on restart**

`self.shutdown_token` is a field on `MqttSource`. It's cancelled here during `stop()`, but **never replaced with a new token**. If `start()` is called again after `stop()` (e.g., during a reconnection cycle or manual restart), the processor and batcher tasks receive an already-cancelled token and will exit immediately on their first `tokio::select!` iteration.

The source will appear to start (status goes to Running) but silently do nothing — no messages will be processed.

**Fix:** Replace the token before starting new tasks in `start()`:
```rust
// In start(), before spawning tasks:
self.shutdown_token = CancellationToken::new();
```
Note: this requires changing `shutdown_token` or using an interior-mutable pattern since `start(&self)` takes `&self` not `&mut self`. Consider wrapping in `RwLock<CancellationToken>` or creating a fresh token per start cycle and storing it.

components/sources/mqtt/src/mqtt.rs
```diff
@@ -0,0 +1,570 @@
...
+
+        self.base
+            .set_status(
+                ComponentStatus::Stopping,
+                Some("Stopping MQTT source".to_string()),
+            )
+            .await;
+
+        // shutdown signal to the connection task
+        if let Some(tx) = self.base.shutdown_tx.write().await.take() {
```
danielgerlag, line 495: 🟠 **High: Shutdown race — uncoordinated task teardown can lose buffered messages**

The sequence here is:
1. Line 491-493: Signal connection task to stop
2. Line 496: Cancel processor + batcher **simultaneously**
3. Lines 499-508: Await connection task (3s timeout)
4. Lines 510-522: Await processor (3s timeout), then **abort**

Problem: The connection task still holds `processor_tx` and may still be receiving MQTT messages between step 1 and its actual shutdown. Meanwhile the processor's cancellation token fires immediately (step 2), so it starts draining. If the processor drains its channel and exits before the connection task sends its final messages, those messages are lost.

Worse: if the processor times out and is aborted (line 519), any messages buffered in the `processor_rx` channel are dropped.

**Suggestion:** Consider:
- Drop `processor_tx` (close the channel) *before* cancelling the processor token, letting the processor naturally drain and exit on channel closure
- Or: cancel the connection task first and await it, *then* cancel the processor

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,869 @@
...
+                    serde_json::Value::Array(_) | serde_json::Value::Object(_) => {
+                        return Err(anyhow::anyhow!(
+                            "Property '{path}' must be a primitive value for payload reference"
+                        ));
+                    }
+                };
+                out.push_str(&stringified);
+                i = end;
+            } else {
+                out.push(bytes[i] as char);
```
danielgerlag: 🟡 **Medium: Non-ASCII corruption in `resolve_payload_refs`**

This function iterates byte-by-byte over a UTF-8 string and uses `bytes[i] as char` for non-`$.` characters. For multi-byte UTF-8 sequences (any non-ASCII character — e.g., sensor names in Chinese, accented building names, emoji), this casts individual bytes to char, producing garbled output.

Example: the string `"café"` is 5 bytes (`63 61 66 c3 a9`). The `c3` and `a9` bytes would be pushed as separate chars `'Ã'` and `'©'` instead of `'é'`.

**Fix:** Iterate over `s.chars()` (or `char_indices()`) instead of raw bytes. The `$` check can still be done on chars since `'$'` is always a single byte in UTF-8:
```rust
let mut chars = s.char_indices().peekable();
while let Some((i, c)) = chars.next() {
    if c == '$' {
        if let Some(&(_, '.')) = chars.peek() { ... }
    } else {
        out.push(c);
    }
}
```

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,869 @@
...
+        mapping: &TopicMapping,
+        params: &Params,
+        entity_id: &str,
+        timestamp: u64,
+        properties: &mut serde_json::Map<String, serde_json::Value>,
+    ) -> MqttSourceChange {
+        for inject in &mapping.properties.inject {
+            for (key, template) in inject {
+                let value = Self::map_template(template, params);
+                properties.insert(key.clone(), serde_json::Value::String(value));
```
danielgerlag, line 157: 🟡 **Medium: `inject` silently overwrites payload properties**

Injected topic variables are inserted *after* payload properties are already in the map. If the MQTT payload JSON contains a key that collides with an `inject` key (e.g., payload has `{"room": "payload-room"}` and inject maps `"room"` → `"{room}"`), the inject **silently wins** and overwrites the payload value.

Your test at line 519 actually demonstrates this — it expects `"room"` to be `"r9"` (from topic) not `"payload-room"` (from payload). This may be intentional behavior but it's undocumented and could surprise users.

**Suggestion:** Either:
1. Document this precedence clearly in the README/config docs ("inject values take precedence over payload fields with the same name")
2. Or log a warning when an inject key collides with an existing payload key

components/sources/mqtt/src/processor.rs
```diff
@@ -0,0 +1,303 @@
...
+                        ),
+                    };
+
+                    tokio::select! {
+                        _ = cancellation_token.cancelled() => {
+                            debug!("[{source_id}] Cancelled while waiting to send change {idx} to batcher");
+                            break;
+                        }
+                        send_result = batch_tx.send(change_event) => {
+                            if let Err(e) = send_result {
```
danielgerlag, line 284: 🟠 **High: Partial data loss within a single MQTT packet's generated changes**

When `batch_tx.send(change_event)` fails for one change in a multi-change batch (e.g., an entity + 2 nodes + 2 relations from a single MQTT message), the failed change is logged but the loop continues sending subsequent changes.

This means you can end up with **partial graph state** — e.g., a relation is created but its endpoint node was dropped, or the entity is missing but its relations exist. This creates inconsistent state that continuous queries will operate on.

**Suggestion:** If a send fails, either:
1. Stop processing that packet's changes entirely (all-or-nothing per MQTT message)
2. Or break out and signal an error state — partial graph updates are worse than no update at all

danielgerlag (commented):
Thanks for this thorough implementation! I've reviewed the code in depth and found several issues that should be addressed before merging — two of which are critical production bugs (silent data loss and a restart failure). See inline comments.

components/sources/mqtt/src/pattern.rs
ahmed-kamal2004: it is already done in the current readme

`inject: copies topic placeholder values into properties (overrides any payload values with the same keys in payload_spread mode)`

components/sources/mqtt/tests/integration_tests.rs
```diff
@@ -0,0 +1,1430 @@
...
+    wait_for_source_status(
+        &core,
+        &source_slot_name,
+        drasi_lib::ComponentStatus::Running,
+    )
+    .await?;
+
+    // stop the new container
+    new_guard.cleanup().await;
+    tokio::time::sleep(Duration::from_secs(1 + 2 + 4 + 8 + 16 + 32 + 64 + 10)).await; // wait for a few seconds to allow the core to attempt reconnections
```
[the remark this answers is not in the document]
ahmed-kamal2004: hmm, I don't think this needs to be changed.

components/sources/mqtt/src/config.rs
```diff
@@ -0,0 +1,1886 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub connect_properties: Option<MqttConnectProperties>, // MQTT v5 connect properties
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub subscribe_properties: Option<MqttSubscribeProperties>, // MQTT v5 subscribe properties
+
+    //...... Adaptive batching config parameters
+    /// Adaptive batching: maximum batch size
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub adaptive_max_batch_size: Option<usize>,
```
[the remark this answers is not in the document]
ahmed-kamal2004: Following the Http approach.

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+        // authentication with the identity provider (higher precedence than static credentials from config if both are provided)
+        let effective_provider = identity_provider;
+        let config_provider = config
+            .identity_provider
+            .as_ref()
+            .map(|p| Arc::from(p.clone_box()));
+
+        let provider = effective_provider.or(config_provider);
+
+        // try Mqtt v5 first
```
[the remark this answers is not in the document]
ahmed-kamal2004: not in v1 design document ( out of scope )

components/sources/mqtt/src/config.rs
```diff
@@ -0,0 +1,1886 @@
...
+    pub keep_alive: Option<u64>,
+
+    /// Clean or Persistent session for MQTT connection (default: true)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub clean_start: Option<bool>,
+
+    //...... MQTT v3 specific config parameters (if any)
+    /// Max incoming packet size
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_incoming_packet_size: Option<usize>,
```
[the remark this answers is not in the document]
ahmed-kamal2004: In the current implementation, we are relying on the `rumqttc` default values

components/sources/mqtt/src/config.rs
```diff
@@ -0,0 +1,1886 @@
...
+                    ca,
+                    alpn: alpn.clone(),
+                    client_auth,
+                })
+            }
+        }
+    }
+
+    fn read_file(path: &str, field: &str) -> anyhow::Result<Vec<u8>> {
+        fs::read(path).map_err(|e| anyhow::anyhow!("Failed to read TLS {field} at '{path}': {e}"))
```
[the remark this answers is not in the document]
ahmed-kamal2004: Path restriction is not within V1

components/sources/mqtt/src/processor.rs
```diff
@@ -0,0 +1,303 @@
...
+                    } else {
+                        info!("[{source_id}] MQTT processing loop stopped - channel closed");
+                        break;
+                    }
+                }
+            }
+        }
+    }
+
+    fn process(mapper: &PatternMatcher, packet: &MqttPacket) -> Vec<MqttSourceChange> {
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,571 @@
...
+        mut shutdown_rx: tokio::sync::oneshot::Receiver<()>,
+        mut event_loop: MqttEventLoopWrapper,
+        config: &MqttSourceConfig,
+        mut processor_tx: tokio::sync::mpsc::Sender<MqttPacket>,
+    ) -> anyhow::Result<()> {
+        self.subscribe_to_topics(config)
+            .await
+            .map_err(|e| anyhow::anyhow!("Failed to subscribe to topics: {e:?}"))?;
+
+        let mqtt_version: u8 = match &self.client {
```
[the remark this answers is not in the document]
ahmed-kamal2004: Done, using enum instead of u8

components/sources/mqtt/src/config.rs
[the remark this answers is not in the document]
danielgerlag: Why not?

components/sources/mqtt/src/processor.rs
ahmed-kamal2004: I think this problem `Partial data loss/Partial State` needs a roll back mechanism between the adaptive batcher and the  `MqttProcessor`

I am thinking about a scenario where a packet is extracted to those source changes:
`EntityA`, `NodeB`, `NodeC`, `RelationAB`, `RelationAC`

what if we got an error when sending `NodeC` source change ? given that all the presequent source changes (`EntityA`, `NodeB`) are already processed by the batcher and even may be sent to the query (within a previous batch).

I believe (please correct me if I am wrong) to solve this issue, we need to have like a rollback mechanism between the source components and even to be extended to include the query components, this allows execution of the packet source changes as `Atomic`, and eliminating the `Partial state`.

But given that changing the query behavior is not something recommended, I am suggesting another solution:
- `Commit` mechanism between the processing part and the batching part.
- The batcher collects the source changes of the same packet and publishes (to the query engine) them together.
- if the final packet source change `commit` doesn't happen within a specific period, the batcher drops the whole changes of this packet.

I believe if this issue is not a merge blocking, then it could be done in v2 (to give it more focus)

cc. @amansinghoriginal

components/sources/mqtt/src/config.rs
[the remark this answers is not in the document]
ahmed-kamal2004: @danielgerlag I believe putting restrictions on path needs something like a drasi-wide thoughts/rules about the best path to restrict to, and I didn't find anything similar in the current implementations while I was reading code or docs

that's why I am thinking about leaving it to v2 could be better to have a better view for how it should be handled in Drasi, and what paths are used across environments.

components/sources/mqtt/src/pattern.rs
ahmed-kamal2004: Done in the last commit

components/sources/mqtt/src/mqtt.rs
ahmed-kamal2004: Done, followed the second approach

components/sources/mqtt/src/mqtt.rs
ahmed-kamal2004: Done using `RwLock`

amansinghoriginal (approved):

amansinghoriginal:
@agentofreality Can we have another maintainer review this PR?
It has been pending a second approval for over a month.

components/sources/mqtt/src/pattern.rs
```diff
@@ -0,0 +1,869 @@
...
+                    } else {
+                        break;
+                    }
+                }
+                if end == start {
+                    out.push_str("$.");
+                    i += 2;
+                    continue;
+                }
+                let path = &s[start..end];
```
danielgerlag: **Byte/char index mismatch — wrong output and a potential panic.**

`bytes` is built from `s.chars().collect()`, so `start`/`end` are **char** indices, but `&s[start..end]` slices by **byte** offset. When any non-ASCII character appears in `s` before a `$.path` reference, these diverge:
- Best case: the wrong substring is extracted (e.g. `"café_$.value"` yields `".valu"` instead of `"value"`).
- Worst case: a byte offset lands inside a multibyte char and `&s[start..end]` **panics**, killing the processing task.

Fix by collecting from the `chars` vector instead of byte-slicing the original string:
```rust
let path: String = bytes[start..end].iter().collect();
let value = properties.get(&path).ok_or_else(|| { ... })?;
```
(and use `&path` consistently below). Reproduced locally with a multibyte char preceding the reference.

components/sources/mqtt/src/connection.rs
```diff
@@ -0,0 +1,579 @@
...
+        identity_provider: &Option<Arc<dyn drasi_lib::identity::IdentityProvider>>,
+    ) -> anyhow::Result<(AsyncClientV5, EventLoopV5)> {
+        let options_v5 =
+            Self::config_to_mqtt_options_v5(id, config, identity_provider.clone()).await?;
+
+        let (client_v5, mut event_loop_v5) =
+            AsyncClientV5::new(options_v5, config.event_channel_capacity);
+        for trial in 0..5 {
+            match event_loop_v5.poll().await {
+                Ok(event) => {
```
danielgerlag: **v5 connect can falsely report success.** This returns `Ok` on the *first* `poll()` that yields any `Ok` event, which is frequently the outgoing `CONNECT` packet rather than an incoming `CONNACK`. As a result, a broker that rejects authentication, or that only speaks v3.1.1 without emitting `UnsupportedProtocolVersion`, may not trigger the v3 fallback in `new()` — the failure instead surfaces later inside `run_subscription_loop`.

Consider only treating the connection as established once an `Incoming(Packet::ConnAck)` with a success return code is observed, so the v3 fallback path remains reachable. The same pattern applies to `mqttv3_connect` below.

components/sources/mqtt/src/schema.rs
```diff
@@ -0,0 +1,358 @@
...
+) -> Result<drasi_core::models::ElementValue> {
+    use drasi_core::models::ElementValue;
+    use ordered_float::OrderedFloat;
+    match value {
+        serde_json::Value::Null => Ok(ElementValue::Null),
+        serde_json::Value::Bool(b) => Ok(ElementValue::Bool(*b)),
+        serde_json::Value::Number(n) => {
+            if let Some(i) = n.as_i64() {
+                Ok(ElementValue::Integer(i))
+            } else if n.as_u64().is_some() {
```
danielgerlag: **A `u64` value above `i64::MAX` discards the entire element.** Returning `Err` here propagates up through `convert_mqtt_to_source_change`, so one out-of-range field drops the whole node/relation change rather than just that property. Large unsigned IDs/counters are plausible in IoT payloads.

Consider falling back to `ElementValue::Float(OrderedFloat(n.as_f64()...))` (matching the float branch just below), or at minimum add a comment documenting the intentional drop-on-overflow behavior.

danielgerlag (commented):
A few code-level findings on the MQTT source implementation. These are non-blocking relative to the larger branch-staleness issue, but worth addressing. Inline comments below.

ahmed-kamal2004:
@danielgerlag Please take a look

amansinghoriginal (approved):

danielgerlag (approved):

ahmed-kamal2004:
minimal fix to make the cargo fmt pass, it failed bec the rust toolchain is updated

PR #358, opened 2026-04-15
Fix plugin build failures for adaptive reactions and musl targets

ruokun-niu:
# Description

### Duplicate FFI symbols (linux-amd64)

**Issue:** The `dynamic-plugin` feature in `http-adaptive` and `grpc-adaptive` propagated to their base crates, causing `export_plugin!` to generate duplicate `#[no_mangle]` FFI entry points in both the adaptive crate and its dependency. The linker failed with multiple definition errors for `drasi_plugin_init`, `drasi_plugin_shutdown`, and `drasi_plugin_metadata`.

**Fix:** Changed `dynamic-plugin = ["drasi-reaction-http/dynamic-plugin"]` to `dynamic-plugin = []` in both adaptive crate Cargo.toml files. The base crates are only needed as libraries for their types/logic, not as plugins.

### Missing cdylib output (linux-musl-amd64 & linux-musl-arm64)

**Issue:** Rust's musl targets default to static CRT linking (`crt-static`), which makes `cdylib` unsupported. Cargo silently dropped the shared library output, producing no `.so` files.

**Fix:** Added `.cargo/config.toml` with `rustflags = ["-C", "target-feature=-crt-static"]` for both musl targets. Also rewrote `Dockerfile.cross-musl` to use the multi-stage build pattern with required dependencies.

### Silent build failures in xtask

**Issue:** `build-plugins` generated metadata JSON unconditionally and exited successfully even when no `.so` files were produced, deferring failure to the publish step with unclear errors.

**Fix:** Added fail-fast error reporting when plugin binaries are missing, with diagnostics listing the build directory contents.
## Type of change

Fixes: #issue_number

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,9 @@
+# Cross-compilation target flags (used by `cross` tool)
+#
+# musl targets default to static CRT linking, which prevents building cdylib
+# shared libraries. Disabling crt-static enables cdylib output for plugins.
+[target.x86_64-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
+
+[target.aarch64-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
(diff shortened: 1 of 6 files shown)


agentofreality (approved):

danielgerlag (commented):
Were we not going to just merge the adaptive http into the http component & the adaptive grpc into the main grpc components?

ruokun-niu:
> Were we not going to just merge the adaptive http into the http component & the adaptive grpc into the main grpc components?

Yes, but I think it should be handled in a separate PR? This PR just fixes the build issue for unblocking

danielgerlag (approved):

Issue #361, opened 2026-04-16
Replayable Sources 03 - Add Nested Transaction Support to Index Plugins


amansinghoriginal:
## Background & Context

We are implementing [Checkpoint-Based Recovery](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints) for at-least-once delivery in Drasi Lib.

A key requirement from [Query Checkpointing (doc 02) — Nested Transactions](https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Source-Checkpoints/02-query-checkpointing.md#nested-transactions-to-avoid-changes-in-core-engine) is that `drasi-lib` must write a source checkpoint into the **same database transaction** that the core engine uses for index updates. This ensures the checkpoint and index state are always consistent — if one rolls back, both roll back.

The core engine's `ContinuousQuery::process_source_change()` internally creates a `SessionGuard` that calls `begin()`/`commit()` on the shared `SessionControl`. To write additional data (the checkpoint) into that same transaction without modifying the core engine, `drasi-lib` needs to wrap core's processing in an **outer** `begin()`/`commit()` pair:

```
Lib:    session_control.begin()        ← opens real DB transaction (depth=1)
Core:     session_control.begin()      ← nested, depth++ only (depth=2)
Core:     ...index writes...
Core:     session_control.commit()     ← nested, depth-- only (depth=1)
Lib:    checkpoint_writer.stage(...)   ← writes into the still-open txn
Lib:    session_control.commit()       ← real commit (depth=0)
```

**Today, both `RocksDbSessionState::begin()` and `GarnetSessionState::begin()` return an error if a transaction is already active.** This issue adds a nesting depth counter so inner calls are no-ops.

### 1. RocksDB: `RocksDbSessionState` (`components/indexes/rocksdb/src/session_state.rs`)

Add an `AtomicU32` depth counter. Modify:

- **`begin()`**: If `active_txn` is `None`, create the real transaction and set depth=1 (current behavior). If `active_txn` is `Some`, increment depth and return `Ok(())` (new: nested no-op).
- **`commit()`**: If depth > 1, decrement depth and return `Ok(())` (nested no-op). If depth == 1, take and commit the real transaction, set depth=0 (current behavior).
- **`rollback()`**: Always take and drop the real transaction regardless of depth (a rollback at any level aborts everything). Reset depth=0.

[...]

Issue #345, opened 2026-03-31
Replayable Sources 01 - Add Sequence Stamping & Replay Contracts to `drasi-lib` Sources


amansinghoriginal:
## Background & Context
Currently, Drasi Lib provides **at-most-once delivery** between sources and queries via pure in-memory channels (`tokio::sync::mpsc`/`broadcast`). If a process crashes or a source-change processing fails, in-flight events are silently lost, leading to silent state divergence. Furthermore, persistent indexes (RocksDB/Garnet) are currently pointless for crash recovery because there is no mechanism to determine "what has this index already processed?"

To solve this, we are undertaking a major architectural redesign to implement **Checkpoint-Based Recovery** for at-least-once guarantees. See the design documents here - https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints.

The core of this protocol requires that:
1. Sources can stamp events with strictly monotonic, replayable sequence numbers (like Postgres LSNs or internal WAL counters).
2. Queries can communicate their confirmed, durably-processed positions back to the sources.
3. Sources can resume from specific positions upon subscription.

**Important Architectural Constraint:** 
Drasi-Lib must remain **as performant as possible** for volatile (in-memory) setups. Volatile sources (like transient HTTP without a WAL) will *not* assign sequence numbers (`None`). Volatile queries will not track them. We should aim for a zero-overhead addition for non-persistent pipelines.

## Scope of Work
This initial issue is strictly about laying down the **contract & plumbing changes** across `drasi-lib` to support routing sequences and position handles. (The actual implementation of the local Redb WAL and the RocksDB checkpoint staging will happen in follow-up PRs).

This work item involves the following structural changes, updating all relevant constructors, tests, and mock plugins to default to the non-durable/volatile state (`None` or `false`):

### 1. Event Wrapper Updates
Update the internal envelope that ferries events from sources to queries.
* Add `sequence: Option<u64>` to `SourceEventWrapper`.
* Ensure that the default/existing constructors initialize this to `None` so we don't break existing volatile sources.

### 2. Source Trait Updates
Sources must advertise whether they support positional replay (which orchestration will later use to validate compatibility with persistent queries).
* Add `fn supports_replay(&self) -> bool { false }` to the `Source` trait. 

### 3. Subscription Settings Updates
When a query subscribes to a source, it needs to be able to request a specific starting point and signal if it requires a feedback loop.
[...]

PR #362, opened 2026-04-16
Add Write-Ahead Log plugin contract + redb implementation (#346)

### LINKED ISSUE #346: Replayable Sources 02 - Implement Transient Source WAL (Redb) for Source Recovery
## Background & Context
We are implementing Checkpoint-Based Recovery to provide **at-least-once** delivery guarantees in Drasi Lib. 

While Log-Tailing sources (like Postgres or Kafka) can rely on their upstream systems for durability, **Transient Sources** (like HTTP webhooks or gRPC) have no upstream log. To support crash recovery for persistent queries, transient sources must optionally persist incoming events to a local Write-Ahead Log (WAL) *before* ACKing the caller and dispatching the event in-memory.

**Scope of this Issue:** 
This issue is strictly about building a reusable, highly-optimized, and isolated Write-Ahead Log (WAL) module backed by `redb` (a pure-Rust, embedded key-value store). **We will not update the actual transient source plugins (HTTP, gRPC, etc.) in this PR.** We are only building the generic infrastructure/service that these sources will optionally instantiate and use in follow-up issues.

*Note: This relies on the sequence numbering and `Arc<AtomicU64>` positional handles introduced in #345*

## Architecture & Separation of Concerns
We need a standalone struct/module (e.g., `TransientWal`) that wraps `redb`. 

**Important:** The `TransientWal` should have zero knowledge of Drasi topological concepts like `Query` subscribers, `Arc<AtomicU64>` position handles, or Tokio broadcast channels. It is strictly a bounded, query-able durable queue.

Storage layout within `redb` should be scoped per-source:
* **`events` table:** `sequence (u64)` -> `serialized SourceChange`.
* **`metadata` table:** `"counter"` -> `last sequence number (u64)` (persisted monotonic counter).

### 1. WAL Configuration Strategy
Define a generic configuration struct (e.g., `TransientWalConfig`) that sources will eventually plug into their own configs:
* `max_size_mb: u32` (defines the capacity limit).
* `on_full: OnFullStrategy` (Enum with `RejectIncoming` and `OverwriteOldest`).

### 2. Initialization & The Write Path (`append`)
* **Initialization:** Implement a constructor that takes a storage path/identifier and config. On startup, read the last counter value from the `metadata` table (or initialize to 0).
* **Write API:** Expose an `append(event: &SourceChange) -> Result<u64, WalError>` method that safely executes:
  1. Increment the monotonic sequence: `N = counter.fetch_add(1) + 1`.
  2. Persist `(N, event)` to both the `events` and `metadata` tables in a single transaction.
  3. **Fsync:** Commit the Redb transaction to disk (per-event fsync).
  4. Return `Ok(N)` so the caller plugin can stamp its in-memory event and ACK the outside producer.

### 3. Backpressure & Disk Exhaustion
Before/during the write transaction in `append`, estimate or check Redb disk usage against `max_size_mb`.
* If `on_full == RejectIncoming`: Abort and return `WalError::CapacityExhausted`. (The caller source will translate this to 503s/etc.).
* If `on_full == OverwriteOldest`: Force eviction of the oldest records to make room, creating an intentional gap in the log.

### 4. Pruning API 
Expose a simple `prune_up_to(sequence: u64) -> Result<(), WalError>` method.
* Executes: `DELETE FROM events WHERE sequence <= sequence`.
* *Note: The caller source is responsible for tracking queries and computing the min-watermark before calling this.*

### 5. Read / Replay API
Expose a method to support historical reading, e.g., `read_from(sequence: u64) -> Result<WalIterator, WalError>`.
* If `sequence` has already been pruned or is invalid, return `WalError::PositionUnavailable`.
* If valid, return an iterator or stream of deserialized `SourceChange` events starting from `sequence` through the current head of the log. 
* *Note: The caller source handles tearing down its own channels and managing the active rewinding logic.*

## Implementation Notes
* **Testability:** Since we are not plugging this into an actual source right now, comprehensive unit and integration tests using mock `SourceChange` data are required to validate the fsync path, the `max_size_mb` boundary behavior, and pruning logic.
* **Cleanup:** Expose a static/class method or trait to safely drop and delete the redb files off the filesystem entirely for a given source ID.
* **Redb File Size**: Redb sems to not have an O(1) "current table data size" metric that updates instantly on inserts. A workaround for this could be to check the actual file size on disk via standard filesystem APIs (std::fs::metadata) before an append without bogging down the write path.

amansinghoriginal:
## Summary

- Define `WalProvider` trait in `drasi-lib` as a user-pluggable component (peer to `StateStoreProvider`) — single trait with `source_id` partition key on every method
- Add `drasi-wal-redb` crate implementing `WalProvider` backed by redb, with one redb file per source
- Core domain types remain free of serde derives; serialization goes through explicit DTOs in the redb crate with a versioned `WalRecord` envelope for format evolution
- `WriteAheadLogConfig` has `max_events` (minimum 16, validated) and `CapacityPolicy` (`RejectIncoming` for backpressure, `OverwriteOldest` for availability)
- `SourceChange::Future` rejected at append; DTO omits the variant entirely

## Testing

29 tests against `Arc<dyn WalProvider>` covering basic ops, pruning, persistence across restarts, both capacity policies (including convergence after reduced `max_events` on restart), DTO roundtrip, multi-source isolation, registration idempotency and conflict, source-not-registered errors, format version mismatch, concurrent access, and large events. Full workspace clippy clean, `cargo +nightly fmt --check` passes.

Resolves #346


core/src/models/element.rs
```diff
@@ -92,15 +93,50 @@ pub fn validate_effective_from(value: ElementTimestamp) -> Result<(), String> {
...
+    }
+
+    pub fn deserialize_arc_slice_arc_str<'de, D>(
+        deserializer: D,
+    ) -> Result<Arc<[Arc<str>]>, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let vec: Vec<String> = Vec::deserialize(deserializer)?;
+        let arc_vec: Vec<Arc<str>> = vec.into_iter().map(|s| Arc::from(s.as_str())).collect();
```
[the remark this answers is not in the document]
amansinghoriginal: Leaving as-is because this is the cold replay path, not a hot path.
`Arc<str>::deserialize` forwards to `Box<str>::deserialize`, which forwards to `String::deserialize` + `into_boxed_str`, then `Arc::from(Box<str>`) allocates a new buffer. The suggested fix is equal or worse.

components/state_stores/redb/src/wal/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+        path: impl AsRef<Path>,
+        config: WriteAheadLogConfig,
+    ) -> Result<Self, WalError> {
+        let path_buf = path.as_ref().to_path_buf();
+        let path_clone = path_buf.clone();
+
+        let (db, counter_value) = tokio::task::spawn_blocking(move || {
+            let db = Database::create(&path_clone).map_err(|e| {
+                WalError::StorageError(format!("Failed to open WAL database at {path_clone:?}: {e}"))
+            })?;
```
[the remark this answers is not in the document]
amansinghoriginal: Good catch. fixed.

components/state_stores/redb/Cargo.toml
```diff
@@ -27,8 +27,11 @@ workspace = true
 
 [dependencies]
 drasi-lib.workspace = true
+drasi-core.workspace = true
 
 redb = "2.6.3"
+bincode = "1"
+serde = { version = "1", features = ["derive"] }
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed

components/state_stores/redb/src/wal/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+            return Err(WalError::InvalidEvent(
+                "SourceChange::Future events cannot be persisted to WAL".to_string(),
+            ));
+        }
+
+        // Serialize before entering the blocking closure to minimize blocking time
+        let bytes = bincode::serialize(event)
+            .map_err(|e| WalError::SerializationError(format!("bincode serialize failed: {e}")))?;
+
+        // Allocate sequence number (monotonic, gaps acceptable on txn failure)
```
agentofreality: more strictly, this is "monotonic increasing"

components/state_stores/redb/src/wal/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+                })?;
+
+                // Check capacity and apply policy
+                let current_count = events_table.len().map_err(|e| {
+                    WalError::StorageError(format!("Failed to read event count: {e}"))
+                })?;
+
+                if current_count >= max_events {
+                    match capacity_policy {
+                        CapacityPolicy::RejectIncoming => {
```
agentofreality: I wonder when we would want to reject if we reach capacity? If a SOurce is using the WAL this would basically mean the Source stopped working after a period of time wouldnt it?

components/state_stores/redb/src/wal/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+                            return Err(WalError::CapacityExhausted);
+                        }
+                        CapacityPolicy::OverwriteOldest => {
```
agentofreality: So if we end up here we iterate through the full log each time a new entry is written?

components/state_stores/redb/src/wal/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+                let mut events_table = write_txn.open_table(EVENTS_TABLE).map_err(|e| {
+                    WalError::StorageError(format!("Failed to open events table: {e}"))
+                })?;
+
+                // Check capacity and apply policy
+                let current_count = events_table.len().map_err(|e| {
+                    WalError::StorageError(format!("Failed to read event count: {e}"))
+                })?;
+
+                if current_count >= max_events {
```
agentofreality: Edge Case, but If `current_count` exceeds `max_events` by more than 1 (e.g., WAL was created with `max_events=1000`, then reopened with `max_events=500`), only one entry is evicted. The WAL remains above the configured capacity until enough one-at-a-time evictions bring it back down.

**Suggested fix:** Evict in a loop until `current_count < max_events`:

```rust
CapacityPolicy::OverwriteOldest => {
    while events_table.len()? >= max_events {
        let oldest_key = {
            let mut iter = events_table.iter()?;
            match iter.next() {
                Some(Ok(entry)) => Some(entry.0.value()),
                _ => None,
            }
        };
        if let Some(key) = oldest_key {
            events_table.remove(key)?;
        } else {
            break;
        }
    }
}
```

components/state_stores/redb/src/wal/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+
+impl WriteAheadLog {
+    /// Open or create a WAL at the given path.
+    ///
+    /// On startup, the counter is restored from the metadata table. If the
+    /// database does not exist, a new one is created with counter starting at 0.
+    pub async fn new(
+        path: impl AsRef<Path>,
+        config: WriteAheadLogConfig,
+    ) -> Result<Self, WalError> {
```
agentofreality: We should put some lower limit on the max_events. 

e.g. If max_events = 0 things are going to fail and If max_events is small things are going to thrash.

components/state_stores/redb/src/wal/config.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Maximum number of events the WAL will store before applying the capacity policy.
+    pub max_events: u64,
+
+    /// Policy to apply when the WAL reaches `max_events`.
+    pub capacity_policy: CapacityPolicy,
+}
+
+/// Policy for handling new appends when the WAL is at capacity.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CapacityPolicy {
```
agentofreality: This is a fieldless enum — it could derive `Copy`, which avoids the need to `.clone()`

agentofreality (changes requested):
Why isnt the WAL implemented using a trait following the pattern of all other plugins? 

Also, currently the component/state_store path has implications that redb is a state_store (currently intended for simply key/value state); now it is also a WAL store.

core/src/models/source_change.rs
```diff
@@ -12,11 +12,13 @@
...
 // limitations under the License.
 
+use serde::{Deserialize, Serialize};
+
 use crate::interface::FutureElementRef;
 
 use super::{Element, ElementMetadata, ElementReference, ElementTimestamp};
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
```
danielgerlag: We should avoid serializing internal domain objects, as it makes it hard for them to change / evolve in the future because there may be serialized versions of them out there. We should create corresponding DTOs to make the serialization format explicit and version-able. This applies to everything under the models folder in core,.

danielgerlag (changes requested):

core/src/models/element.rs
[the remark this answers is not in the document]
amansinghoriginal: Moot now. Core types no longer implement Serialize/Deserialize. This function is gone, and serialization happens via DTOs in the new `drasi-wal-redb` crate.

components/state_stores/redb/src/wal/mod.rs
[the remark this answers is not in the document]
amansinghoriginal: Moot now. This file is gone. The WAL moved to the new `components/wals/redb/` crate, which declares tokio with `rt-multi-thread` explicitly.

components/state_stores/redb/Cargo.toml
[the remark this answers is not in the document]
amansinghoriginal: Moot now. `components/state_stores/redb/Cargo.toml` is reverted to pre-PR state. The WAL moved to its own crate.

components/state_stores/redb/src/wal/mod.rs
amansinghoriginal: Fixed. The `WalProvider::append` docstring now uses "monotonic-increasing sequence number".

components/state_stores/redb/src/wal/mod.rs
amansinghoriginal: Fair point. Kept `RejectIncoming` as the default because the design doc relies on it to propagate backpressure to upstream producers (e.g., an HTTP source returns 503, producer retries). `OverwriteOldest` is the availability choice, which creates gaps for slow consumers and relies on the query recovery policy. Expanded the `CapacityPolicy` docstring to spell out the trade-off so source authors pick based on their upstream contract.

components/state_stores/redb/src/wal/mod.rs
amansinghoriginal: No, it is O(log n), not a full scan. Switched to `table.first()` in the new code so the intent is obvious.

components/state_stores/redb/src/wal/mod.rs
amansinghoriginal: Good catch, fixed. The new code loops until `table.len() < max_events` before inserting. Added `test_overwrite_oldest_converges_after_capacity_reduction` covering the exact scenario you described (reopen with smaller `max_events`).

components/state_stores/redb/src/wal/mod.rs
amansinghoriginal: Added `MIN_MAX_EVENTS = 16` and a `WriteAheadLogConfig::validate()` method. `register()` calls it and returns `WalError::InvalidConfig` for anything below the minimum. Covered by `test_rejects_invalid_config_below_min_max_events`.

components/state_stores/redb/src/wal/config.rs
amansinghoriginal: Done. `CapacityPolicy` is now `Copy`; the `.clone()` in the append path is gone.

core/src/models/source_change.rs
amansinghoriginal: Done. Reverted all serde derives on the core model types (`SourceChange`, `Element`, `ElementMetadata`, `ElementReference`, `ElementValue`, `ElementPropertyMap`, `FutureElementRef`). The redb WAL crate now owns its wire format via DTOs in `components/wals/redb/src/dto.rs`, with a `WalRecord` envelope that carries an explicit version field so the format can evolve independently from the core model.

amansinghoriginal:
Thanks for the detailed feedback, both of you. Significant rearchitecture in the latest push:

**Architecture:**
- Introduced a `WalProvider` trait in `drasi-lib` following the `StateStoreProvider` plugin pattern (single trait, `source_id` as partition key).
- Moved the redb implementation out of `components/state_stores/redb/` into a new `drasi-wal-redb` crate under `components/wals/`. The state-store crate is back to state-store only.
- Serialization now goes through explicit DTOs in the redb WAL crate, wrapped in a `WalRecord` envelope with an explicit version field. Core domain types (`SourceChange`, `Element`, etc.) no longer implement `Serialize`/`Deserialize`, so the core model can evolve independently from the WAL wire format.

**Inline fixes:**
- `CapacityPolicy` derives `Copy` (no more `.clone()` in the append hot path).
- `MIN_MAX_EVENTS = 16` with `WriteAheadLogConfig::validate()` called in `register()`.
- `OverwriteOldest` uses a loop until under capacity (handles reduced `max_events` across restarts).
- `table.first()` instead of `table.iter().next()` for clarity.
- `append` docstring uses "monotonic-increasing sequence number".
- `CapacityPolicy` docstring expanded with explicit backpressure vs availability trade-off.

The new crate has 29 tests (all passing) including new coverage for the edge cases called out in the review. Zero clippy warnings across all changed crates.

@agentofreality @danielgerlag ready for another look when you have time.

components/wals/redb/src/provider.rs
```diff
@@ -0,0 +1,478 @@
...
+    /// redb file at `{root_dir}/{source_id}.redb`.
+    pub fn new(root_dir: impl AsRef<Path>) -> Self {
+        Self {
+            root_dir: root_dir.as_ref().to_path_buf(),
+            states: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    fn wal_path(&self, source_id: &str) -> PathBuf {
+        self.root_dir.join(format!("{source_id}.redb"))
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed. Added `validate_source_id()` that rejects empty, `.`-prefixed, path separators, and anything outside `[A-Za-z0-9_-]`. Called from `register()` and `delete_wal()`. Covered by `test_register_rejects_invalid_source_ids` and `test_delete_wal_rejects_invalid_source_id`.

components/wals/redb/src/provider.rs
```diff
@@ -0,0 +1,478 @@
...
+            if let Some(existing) = states.get(source_id) {
+                if existing.config.max_events == config.max_events
+                    && existing.config.capacity_policy == config.capacity_policy
+                {
+                    return Ok(());
+                } else {
+                    return Err(WalError::SourceAlreadyRegistered(source_id.to_string()));
+                }
+            }
+        }
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed. `register()` now holds the write lock across the full check → open → insert flow. Registration is a cold per-source path, so serializing it globally is acceptable. Added `test_concurrent_register_same_source_is_safe` (16-way concurrent register) and `test_concurrent_register_conflicting_configs_errors_once`.

lib/src/wal/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Write-Ahead Log plugin contract.
+//!
+//! Defines the [`WalProvider`] trait for durable event logging used by
+//! transient sources (HTTP, gRPC, etc.) for crash recovery. External crates
+//! (e.g., `drasi-wal-redb`) provide concrete implementations.
+
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed, updated the PR description.

components/wals/redb/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+[dependencies]
+drasi-lib.workspace = true
+drasi-core.workspace = true
+
+redb = "2.6.3"
+bincode = "1"
+serde = { version = "1", features = ["derive"] }
+ordered-float = "3.7"
+async-trait = "0.1"
+thiserror = "2.0"
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed, removed it.

lib/src/wal/error.rs
```diff
@@ -0,0 +1,67 @@
...
+    #[error("WAL capacity exhausted for source '{0}'")]
+    CapacityExhausted(String),
+
+    /// The requested replay position has been pruned or never existed.
+    #[error("Position {requested} unavailable for source '{source_id}' (oldest available: {oldest_available:?})")]
+    PositionUnavailable {
+        source_id: String,
+        requested: u64,
+        oldest_available: Option<u64>,
+    },
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed. Narrowed the variant doc to just the pruned case and added a note that reads past head return an empty result rather than an error, matching `read_from`'s actual behavior.

components/wals/redb/src/provider.rs
```diff
@@ -0,0 +1,517 @@
...
+
+            match read_txn.open_table(METADATA_TABLE) {
+                Ok(table) => match table.get(COUNTER_KEY) {
+                    Ok(Some(value)) => {
+                        let bytes = value.value();
+                        if bytes.len() == 8 {
+                            u64::from_le_bytes(bytes.try_into().unwrap_or([0u8; 8]))
+                        } else {
+                            0
+                        }
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed. `open_or_create_db` now returns `StorageError` when the persisted counter bytes aren't exactly 8 long, instead of silently resetting to 0. Added `test_corrupt_counter_errors_on_open`.

components/wals/redb/src/provider.rs
```diff
@@ -0,0 +1,517 @@
...
+
+                let current_count = events_table
+                    .len()
+                    .map_err(|e| WalError::StorageError(format!("read event count failed: {e}")))?;
+
+                if current_count >= max_events {
+                    match capacity_policy {
+                        CapacityPolicy::RejectIncoming => {
+                            return Err(WalError::CapacityExhausted(source_id_owned));
+                        }
```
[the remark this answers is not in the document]
amansinghoriginal: Good catch, fixed. Moved `counter.fetch_add` inside the transaction, after the capacity check passes. Rejected appends no longer burn a sequence number. Added `test_rejected_append_does_not_consume_sequence` covering fill, reject, prune, re-append with assertion that the next sequence is contiguous.

danielgerlag (approved):

amansinghoriginal:
@agentofreality Can we proceed with this one?

danielgerlag (commented):
Could you add some context around how/where this fits into the bigger picture?

amansinghoriginal:
> Could you add some context around how/where this fits into the bigger picture?

Does the background-and-context section in the Issue not cover it? - https://github.com/drasi-project/drasi-core/issues/346

danielgerlag:
> > Could you add some context around how/where this fits into the bigger picture?
> 
> Does the background-and-context section in the Issue not cover it? - #346

That paragraph helps, and should be included directly in the PR but it is not enough, there are lots of moving parts to this, we need a higher level map

amansinghoriginal:
> > > Could you add some context around how/where this fits into the bigger picture?
> > 
> > 
> > Does the background-and-context section in the Issue not cover it? - #346
> 
> That paragraph helps, and should be included directly in the PR but it is not enough, there are lots of moving parts to this, we need a higher level map

But you've already approved this PR right?
We are pointing all issues to the high level design docs already - https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints

danielgerlag:
> > > > Could you add some context around how/where this fits into the bigger picture?
> > > 
> > > 
> > > Does the background-and-context section in the Issue not cover it? - #346
> > 
> > 
> > That paragraph helps, and should be included directly in the PR but it is not enough, there are lots of moving parts to this, we need a higher level map
> 
> But you've already approved this PR right? We are pointing all issues to the high level design docs already - https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints

Reviewer UX is poor, hard to reason about the bigger picture

danielgerlag:
## Where PR 362 Fits in the Bigger Picture

**The problem:** Drasi currently has **at-most-once delivery** — if the process crashes, in-flight events between sources and queries are lost. Persistent indexes become stale with no way to recover.

**The solution:** [Checkpoint-based recovery](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints) — sources stamp events with sequence numbers, queries persist checkpoints atomically alongside index updates, and on restart sources replay from the last confirmed position.

**Two source categories need different treatment:**
- **Log-tailing sources** (Postgres, Kafka) already have an upstream durable log to replay from
- **Transient sources** (HTTP webhooks, gRPC) have *no* upstream log — they need a **local WAL** to persist events before ACKing the caller

**PR 362 builds that local WAL** — a `WalProvider` trait in `drasi-lib` plus a redb-backed implementation. It is strictly infrastructure; no source plugins are modified yet.

## Dependency Graph (from [#392](https://github.com/drasi-project/drasi-core/issues/392))

```
#345 Sequence stamping ✅ ──────┐
                                ├──► #366 SourceBase resume ──┐
#346 Transient WAL (THIS PR) 🔵─┤                             ├──► #368 WAL integration (HTTP/gRPC)
                                │                             │
#361 Nested transactions ✅ ─────┼──► #367 CheckpointWriter ──┤
                                │                             ├──► #370 Query processor integration
#363 Checkpoint types ✅ ────────┘                             │
                                                              └──► #371 Orchestration & recovery
```

PR 362 is a **leaf prerequisite** — it unblocks #368 (plugging the WAL into HTTP/gRPC sources) but has no downstream integration of its own. The WAL is a bounded, queryable durable queue with zero knowledge of queries or subscriptions.

amansinghoriginal:
> > > > > Could you add some context around how/where this fits into the bigger picture?
> > > > 
> > > > 
> > > > Does the background-and-context section in the Issue not cover it? - #346
> > > 
> > > 
> > > That paragraph helps, and should be included directly in the PR but it is not enough, there are lots of moving parts to this, we need a higher level map
> > 
> > 
> > But you've already approved this PR right? We are pointing all issues to the high level design docs already - https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints
> 
> Reviewer UX is poor, hard to reason about the bigger picture

Aren't agents able to fetch the content from design-documents repo?
Otherwise, we keep duplicating text content everywhere - Design docs -> Issues -> PR.

amansinghoriginal:
> ## Where PR 362 Fits in the Bigger Picture
> **The problem:** Drasi currently has **at-most-once delivery** — if the process crashes, in-flight events between sources and queries are lost. Persistent indexes become stale with no way to recover.
> 
> **The solution:** [Checkpoint-based recovery](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints) — sources stamp events with sequence numbers, queries persist checkpoints atomically alongside index updates, and on restart sources replay from the last confirmed position.
> 
> **Two source categories need different treatment:**
> 
> * **Log-tailing sources** (Postgres, Kafka) already have an upstream durable log to replay from
> * **Transient sources** (HTTP webhooks, gRPC) have _no_ upstream log — they need a **local WAL** to persist events before ACKing the caller
> 
> **PR 362 builds that local WAL** — a `WalProvider` trait in `drasi-lib` plus a redb-backed implementation. It is strictly infrastructure; no source plugins are modified yet.
> 
> ## Dependency Graph (from [#392](https://github.com/drasi-project/drasi-core/issues/392))
> ```
> #345 Sequence stamping ✅ ──────┐
>                                 ├──► #366 SourceBase resume ──┐
> #346 Transient WAL (THIS PR) 🔵─┤                             ├──► #368 WAL integration (HTTP/gRPC)
>                                 │                             │
> #361 Nested transactions ✅ ─────┼──► #367 CheckpointWriter ──┤
>                                 │                             ├──► #370 Query processor integration
> #363 Checkpoint types ✅ ────────┘                             │
>                                                               └──► #371 Orchestration & recovery
> ```
> 
> PR 362 is a **leaf prerequisite** — it unblocks #368 (plugging the WAL into HTTP/gRPC sources) but has no downstream integration of its own. The WAL is a bounded, queryable durable queue with zero knowledge of queries or subscriptions.

Who is this helping? Is this for human reviewers or agents?
If agents, aren't agents able to fetch the docs from design-docs repo?

danielgerlag:
> > ## Where PR 362 Fits in the Bigger Picture
> > **The problem:** Drasi currently has **at-most-once delivery** — if the process crashes, in-flight events between sources and queries are lost. Persistent indexes become stale with no way to recover.
> > **The solution:** [Checkpoint-based recovery](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints) — sources stamp events with sequence numbers, queries persist checkpoints atomically alongside index updates, and on restart sources replay from the last confirmed position.
> > **Two source categories need different treatment:**
> > 
> > * **Log-tailing sources** (Postgres, Kafka) already have an upstream durable log to replay from
> > * **Transient sources** (HTTP webhooks, gRPC) have _no_ upstream log — they need a **local WAL** to persist events before ACKing the caller
> > 
> > **PR 362 builds that local WAL** — a `WalProvider` trait in `drasi-lib` plus a redb-backed implementation. It is strictly infrastructure; no source plugins are modified yet.
> > ## Dependency Graph (from [#392](https://github.com/drasi-project/drasi-core/issues/392))
> > ```
> > #345 Sequence stamping ✅ ──────┐
> >                                 ├──► #366 SourceBase resume ──┐
> > #346 Transient WAL (THIS PR) 🔵─┤                             ├──► #368 WAL integration (HTTP/gRPC)
> >                                 │                             │
> > #361 Nested transactions ✅ ─────┼──► #367 CheckpointWriter ──┤
> >                                 │                             ├──► #370 Query processor integration
> > #363 Checkpoint types ✅ ────────┘                             │
> >                                                               └──► #371 Orchestration & recovery
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > PR 362 is a **leaf prerequisite** — it unblocks #368 (plugging the WAL into HTTP/gRPC sources) but has no downstream integration of its own. The WAL is a bounded, queryable durable queue with zero knowledge of queries or subscriptions.
> 
> Who is this helping? Is this for human reviewers or agents? If agents, aren't agents able to fetch the docs from design-docs repo?

This is for human reviewers, myself included. It is hard to reason about each PR in isolation and links to multiple verbose design documents. This is meant to improve the reviewer UX.

agentofreality (approved):

PR #373, opened 2026-04-16
Add cross-platform FFI integration test workflow

danielgerlag:
## Summary

Add a new GitHub Actions workflow (`test-ffi.yml`) that runs the host-sdk FFI integration tests on **Linux**, **macOS**, and **Windows**.

## Motivation

The existing CI (`test.yml`) only runs FFI integration tests on Linux. Since FFI behavior (shared library loading, calling conventions, memory alignment, symbol visibility) varies across operating systems, cross-platform CI coverage helps catch platform-specific regressions early.

## What it does

1. **Builds 3 cdylib test plugins** (`drasi-source-mock`, `drasi-reaction-log`, `drasi-reaction-sse`) with `--features dynamic-plugin`
2. **Copies** them to `target/debug/plugins/` using platform-aware naming (`.so` / `.dylib` / `.dll`)
3. **Verifies** all 3 plugins exist (fails fast on build issues)
4. **Runs 42 integration tests** covering: plugin loading, metadata validation, callback wiring, vtable dispatch, lifecycle management, error handling, and drop cleanup

## Design decisions

- **No external dependencies** — the test plugins don't require protobuf, jq, or Redis
- **Matrix variables** (`lib_prefix`, `lib_ext`) handle platform-specific library naming cleanly
- **`fail-fast: false`** so all 3 platforms report independently
- **`shell: bash`** on all platforms for consistency
- Uses pinned action SHAs matching existing workflows
- Existing `test.yml` unchanged — keeps `make test-host-sdk` for Linux redundancy

### DIFF
--- a/.github/workflows/test-ffi.yml
+++ b/.github/workflows/test-ffi.yml
@@ -1,0 +1,122 @@
+name: Cross-Platform FFI Tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - release/*
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
+      - feature/*
+      - feature-lib
+      - release/*
+
+env:
+  CARGO_TERM_COLOR: always
+
+permissions:
+  contents: read
+
+jobs:
+  ffi-test:
+    name: FFI Tests (${{ matrix.name }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            name: Linux x86_64
+            lib_prefix: lib
+            lib_ext: so
+          - os: macos-latest
+            name: macOS arm64
+            lib_prefix: lib
+            lib_ext: dylib
+          - os: windows-latest
+            name: Windows x86_64
+            lib_prefix: ""
+            lib_ext: dll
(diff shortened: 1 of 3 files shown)


.github/workflows/test-ffi.yml
```diff
@@ -0,0 +1,122 @@
+name: Cross-Platform FFI Tests
+
+on:
+  workflow_dispatch:
+  push:
```
ruokun-niu: Do we need to trigger this workflow when the main gets updated? Since we always have to go through a PR, I think just having the `pull_request` condition would be enough

.github/workflows/test-ffi.yml
```diff
@@ -0,0 +1,122 @@
...
+    branches:
+      - main
+      - release/*
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
+      - feature/*
+      - feature-lib
```
ruokun-niu: Do we still need to have `feature-lib` as one of the branches

.github/workflows/test-ffi.yml
danielgerlag: sounds good

ruokun-niu (approved):

agentofreality (approved):

PR #376, opened 2026-04-18
Fix intermittent SIGSEGV in cross-platform FFI plugin tests

danielgerlag:
## Problem

The cross-platform FFI plugin tests (`.github/workflows/test-ffi.yml`) were intermittently failing with `signal: 11, SIGSEGV: invalid memory reference` on macOS and Windows. Linux always passed. Two distinct crash signatures were observed:

- **macOS** — `test_reaction_enqueue_query_result` crashed immediately after the `Stopping reaction: enqueue-test` log line.
- **Windows** — `test_e2e_cdylib_source_query_reaction` crashed during source change dispatch.

Linux's tolerance of malformed frees (glibc) was masking the same UB that macOS guard pages and Windows LFH decommit detect strictly.

## Root causes identified

Deep analysis of the FFI layer (`components/plugin-sdk/src/ffi/` and `components/host-sdk/src/`) surfaced four bug classes:

| | Bug | Symptom |
|---|---|---|
| **B** | `ReactionProxy::stop`/`Drop` explicitly dropped the FFI receiver while the plugin forwarder task was still calling back into it. | UAF when the forwarder's leaked Arc clone resolves a `.await` on a dropped channel. |
| **C** | `_callback_ctx` Arc passed to plugin via `Arc::as_ptr` could be freed while a late `instance_log_callback` was in flight on the plugin runtime. | Late callback writes into freed memory → macOS crash right after "Stopping reaction" log. |
| **E** | Several host-side `extern "C"` callbacks (state store bridge, identity bridge, bootstrap proxy, source `host_executor`) had no `catch_unwind`, plus `.expect()` calls on tokio runtime construction. | Panics unwinding across the FFI boundary → non-unwinding abort. |
| **F** | Plugin-side `extern "C"` functions (drop_fns, forwarder bodies) had no `catch_unwind`. The reaction `start_result_push_fn` forwarder was particularly fragile: a panic anywhere in the loop prevented the sentinel callback from being sent, causing the host's Drop to time out at 5 s and freeing the wrapper while the forwarder was still running. | Drop timeout → cascading UAF → SIGSEGV. |

A fifth class (Bug **A** — cross-cdylib `Box::from_raw` on `FfiSourceEvent` envelopes) was identified but deferred: both cdylibs default to the System allocator, which routes through process-global `libc` `malloc`/`free`, and changing the envelope ownership contract carries higher regression risk than the marginal safety benefit now that B/C/F address the observed crash signatures. Tracked as a follow-up.

### Host SDK (`components/host-sdk/`)

- **`proxies/reaction.rs`** — Bug B: removed `*rx_guard = None` from both `stop()` and `Drop`; the leaked push-context Arc keeps the receiver alive until the forwarder sends the exit sentinel. Bug C: replaced `Arc::as_ptr` with `Arc::into_raw` for the per-instance callback context, and `mem::forget` in `Drop` to intentionally leak the Arc — matching the existing process-leaked-cdylib lifecycle (plugins are never `dlclose`d, see `loader.rs:60-79`).
- **`proxies/source.rs`** — same Bug C fix; wrapped `host_executor` in `catch_unwind` and removed `.expect` on runtime build.
- **`state_store_bridge.rs`** — added an `ffi_guard<T,F>(default, f)` helper and wrapped all 13 `ss_*` extern "C" fns plus `ss_drop`. Removed `.expect` on tokio runtime construction.
- **`identity_bridge.rs`**, **`proxies/bootstrap_provider.rs`** — wrapped remaining extern "C" callbacks in `catch_unwind`.

### Plugin SDK (`components/plugin-sdk/src/ffi/vtable_gen.rs`)

- Added a module-level `ffi_guard<T,F>(default, f)` helper for void / non-`FfiResult`-returning extern fns (since `catch_panic_ffi` only works for `FfiResult`).
- **Drop-guard pattern** on all three spawned forwarder tasks (reaction `start_result_push_fn`, source `start_push_fn`, bootstrap `bootstrap_start_push`): a `SentinelOnDrop` struct ensures the host's exit sentinel is sent on **every** scope exit — normal completion, panic, or task drop. This eliminates the host-side 5 s Drop timeout that was the precondition for the cascading UAF.
- Wrapped each `await`ed `callback(...)` invocation inside the forwarder loops in `catch_unwind` so a host-side panic doesn't take down the forwarder before the sentinel guard can fire.
- Wrapped all 12 plugin-side `drop_fn` extern "C" callbacks (source/reaction/bootstrap/identity vtables, plugin descriptor wrappers, change-receiver/bootstrap-receiver state, and event opaque drops) in `ffi_guard` with null-pointer checks.

### Tests (`components/host-sdk/tests/integration_test.rs`)

- Added `test_reaction_start_stop_stress` — 25 iterations × 8 enqueues, exercising the full create → init → start → enqueue → stop → drop lifecycle to flush out the shutdown races that previously manifested as intermittent SIGSEGV.

## Validation

- `cargo build -p drasi-plugin-sdk -p drasi-host-sdk` — clean.
- `cargo clippy -p drasi-plugin-sdk -p drasi-host-sdk --all-targets` — clean (only one pre-existing unrelated warning).
- `cargo test -p drasi-host-sdk --test integration_test -- --test-threads=1` — **44 passed, 0 failed**, including the new stress test.
- **10 consecutive runs** of the full FFI integration test suite all green (~7.7 s each), versus the prior intermittent failures.

## Follow-ups (not in this PR)

- **Bug A**: Change `FfiSourceEvent`/`FfiBootstrapEvent` ownership contract so the plugin owns and frees the envelope after callback returns (host treats it as a borrow). Requires `PluginMetadata` core-version bump for ABI break. Low priority given libc allocator equivalence in practice.

### DIFF
--- a/components/plugin-sdk/src/ffi/vtable_gen.rs
+++ b/components/plugin-sdk/src/ffi/vtable_gen.rs
@@ -52,6 +52,21 @@ use crate::descriptor::{
 
 type LifecycleEmitterFn = fn(&str, FfiLifecycleEventType, &str);
 
+/// Wraps a closure in `catch_unwind`, returning `default` on panic.
+///
+/// Use this for `extern "C"` functions that don't return `FfiResult` (so
+/// `catch_panic_ffi` doesn't apply) and for the bodies of forwarder closures
+/// spawned on the plugin runtime — a panic in a forwarder must not prevent
+/// the sentinel callback from being sent, otherwise the host's drop path
+/// times out (5s) and leaks resources.
+#[inline]
+fn ffi_guard<T, F: FnOnce() -> T>(default: T, f: F) -> T {
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(f)) {
+        Ok(v) => v,
+        Err(_) => default,
+    }
+}
+
 // Compile-time assertions that transmute between raw pointers and callback
 // function pointers is safe (same size and alignment).
 const _: () = assert!(
@@ -456,7 +471,11 @@ pub fn build_source_vtable<T: Source + 'static>(
     }
 
     extern "C" fn drop_fn<T: Source + 'static>(state: *mut c_void) {
-        unsafe { drop(Box::from_raw(state as *mut SourceWrapper<T>)) };
+        ffi_guard((), || {
+            if !state.is_null() {
+                unsafe { drop(Box::from_raw(state as *mut SourceWrapper<T>)) };
+            }
+        });
     }
 
     let cached_id = source.id().to_string();
@@ -776,7 +795,11 @@ pub fn build_source_vtable_from_boxed(
     }
 
     extern "C" fn drop_fn(state: *mut c_void) {
-        unsafe { drop(Box::from_raw(state as *mut DynSourceWrapper)) };
+        ffi_guard((), || {
+            if !state.is_null() {
+                unsafe { drop(Box::from_raw(state as *mut DynSourceWrapper)) };
+            }
+        });
     }
 
     let cached_id = source.id().to_string();
@@ -1086,7 +1109,11 @@ pub fn build_reaction_vtable<T: Reaction + 'static>(
     }
 
     extern "C" fn drop_fn<T: Reaction + 'static>(state: *mut c_void) {
-        unsafe { drop(Box::from_raw(state as *mut ReactionWrapper<T>)) };
+        ffi_guard((), || {
+            if !state.is_null() {
+                unsafe { drop(Box::from_raw(state as *mut ReactionWrapper<T>)) };
+            }
+        });
     }
 
     let cached_id = reaction.id().to_string();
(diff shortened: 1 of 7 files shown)


components/plugin-sdk/src/ffi/vtable_gen.rs
```diff
@@ -1353,53 +1380,75 @@ pub fn build_reaction_vtable_from_boxed(
...
+            // process-global plugin runtime).
+            let arc = wrapper_arc(state);
+
+            // Bug F: Drop guard ensures the sentinel callback is ALWAYS sent
+            // — even if the forwarder task panics on an `enqueue_query_result`
+            // .await against a shutting-down priority queue, or if tokio drops
+            // the spawned task. Without this, the host's Drop times out at
+            // 5 s and leaks resources, allowing Bugs B/C to compound into a
+            // hard SIGSEGV.
+            struct SentinelOnDrop {
```
amansinghoriginal, line 1404: Nit, non-blocking: this, `SourceSentinelOnDrop`, and `BootstrapSentinelOnDrop` are structurally identical (`ctx_raw` plus a callback fn pointer, `Drop` fires the sentinel under `catch_unwind`). Could collapse into a single generic guard over the callback fn type, or a local macro. Would save ~30 lines and keep the sentinel encoding in one place. Fine as a follow-up.

amansinghoriginal (approved):
Thanks for this fix. Created a separate issue for the deferred bug #378 and assigned to you for now. This PR looks good, except one minor suggestion.

agentofreality (approved):

Issue #399, opened 2026-04-24
Resumable Reactions 05 - ReactionBase checkpoint primitives & optional standard loop


amansinghoriginal:
## Background

Part of the [Reaction Recovery](https://github.com/drasi-project/drasi-core/issues/393) initiative. See [design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints), specifically doc 02 section 1 (Primitives) and section 2.

## Work

- Define `ReactionCheckpoint` type (`sequence: u64`, `config_hash: u64`) with serde derives
- Implement checkpoint helpers on `ReactionBase`:
  - `read_checkpoint(query_id)` and `read_all_checkpoints()` (batched startup hydration via `get_many`)
  - `write_checkpoint(query_id, checkpoint)`
  - Key layout: `"checkpoint:{query_id}"` in the state store, partitioned by `store_id = reaction_id`
  - Serialization via bincode
  - Read failure on startup fails reaction startup
- Optional `run_standard_loop` helper that wraps dequeue, dedup, side-effect invocation, and checkpoint write in the correct order. Reactions needing custom scheduling skip the helper.

## Dependencies

- #398 (Reaction trait extensions & bootstrap hook)

Issue #397, opened 2026-04-24
Resumable Reactions 03 - Query bootstrap APIs (fetch_snapshot & fetch_outbox)


amansinghoriginal:
## Background

Part of the [Reaction Recovery](https://github.com/drasi-project/drasi-core/issues/393) initiative. See [design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints), specifically doc 01 section 5.

## Work

- Define `SnapshotResponse` type that delivers rows as a **stream** (not a bulk Vec), using the O(1) `im::HashMap` clone under the read lock, then streaming outside it
- Define `OutboxResponse` type (`results: Vec<QueryResult>`, `latest_sequence: u64`, `config_hash: u64`) as a bulk response (bounded by outbox capacity)
- Define `OutboxError::PositionUnavailable` with `requested`, `earliest_available`, `latest_sequence`, `config_hash`
- Add `fetch_snapshot()` and `fetch_outbox(after_sequence)` to the `Query` trait
- Implement both in `DrasiQuery`, reading from `RwLock<QueryOutputState>`
- Both APIs block (await) the query's bootstrap gate if the query is still in `Starting` status, preventing reactions from receiving partially-populated state
- Remove `as_any()` from the `Query` trait (replaced by typed APIs)

## Out of scope

Snapshot-lifetime tracking (active clone counter, oldest snapshot age, mutations-since-oldest) lands in #406 (observability) alongside the metric emission, not here.

## Dependencies

- #396 (QueryOutputState & in-memory outbox)

Issue #396, opened 2026-04-24
Resumable Reactions 02 - QueryOutputState & in-memory outbox


amansinghoriginal:
## Background

Part of the [Reaction Recovery](https://github.com/drasi-project/drasi-core/issues/393) initiative. See [design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints), specifically doc 01 section 2.

## Work

- Add `im` crate dependency to the workspace
- Define `QueryOutputState` struct containing:
  - `results: im::HashMap<u64, serde_json::Value>` (live result set keyed by `row_signature`)
  - `as_of_sequence: u64` (result sequence the snapshot reflects)
  - `outbox: VecDeque<QueryResult>` (ring buffer of recent results)
- Replace `current_results: Arc<RwLock<Vec<serde_json::Value>>>` with `Arc<RwLock<QueryOutputState>>`
- Replace O(N) Vec scans (`retain`, `position` with JSON equality) with O(1) `im::HashMap` operations keyed by `row_signature`
- Maintain in-memory result sequence counter (`last_result_seq`), increment only on non-empty diffs
- Implement ring buffer push/pop logic with configurable `outbox_capacity`
- Update both the bootstrap path and the live processing path in `lib/src/queries/manager.rs`
- Add `outbox_capacity` to `QueryConfig`

## Dependencies

- #395 (QueryResult sequence & ResultDiff row_signature)

Issue #394, opened 2026-04-24
Durable Pipeline for Drasi Lib


amansinghoriginal:
Drasi Lib currently connects sources, queries, and reactions with in-memory channels. Persistent indexes (RocksDB, Garnet) exist but are not usable for production: on restart, stale index state cannot be reconciled with source streams, and reactions have no mechanism to bootstrap from or catch up with query state.

This initiative makes the full pipeline durable and resumable:

- **Source to Query**: Sources stamp events with replayable sequences. Queries persist checkpoints atomically with index updates. On restart, queries resume from their checkpoint and sources replay from that position. ([design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Source-Checkpoints))

- **Query to Reaction**: Queries maintain a sequenced outbox and live result snapshot. Reactions can bootstrap from current state (`fetch_snapshot`), catch up from a checkpoint (`fetch_outbox`), and resume after crashes using durable checkpoints with configurable recovery policies. ([design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints))

### Dependency Graph

```mermaid
graph LR
    SR["#392 Source Recovery"] --> RR["#393 Reaction Recovery"]
```

Reaction Recovery's persistent storage layer (outbox and live_results column families, extended outer transaction) depends on Source Recovery's CheckpointWriter trait and query processor integration.

### Sub-issues

- [ ] #392 — Source Recovery (Replayable Sources)
- [ ] #393 — Reaction Recovery (Resumable Reactions)

PR #410, opened 2026-04-25
Resumable Reactions 01 - QueryResult sequence & ResultDiff row_signature

### LINKED ISSUE #395: Resumable Reactions 01 - QueryResult sequence & ResultDiff row_signature
## Background

Part of the [Reaction Recovery](https://github.com/drasi-project/drasi-core/issues/393) initiative. See the [Reaction Checkpoints design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints), specifically doc 01 section 6.

The rest of the reaction-recovery work depends on each emitted query result being individually addressable. Today, results flow through the query-to-reaction channel without a stable identity: there is no sequence number reactions can checkpoint or dedup against, and no key for the live-results map other than full-row JSON equality (an O(N) scan). The in-memory outbox, the `fetch_snapshot` and `fetch_outbox` bootstrap APIs, framework-level dedup, and the persistent `live_results` namespace all build on the pieces of identity established here.

## Work

- Add a sequence number to `QueryResult`. This becomes the value reactions persist in their checkpoint, the position the outbox is keyed by, and the `as_of_sequence` returned by the bootstrap APIs.

- Add a row signature to every `ResultDiff` variant (`Add`, `Delete`, `Update`, `Aggregation`). The core engine already stamps a signature on every `QueryPartEvaluationContext` (the path-solver binding for non-aggregating rows, the grouping-key hash for aggregations); the lib layer currently ignores it and falls back to JSON-equality matching when reconciling live results. Propagating the signature through the core-to-lib conversion lets the live-results map be keyed directly, replacing the O(N) scan with O(1) updates.

- Handle aggregation identity values. When an aggregation's `after` value equals the group's identity (`count: 0`, `sum: 0`, `avg: null`, `min`/`max: null`, `collect: []`), the group has emptied and the diff must be emitted as a removal rather than an overwrite. Without this, emptied groups would persist as zero-valued ghost rows in `live_results` indefinitely.

## Dependencies

The aggregation identity-value handling depends on drasi-project/drasi-core#384, which adds the signal needed to distinguish a real value from the group's identity.

amansinghoriginal:
Closes #395.

Foundational change for the [Reaction Recovery](https://github.com/drasi-project/drasi-core/issues/393) track in the [Durable Pipeline](https://github.com/drasi-project/drasi-core/issues/394) initiative. Full motivation and the downstream plan are in the [Reaction Checkpoints design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints) (doc 01 section 6).

Today, the query-to-reaction channel emits results with no stable row identity: no sequence number reactions can checkpoint or dedup against, and no key for the live-results map other than full-row JSON equality. The in-memory outbox (#396), the bootstrap APIs (#397), framework-level dedup (#399), and the persistent live_results namespace all depend on this identity being present.

This PR adds it. `QueryResult` gains a `sequence: u64` field. Every non-`Noop` `ResultDiff` variant gains a `row_signature: u64` field. Core already stamps a stable per-row hash on each `QueryPartEvaluationContext` (path-solver binding for non-aggregating rows, grouping-key hash for aggregations), but the lib's core-to-lib conversion was discarding it. The conversion now propagates it through to downstream consumers. The `sequence` field is initialized to `0` for now; the per-emission counter that drives it lands in #396.

## Stacked on #384

The bottom commit is the #384 fix (`fix(core): emit Removing when aggregating groups empty out`) being merged as a separate PR. Once that PR lands upstream this branch will be rebased to drop it. The #384 fix is what lets the third bullet of the #395 issue text (aggregation identity values surfacing as a removal) work without any lib-side identity-detection code, since core now emits `Removing` directly when a group empties.


danielgerlag:
## 🗺️ Where This Fits — The Bigger Picture

### Initiative: Durable Pipeline for Drasi Lib (#394)

Drasi Lib currently connects sources → queries → reactions with in-memory channels only. The **Durable Pipeline** initiative (#394) makes the full pipeline resumable in two parallel tracks:

1. **Source Recovery** (#392) — sources → queries. Sources stamp events with replayable sequences; queries persist checkpoints atomically with index updates. **Separate track, several PRs already in-flight (#377, #372, etc.).**
2. **Reaction Recovery** (#393) — queries → reactions. Queries maintain a sequenced outbox and live result snapshot; reactions can bootstrap from current state, catch up from a checkpoint, and resume after crashes. **This PR is the first in this track.**

The two tracks are independent until they converge at persistent storage (#402), where the outbox and live_results are persisted inside the same outer transaction that Source Recovery's `CheckpointWriter` uses.

### Reaction Recovery Dependency Chain

This is a 13-issue effort. This PR (#395) is the **foundation** — every other issue depends on it:

```
#384 Aggregation identity fix (core bug)
  └→ #395 QueryResult sequence & ResultDiff row_signature ◀◀◀ THIS PR
       ├→ #396 QueryOutputState & in-memory outbox
       │    └→ #397 Query bootstrap APIs (fetch_snapshot / fetch_outbox)
       │         └→ #400 Reaction startup orchestration ←←← CONVERGENCE
       │
       ├→ #405 Update simple reaction plugins (8 plugins)
       ├→ #408 Update http & grpc reaction plugins
       │
       └→ (all downstream issues use sequence + row_signature)

#398 Reaction trait extensions & bootstrap hook (no deps)
  └→ #399 ReactionBase checkpoint & dedup primitives
       └→ #400 Reaction startup orchestration

Source Recovery (#367, #370)
  └→ #402 Persistent storage traits & transaction integration
       ├→ #403 RocksDB outbox & live_results implementation
       └→ #404 Garnet outbox & live_results implementation

#400 + #396 + #399 → #401 Runtime gap detection
#396 + #399 + #400 → #406 Observability instrumentation
```

### The Problem This PR Solves

Today, the query→reaction channel has **no stable identity** for results:

- **No sequence number**: Reactions have no position to checkpoint against, no way to say "I've processed up to here, replay from here on restart."
- **No row key**: The live-results map uses full-row JSON equality (O(N) scan) instead of a stable key for O(1) lookups.

Without these, you cannot build an outbox (#396), snapshot/catch-up APIs (#397), framework-level dedup (#399), or persistent live_results (#403/#404).

### What This PR Establishes

1. **`QueryResult.sequence: u64`** — a per-emission sequence number. Initialized to `0` here; the actual counter lands in #396. This becomes the value reactions checkpoint against, the outbox key, and the `as_of_sequence` in bootstrap APIs.

2. **`ResultDiff.row_signature: u64`** — propagated from core's per-row hash (path-solver binding for non-aggregating rows, grouping-key hash for aggregations). Core already computes this; the lib conversion was discarding it. This becomes the live-results map key, replacing O(N) JSON scans with O(1) updates.

3. **Aggregation identity-value handling** (via #384 dependency) — when a group empties, the diff is emitted as a removal rather than an overwrite to a zero-valued ghost row. Without this, `live_results` would accumulate phantom entries.

### Relationship to Design Doc PR #11

[drasi-project/design-documents#11](https://github.com/drasi-project/design-documents/pull/11) (Separate Event Sequence from Source Position) affects the **source→query** pipeline, not the query→reaction pipeline. This PR is not directly impacted. However, worth noting that the design doc proposes extending `ResultSequenceCounter` to carry source position bytes — this could interact with #402 (persistent storage) where both tracks converge.

danielgerlag (changes requested):
This PR includes the commits from #409 
These are both large changes in and of themselves.
Please rebase this PR without the commits from PR409.
The commit that was intended for this PR looks pretty good, happy to approve after rebase.

agentofreality:
This PR is redundant because we will merge PRs #409 and #431

PR #422, opened 2026-04-29
fix: lossless config persistence — preserve ConfigValue envelopes, camelCase keys, and passwords

danielgerlag:
## Problem

When `drasi-server` persists configuration via `snapshot_configuration()` → `properties()`, three issues caused data loss:

1. **ConfigValue envelopes flattened** — Secret references (`{kind: secret, name: DB_PASSWORD}`) and environment variable references were resolved to plaintext, losing the original envelope structure
2. **camelCase keys converted to snake_case** — Properties like `slotName` became `slot_name` because `properties()` was serializing internal config structs (which lack `#[serde(rename_all = "camelCase")]`)
3. **Passwords stripped** — Five plugins actively removed password fields in `properties()` (a display-safety measure that became destructive when persistence started using `properties()` as data source)

## Root Cause

PR #349 changed `properties()` from hand-built HashMaps to `serde_json::to_value(&self.config)`, introducing the snake_case issue. When drasi-server PR #84 then replaced its shadow-cache persistence with `snapshot_configuration()` → `properties()` roundtripping, all three issues became data loss bugs.

## Solution

Store the original raw config JSON in each component so `properties()` can return it with ConfigValue envelopes intact:

- **`SourceBase` / `ReactionBase`**: Added `raw_config: Option<serde_json::Value>` field with `set_raw_config()` / `raw_config()` accessors
- **All 12 plugin descriptors**: Call `set_raw_config(config_json.clone())` during construction, preserving the original JSON before resolution
- **All 12 plugin `properties()` methods**: Return `raw_config` when available; fall back to DTO reconstruction for builder-created components (backward compatible)
- **Password fields**: Removed `map.remove("password")` from 5 plugins (postgres, mssql, storedproc-*)
- **camelCase fixes**: Added DTO reconstruction fallback in postgres, http, sse plugins

## Testing

- 4 new snapshot tests verifying secret/env-var/camelCase/mixed-value preservation
- All 710 drasi-lib tests pass
- All 123 plugin unit tests pass (across all 12 plugins)
- `cargo fmt` and `cargo clippy` clean

## Dependencies

drasi-server will have a companion PR adding persistence roundtrip tests.

### DIFF
--- a/components/reactions/http/src/http.rs
+++ b/components/reactions/http/src/http.rs
@@ -33,7 +33,7 @@ use drasi_lib::Reaction;
 use super::HttpReactionBuilder;
 
 pub struct HttpReaction {
-    base: ReactionBase,
+    pub(crate) base: ReactionBase,
     config: HttpReactionConfig,
 }
 
@@ -242,6 +242,10 @@ impl Reaction for HttpReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
+        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
+            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
+        }
+
         use crate::descriptor::{CallSpecDto, HttpQueryConfigDto, HttpReactionConfigDto};
         use drasi_plugin_sdk::ConfigValue;
 
--- a/components/reactions/http/src/http.rs
+++ b/components/reactions/http/src/http.rs
@@ -242,10 +242,6 @@ impl Reaction for HttpReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
-        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
-            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
-        }
-
         use crate::descriptor::{CallSpecDto, HttpQueryConfigDto, HttpReactionConfigDto};
         use drasi_plugin_sdk::ConfigValue;
 
@@ -282,10 +278,7 @@ impl Reaction for HttpReaction {
                 .collect(),
         };
 
-        match serde_json::to_value(&dto) {
-            Ok(serde_json::Value::Object(map)) => map.into_iter().collect(),
-            _ => HashMap::new(),
-        }
+        self.base.properties_or_serialize(&dto)
     }
 
     fn query_ids(&self) -> Vec<String> {
--- a/components/reactions/grpc/src/descriptor.rs
+++ b/components/reactions/grpc/src/descriptor.rs
@@ -135,7 +135,9 @@ impl ReactionPluginDescriptor for GrpcReactionDescriptor {
             builder = builder.with_metadata(key, value);
         }
 
-        let reaction = builder.build()?;
+        let mut reaction = builder.build()?;
+        reaction.base.set_raw_config(config_json.clone());
+
         Ok(Box::new(reaction))
     }
 }
--- a/components/reactions/grpc/src/grpc.rs
+++ b/components/reactions/grpc/src/grpc.rs
@@ -37,7 +37,7 @@ use crate::proto::{
 
 /// gRPC reaction that sends query results to an external gRPC service
 pub struct GrpcReaction {
-    base: ReactionBase,
+    pub(crate) base: ReactionBase,
     config: GrpcReactionConfig,
 }
 
@@ -356,6 +356,10 @@ impl Reaction for GrpcReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
+        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
+            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
+        }
+
         use crate::descriptor::GrpcReactionConfigDto;
         use drasi_plugin_sdk::ConfigValue;
 
--- a/components/reactions/grpc/src/grpc.rs
+++ b/components/reactions/grpc/src/grpc.rs
@@ -356,10 +356,6 @@ impl Reaction for GrpcReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
-        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
-            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
-        }
-
         use crate::descriptor::GrpcReactionConfigDto;
         use drasi_plugin_sdk::ConfigValue;
 
@@ -378,10 +374,7 @@ impl Reaction for GrpcReaction {
             metadata: self.config.metadata.clone(),
         };
 
-        match serde_json::to_value(&dto) {
-            Ok(serde_json::Value::Object(map)) => map.into_iter().collect(),
-            _ => HashMap::new(),
-        }
+        self.base.properties_or_serialize(&dto)
     }
 
     fn query_ids(&self) -> Vec<String> {
--- a/components/reactions/http/src/descriptor.rs
+++ b/components/reactions/http/src/descriptor.rs
@@ -159,7 +159,9 @@ impl ReactionPluginDescriptor for HttpReactionDescriptor {
             builder = builder.with_route(query_id, map_query_config(config));
         }
 
-        let reaction = builder.build()?;
+        let mut reaction = builder.build()?;
+        reaction.base.set_raw_config(config_json.clone());
+
         Ok(Box::new(reaction))
     }
 }
--- a/components/reactions/log/src/descriptor.rs
+++ b/components/reactions/log/src/descriptor.rs
@@ -128,7 +128,9 @@ impl ReactionPluginDescriptor for LogReactionDescriptor {
             builder = builder.with_route(query_id, map_query_config(config));
         }
 
-        let reaction = builder.build()?;
+        let mut reaction = builder.build()?;
+        reaction.base.set_raw_config(config_json.clone());
+
         Ok(Box::new(reaction))
     }
 }
--- a/components/reactions/log/src/log.rs
+++ b/components/reactions/log/src/log.rs
@@ -26,7 +26,7 @@ use drasi_lib::reactions::common::base::{ReactionBase, ReactionBaseParams};
 use drasi_lib::Reaction;
 
 pub struct LogReaction {
-    base: ReactionBase,
+    pub(crate) base: ReactionBase,
     config: LogReactionConfig,
 }
 
@@ -328,6 +328,10 @@ impl Reaction for LogReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
+        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
+            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
+        }
+
         use crate::descriptor::{LogReactionConfigDto, QueryConfigDto, TemplateSpecDto};
 
         fn map_spec_to_dto(spec: &crate::TemplateSpec) -> TemplateSpecDto {
--- a/components/reactions/log/src/log.rs
+++ b/components/reactions/log/src/log.rs
@@ -328,10 +328,6 @@ impl Reaction for LogReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
-        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
-            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
-        }
-
         use crate::descriptor::{LogReactionConfigDto, QueryConfigDto, TemplateSpecDto};
 
         fn map_spec_to_dto(spec: &crate::TemplateSpec) -> TemplateSpecDto {
@@ -358,10 +354,7 @@ impl Reaction for LogReaction {
             default_template: self.config.default_template.as_ref().map(map_qc_to_dto),
         };
 
-        match serde_json::to_value(&dto) {
-            Ok(serde_json::Value::Object(map)) => map.into_iter().collect(),
-            _ => HashMap::new(),
-        }
+        self.base.properties_or_serialize(&dto)
     }
 
     fn query_ids(&self) -> Vec<String> {
--- a/components/reactions/sse/src/descriptor.rs
+++ b/components/reactions/sse/src/descriptor.rs
@@ -172,7 +172,9 @@ impl ReactionPluginDescriptor for SseReactionDescriptor {
             builder = builder.with_route(query_id, map_query_config(config));
         }
 
-        let reaction = builder.build()?;
+        let mut reaction = builder.build()?;
+        reaction.base.set_raw_config(config_json.clone());
+
         Ok(Box::new(reaction))
     }
 }
--- a/components/reactions/sse/src/descriptor.rs
+++ b/components/reactions/sse/src/descriptor.rs
@@ -105,6 +105,45 @@ fn map_query_config(dto: &SseQueryConfigDto) -> crate::QueryConfig {
     }
 }
 
+impl From<&crate::TemplateSpec> for SseTemplateSpecDto {
+    fn from(spec: &crate::TemplateSpec) -> Self {
+        Self {
+            template: spec.template.clone(),
+            path: spec.extension.path.clone(),
+        }
+    }
+}
+
+impl From<&crate::QueryConfig> for SseQueryConfigDto {
+    fn from(config: &crate::QueryConfig) -> Self {
+        Self {
+            added: config.added.as_ref().map(SseTemplateSpecDto::from),
+            updated: config.updated.as_ref().map(SseTemplateSpecDto::from),
+            deleted: config.deleted.as_ref().map(SseTemplateSpecDto::from),
+        }
+    }
+}
+
+impl From<&crate::SseReactionConfig> for SseReactionConfigDto {
+    fn from(config: &crate::SseReactionConfig) -> Self {
+        Self {
+            host: Some(ConfigValue::Static(config.host.clone())),
+            port: Some(ConfigValue::Static(config.port)),
+            sse_path: Some(ConfigValue::Static(config.sse_path.clone())),
+            heartbeat_interval_ms: Some(ConfigValue::Static(config.heartbeat_interval_ms)),
+            routes: config
+                .routes
+                .iter()
+                .map(|(k, v)| (k.clone(), SseQueryConfigDto::from(v)))
+                .collect(),
+            default_template: config
+                .default_template
+                .as_ref()
+                .map(SseQueryConfigDto::from),
+        }
+    }
+}
+
 #[derive(OpenApi)]
 #[openapi(components(schemas(SseReactionConfigDto, SseQueryConfigDto, SseTemplateSpecDto,)))]
 struct SseReactionSchemas;
--- a/components/reactions/sse/src/sse.rs
+++ b/components/reactions/sse/src/sse.rs
@@ -78,7 +78,7 @@ fn pre_create_broadcasters_for_query_config(
 
 /// SSE reaction exposes query results to browser clients via Server-Sent Events.
 pub struct SseReaction {
-    base: ReactionBase,
+    pub(crate) base: ReactionBase,
     config: SseReactionConfig,
     broadcasters: Arc<tokio::sync::RwLock<HashMap<String, broadcast::Sender<String>>>>,
     task_handles: Arc<tokio::sync::Mutex<Vec<tokio::task::JoinHandle<()>>>>,
@@ -228,6 +228,11 @@ impl Reaction for SseReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
+        if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
+            return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
+        }
+
+        // Fallback for builder-created components (no descriptor / no raw_config).
         match serde_json::to_value(&self.config) {
             Ok(serde_json::Value::Object(map)) => map.into_iter().collect(),
             _ => HashMap::new(),
--- a/components/reactions/sse/src/sse.rs
+++ b/components/reactions/sse/src/sse.rs
@@ -228,15 +228,10 @@ impl Reaction for SseReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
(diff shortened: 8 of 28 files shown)


components/reactions/http/src/http.rs
```diff
@@ -242,6 +242,10 @@ impl Reaction for HttpReaction {
     }
 
     fn properties(&self) -> HashMap<String, serde_json::Value> {
```
ruokun-niu, line 244: Will this expose the passwords?

danielgerlag: it will allow passwords to be persisted to the config file, and wrap them in an env variable or secret envelope if that was used. Stripping the password broke drasi-server, on restart all passwords were lost.

ruokun-niu: Discussed in sync. resolved

ruokun-niu (approved):

ruokun-niu:
## 🔍 Prior Art Review

**PR:** fix: lossless config persistence — preserve ConfigValue envelopes, camelCase keys, and passwords

No existing solutions found that would improve upon this implementation.

The PR's approach — capturing the original `config_json` as `Option<serde_json::Value>` in `SourceBase`/`ReactionBase` before `ConfigValue` resolution, then returning it verbatim from `properties()` — is the idiomatic Rust pattern for lossless serialization round-trips. The only ecosystem alternative worth noting is `serde_json::value::RawValue` (which preserves raw JSON text rather than a parsed `Value`), but since `config_json` is already a `serde_json::Value` at the call site, switching to `RawValue` would add complexity for no practical benefit here. The existing `serde_json` dependency covers the need completely.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> Generated by [pr-prior-art-reviewer](https://github.com/drasi-project/.github/actions/runs/25396469902/agentic_workflow) · ● 408.5K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-prior-art-reviewer%22&type=issues)

ruokun-niu:
## 🧪 Testing Review

The 4 new snapshot tests cover the core happy-path scenarios well (secret envelopes, env-var envelopes, camelCase keys, JSON roundtrip). Two gaps are worth addressing.

---

🟡 **Password-field preservation is not directly tested in the fallback path**

The most user-visible fix in this PR is removing `map.remove("password")` from 5 plugins' DTO fallback path. None of the 4 new tests exercise this path. The mock types (`RawConfigSource`, `RawConfigReaction`) always have `raw_config` set and bypass the DTO fallback entirely. A regression to this code would go undetected.

Suggested test: Build a `StoredProcReaction` (or any of the 5 affected plugins) via the builder (not the descriptor), so `raw_config` is `None` and the DTO fallback runs. Call `properties()` and assert that the `"password"` key is present in the result.

---

🟡 **No test verifies that real plugin descriptors actually call `set_raw_config()`**

All 4 tests use custom mock types that implement `raw_config` directly. If a real plugin descriptor (e.g., `PostgresSourceDescriptor`, `MsSqlStoredProcReactionDescriptor`) fails to call `set_raw_config()`, the fallback DTO path runs silently and the bug recurs — but no test would catch it.

Suggested test: Invoke one real descriptor (e.g., postgres source or http reaction) end-to-end with a config containing a secret envelope, snapshot, and assert the envelope is preserved in `properties()`. This can be done with the existing test harness pattern once testcontainers are available, or with a lightweight mock config that doesn't require a live DB.

---

🔵 **`snapshot_mixed_config_values_roundtrip` doesn't include a plain-string password**

The test's `password` field is `{"kind": "secret", "name": "DB_PASSWORD"}` — a structured value. It doesn't test that a plain `"password": "hunter2"` value in `raw_config` survives the roundtrip, which is a distinct code path and the scenario described in the PR body ("passwords stripped").



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> Generated by [pr-testing-reviewer](https://github.com/drasi-project/.github/actions/runs/25396468952/agentic_workflow) · ● 447.2K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-testing-reviewer%22&type=issues)

ruokun-niu:
## 🔒 Security Review

### 🟡 Moderate — Plaintext password exposure via `properties()` fallback path

**Files:** `components/sources/postgres/src/lib.rs`, `components/sources/mssql/src/lib.rs`, `components/reactions/storedproc-mssql/src/reaction.rs`, `components/reactions/storedproc-mysql/src/reaction.rs`, `components/reactions/storedproc-postgres/src/reaction.rs`

**What changed:** Five plugins previously filtered passwords from `properties()` with `map.remove("password")`. This PR removes all five guards, and updates the test to assert that plaintext passwords *are* now present (renamed from `test_properties_does_not_expose_password` → `test_properties_includes_password`, line 516 of diff).

**Attack scenario:** `properties()` has two code paths:

1. **Descriptor-created components** (via `set_raw_config`): return the original `{kind: "secret", name: "DB_PASSWORD"}` envelope — safe, no plaintext.
2. **Builder-created components** (fallback DTO reconstruction): now expose plaintext passwords. Any consumer of `properties()` in this path — admin status APIs, dashboards, debug endpoints, or audit logs in the companion `drasi-server` — will receive the cleartext credential.

The test explicitly confirms this for the builder path:
```rust
assert_eq!(
    props.get("password"),
    Some(&serde_json::Value::String("super_secret_password".to_string()))
);
```

**Fix:** Apply a conditional filter in the DTO-reconstruction fallback: omit `password` only when it is a plaintext string (i.e., not a `{kind: "secret", ...}` or `{kind: "EnvironmentVariable", ...}` object). This preserves persistence fidelity for credential references while blocking accidental cleartext exposure:

```rust
// In the DTO-reconstruction fallback path only:
if let Some(serde_json::Value::String(_)) = map.get("password") {
    map.remove("password"); // strip plaintext; secret/env-var objects are kept
}
```

Alternatively, document that `properties()` must never be sent to any client-facing API and add a caller-side redaction step in `drasi-server` before exposing properties externally.

---

No other security concerns identified. The `raw_config` field stores pre-resolution JSON (secret references, not resolved values), which is appropriate. No `unsafe` blocks, no new dependencies, no CI/CD changes.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> Generated by [pr-security-reviewer](https://github.com/drasi-project/.github/actions/runs/25396465682/agentic_workflow) · ● 634.5K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-security-reviewer%22&type=issues)

ruokun-niu:
## 📝 Documentation Review

**PR description** is thorough and well-structured — problem, root cause, solution, and testing are all clear.

---

### Findings

🟡 **Should-Fix** — `lib/src/reactions/common/base.rs`, lines ~215–223

`ReactionBase::set_raw_config` and `ReactionBase::raw_config` have abbreviated single-line doc comments, while the equivalent `SourceBase` methods (same file pattern, `lib/src/sources/base.rs`) have fuller multi-line docs. For a public API that plugin authors may call directly, the doc comments should be consistent.

Current (`ReactionBase`):
```rust
/// Set the original raw config JSON for lossless persistence roundtrips.
pub fn set_raw_config(&mut self, config: serde_json::Value) {

/// Get the original raw config JSON, if set by a descriptor.
pub fn raw_config(&self) -> Option<&serde_json::Value> {
```

Suggested (matching `SourceBase` style):
```rust
/// Set the original raw config JSON for lossless persistence roundtrips.
///
/// Called by plugin descriptors to preserve the original config JSON
/// (including `ConfigValue` envelopes for secrets and env vars) before
/// resolution to plain values.
pub fn set_raw_config(&mut self, config: serde_json::Value) {

/// Get the original raw config JSON, if set by a descriptor.
///
/// Returns `None` for builder-created components that don't have
/// a raw config JSON (they use DTO reconstruction in `properties()`).
pub fn raw_config(&self) -> Option<&serde_json::Value> {
```

---

Everything else is adequate: removed `// Don't expose password` comments are correct (behavior changed, stale comment gone), fallback comments in `sse.rs` and `http/src/lib.rs` are clear, and inline comments in `mock.rs` accurately describe the DTO fallback path.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> Generated by [pr-docs-reviewer](https://github.com/drasi-project/.github/actions/runs/25396467949/agentic_workflow) · ● 1.1M · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-docs-reviewer%22&type=issues)

ruokun-niu:
## 🏗️ Design Review

**PR:** fix: lossless config persistence — preserve ConfigValue envelopes, camelCase keys, and passwords

---

### 🟡 Should-Fix 1 — Post-build mutation bypasses the builder pattern

**Files:** All 12 `descriptor.rs` files; all 12 `base` field declarations changed to `pub(crate)`

The fix calls `component.base.set_raw_config(config_json.clone())` after `builder.build()` in every descriptor. This requires exposing `base` as `pub(crate)` on each concrete type (e.g. `GrpcReaction`, `HttpSource`) so the descriptor module can reach through and mutate it. The builder pattern exists precisely to prevent this kind of post-construction patching.

**Suggested design:** Add `with_raw_config(json: serde_json::Value)` to each builder (or to `SourceBaseParams` / `ReactionBaseParams` directly), thread it through `build()`, and keep `base` private again. Since all builders already accept `config_json` during descriptor construction, plumbing it one more step into `build()` is mechanical. This eliminates the `pub(crate)` escalation across 12 structs.

---

### 🟡 Should-Fix 2 — New descriptor contract is unenforceable at the trait boundary

**Files:** `lib/src/sources/base.rs`, `lib/src/reactions/common/base.rs`; all `SourcePluginDescriptor` / `ReactionPluginDescriptor` implementations

The `SourcePluginDescriptor::from_config()` and `ReactionPluginDescriptor::from_config()` traits return a `Box<dyn Source>` / `Box<dyn Reaction>`. The framework caller (`lib_core.rs`, `add_source`, `add_reaction`) already has `config_json` available when it calls `from_config()`. Nothing in the trait contract enforces that implementations call `set_raw_config()`, so any new third-party or future first-party plugin will silently produce lossy persistence without a compile-time or even runtime warning.

**Suggested design:** The simplest fix is to have the framework-level caller (the place that calls `descriptor.from_config(id, config_json, ...)`) call `source.set_raw_config(config_json.clone())` on the returned `Box<dyn Source>` immediately after `from_config()` returns. This centralizes the behavior in one place (the SDK/framework layer) rather than repeating it in 12 descriptors. Requires promoting `set_raw_config` to the `Source` / `Reaction` traits, or delegating through a blanket method on the descriptor trait itself.

---

### 🔵 Nit — Duplicated early-return pattern across 12 `properties()` implementations

**Files:** All 12 `lib.rs` / `reaction.rs` implementation files

The identical 4-line pattern:
```rust
if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
    return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
}
```
appears verbatim in 12 places. If the `properties()` contract on the `Source` / `Reaction` traits were updated to provide a default implementation that delegates to `base.raw_config()` with a fallback to `self.properties_fallback()`, each plugin would only need to implement the fallback. This removes the duplication and prevents future divergence between plugins.

---

The fix correctly places `raw_config` in `SourceBase`/`ReactionBase` rather than in each plugin struct — that layering decision is sound. The two should-fix items are about hardening the boundary so the pattern stays correct as the plugin ecosystem grows.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> Generated by [pr-design-reviewer](https://github.com/drasi-project/.github/actions/runs/25396466767/agentic_workflow) · ● 828.6K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-design-reviewer%22&type=issues)

ruokun-niu:
## ✅ Correctness Review

**PR:** fix: lossless config persistence — preserve ConfigValue envelopes, camelCase keys, and passwords
**Repo:** drasi-project/drasi-core #422

---

### Summary

The approach is sound: store the raw descriptor JSON at construction time and return it verbatim from `properties()`, with a DTO-reconstruction fallback for builder-created components. Three correctness observations below.

---

🟡 **Should-Fix — `pub(crate) base` bypasses struct encapsulation crate-wide**

Files: `grpc.rs`, `http.rs`, `log.rs`, `sse.rs`, `reaction.rs` (×3) and `mock.rs`

Making `base` `pub(crate)` allows any code in the crate to call `base.set_raw_config()`, `base.set_status()`, `base.set_auto_start()`, etc. without going through the owning struct's API. A future contributor could silently corrupt component state. A narrower fix exposes a single forwarding method on the owning struct instead:

```rust
// Before (in each struct)
pub(crate) base: ReactionBase,

// After — keep `base` private, add one method
fn set_raw_config(&mut self, config: serde_json::Value) {
    self.base.set_raw_config(config);
}
```

The descriptor then calls `reaction.set_raw_config(config_json.clone())` — same effect, encapsulation preserved. Note: `MsSqlSource` and `PostgresReplicationSource` correctly avoid this by keeping `base` private and relying on Rust's child-module access rules instead.

---

🟡 **Should-Fix — Fallback DTO path now emits plaintext passwords for builder-created components**

Files: `mssql/src/lib.rs`, `storedproc-mssql/src/reaction.rs`, `storedproc-mysql/src/reaction.rs`, `storedproc-postgres/src/reaction.rs`, `postgres/src/lib.rs`

Removing `map.remove("password")` is correct for the descriptor-created path (the raw_config guard returns first, so the fallback is never reached when raw_config is set). However, for **builder-created** components—where there is no raw_config and the fallback DTO path executes—`password` is now emitted as a `ConfigValue::Static("plaintext")` entry. If `snapshot_configuration()` is ever invoked on a builder-created component (e.g. in an integration test harness), the snapshot will contain a plaintext password. At minimum, document the changed contract. Alternatively, keep the guard in the fallback path only:

```rust
// In the fallback DTO path only (after the raw_config early-return)
match serde_json::to_value(&dto) {
    Ok(serde_json::Value::Object(mut map)) => {
        // For builder-created components the password is plaintext; omit it
        // to avoid leaking resolved secrets in snapshots.
        map.remove("password");
        map.into_iter().collect()
    }
    _ => HashMap::new(),
}
```

---

🔵 **Nit — Duplicated raw_config extraction pattern (×12)**

Every `properties()` implementation repeats:

```rust
if let Some(serde_json::Value::Object(map)) = self.base.raw_config() {
    return map.iter().map(|(k, v)| (k.clone(), v.clone())).collect();
}
```

Consider adding a convenience method to `SourceBase`/`ReactionBase`:

```rust
/// Returns raw_config as a HashMap if present and an Object, else None.
pub fn raw_config_as_map(&self) -> Option<HashMap<String, serde_json::Value>> {
    if let Some(serde_json::Value::Object(map)) = self.raw_config.as_ref() {
        Some(map.iter().map(|(k, v)| (k.clone(), v.clone())).collect())
    } else {
        None
    }
}
```

Then each plugin becomes a one-liner:
```rust
fn properties(&self) -> HashMap<String, serde_json::Value> {
    if let Some(map) = self.base.raw_config_as_map() { return map; }
    // ... fallback DTO reconstruction
}
```

---

No logic errors, no race conditions, no lifetime or ownership issues identified. The `clone_shared()` paths in both `SourceBase` and `ReactionBase` correctly propagate `raw_config`, so the field survives multi-thread cloning.



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `drasi.io`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> Generated by [pr-correctness-reviewer](https://github.com/drasi-project/.github/actions/runs/25396464053/agentic_workflow) · ● 544.6K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-correctness-reviewer%22&type=issues)

danielgerlag:
> ## 🔒 Security Review
> ### 🟡 Moderate — Plaintext password exposure via `properties()` fallback path
> **Files:** `components/sources/postgres/src/lib.rs`, `components/sources/mssql/src/lib.rs`, `components/reactions/storedproc-mssql/src/reaction.rs`, `components/reactions/storedproc-mysql/src/reaction.rs`, `components/reactions/storedproc-postgres/src/reaction.rs`
> 
> **What changed:** Five plugins previously filtered passwords from `properties()` with `map.remove("password")`. This PR removes all five guards, and updates the test to assert that plaintext passwords _are_ now present (renamed from `test_properties_does_not_expose_password` → `test_properties_includes_password`, line 516 of diff).
> 
> **Attack scenario:** `properties()` has two code paths:
> 
> 1. **Descriptor-created components** (via `set_raw_config`): return the original `{kind: "secret", name: "DB_PASSWORD"}` envelope — safe, no plaintext.
> 2. **Builder-created components** (fallback DTO reconstruction): now expose plaintext passwords. Any consumer of `properties()` in this path — admin status APIs, dashboards, debug endpoints, or audit logs in the companion `drasi-server` — will receive the cleartext credential.
> 
> The test explicitly confirms this for the builder path:
> 
> ```rust
> assert_eq!(
>     props.get("password"),
>     Some(&serde_json::Value::String("super_secret_password".to_string()))
> );
> ```
> 
> **Fix:** Apply a conditional filter in the DTO-reconstruction fallback: omit `password` only when it is a plaintext string (i.e., not a `{kind: "secret", ...}` or `{kind: "EnvironmentVariable", ...}` object). This preserves persistence fidelity for credential references while blocking accidental cleartext exposure:
> 
> ```rust
> // In the DTO-reconstruction fallback path only:
> if let Some(serde_json::Value::String(_)) = map.get("password") {
>     map.remove("password"); // strip plaintext; secret/env-var objects are kept
> }
> ```
> 
> Alternatively, document that `properties()` must never be sent to any client-facing API and add a caller-side redaction step in `drasi-server` before exposing properties externally.
> 
> No other security concerns identified. The `raw_config` field stores pre-resolution JSON (secret references, not resolved values), which is appropriate. No `unsafe` blocks, no new dependencies, no CI/CD changes.
> 
> Warning
> 
> Firewall blocked 1 domain
> The following domain was blocked by the firewall during workflow execution:
> 
> * `drasi.io`
> 
> > To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
> 
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "drasi.io"
> ```
> 
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
> 
> > Generated by [pr-security-reviewer](https://github.com/drasi-project/.github/actions/runs/25396465682/agentic_workflow) · ● 634.5K · [◷](https://github.com/search?q=repo%3Adrasi-project%2Fdrasi-core+%22gh-aw-workflow-call-id%3A+drasi-project%2F.github%2Fpr-security-reviewer%22&type=issues)

False positive, the config file will be saved without the password field in this case. It is up to the user to use the env variable or secret envelope.

danielgerlag:
Thanks for the review. I considered each point and am keeping the current approach for the following reasons:

**1. Post-build mutation (`set_raw_config` after build)**

Intentional. `raw_config` originates in the descriptor — it's the original JSON before resolution. The builder receives resolved values (plain strings/numbers) and doesn't naturally have the original envelope JSON. Calling `set_raw_config()` once, immediately after build, before the source is shared, is a safe two-phase initialization. Threading it through the builder adds an intermediary field to every builder for something only the descriptor sets, and creates a path where `raw_config` can silently be `None` (with a fallback that produces corrupt output).

**2. `pub(crate) base` field**

`pub(crate)` is compiler-enforced — code outside this crate cannot access it. The descriptor is in the same crate and needs one-time write access to stamp the raw config. This is the correct Rust visibility for internal access.

**3. Duplicated 3-line pattern**

Prefer explicit over clever. Each plugin's `properties()` clearly shows: "if raw_config exists, return it; otherwise serialize typed config." A helper that encapsulates this hides fallback behavior — and the fallback can silently produce wrong output (snake_case, missing fields) if raw_config is unexpectedly absent. 3 lines of obvious code in 12 files is preferable to an abstraction that masks a dangerous failure mode.

I've also added a regression test (`test_descriptor_preserves_secret_envelope`) that exercises the full descriptor path with a Secret envelope to prevent future refactoring from breaking this.

agentofreality (approved):

PR #424, opened 2026-05-01
Add HTTP bootstrapper plugin

danielgerlag:
## Summary

Adds a new HTTP bootstrap provider plugin that fetches initial state from REST APIs to populate graph queries.

### Features
- **Multi-endpoint support** — configure multiple API endpoints per bootstrapper
- **5 pagination strategies** — offset/limit, page number, cursor, Link header (RFC 5988), next-URL-in-body
- **4 authentication methods** — Bearer token, API key (header/query), HTTP Basic, OAuth2 Client Credentials with token caching
- **Multi-format parsing** — JSON, YAML, and XML response bodies (auto-detected from Content-Type or configurable override)
- **Handlebars templates** — flexible element mapping with type-preserving property resolution
- **Retry with exponential backoff** — configurable retries with overflow-safe delay calculation
- **Pagination safety** — MAX_PAGES (10,000) limit and cycle detection to prevent infinite loops
- **Full DTO schema** — 11 typed DTOs with `ToSchema` for OpenAPI, `deny_unknown_fields`, and `DtoMapper` integration
- **Config validation** — validates non-empty URLs, page sizes, required fields, and relation template constraints

### Testing
- **40 unit tests** — config deserialization, validation, content parsing, pagination strategies, template type preservation, path extraction
- **20 integration tests** — real HTTP servers via axum covering all pagination types, all auth methods, XML/YAML parsing, descriptor/DTO pipeline, unknown field rejection, validation error propagation, type preservation, label filtering, multi-endpoint, relations, and retry behavior

### Files
- `components/bootstrappers/http/` — full plugin crate
- `Cargo.toml` — workspace member added


danielgerlag:
## Review feedback addressed in commit 88a1530

All issues from this review have been addressed:

### Already fixed in prior commits:
- **Exponential backoff** (provider.rs:277) — Fixed in earlier commit with `base * 2^(attempt-1)` using `checked_shl`
- **Object conversion** (response.rs:197) — Fixed in earlier commit with recursive `ElementValue::Object` instead of JSON stringify
- **Cycle detection** (outdated) — Fixed with `HashSet<(url, params)>` tracking all seen states
- **Error body truncation** (outdated) — Fixed with 256 char limit
- **Unused `rx` variables** (lines 642, 714, 796, 894, 1066, 1288) — Already fixed in prior refactor

### Fixed in this commit:
- **Unused Cargo.toml deps** — Removed `uuid`, `jsonpath-rust`, and `reqwest` stream feature
- **Unused variable bindings** — `mut rx` → `rx`, `properties` → `..` in match patterns
- **Label filter allocation** (provider.rs:384) — Eliminated per-label `String` allocation by comparing `Arc<str>` directly against `&str` via `as_str() == &**l`

### DevSkim false positives:
The localhost references (integration_test.rs:48, 53, 370 and e2e_query_test.rs:116, 121) are in test code that binds to `127.0.0.1:0` for ephemeral test servers. These already have `// DevSkim: ignore DS137138` suppression comments where needed. The remaining unsuppressed ones are in the axum `TcpListener::bind` call which is test-only infrastructure.

danielgerlag:
## Addressing Review Findings

All findings from the drasi-reviewer bot have been addressed in commit `10c4b41`. Here is how each was resolved:

### 🔴 Critical — SSRF (Threads 8, 9)
**Fixed.** Added `validate_pagination_url()` that enforces same-host policy for pagination-followed URLs. Both `link-header` and `next-url` paginators now reject URLs whose host differs from the original endpoint. Added 3 unit tests covering: different-host rejection, same-host acceptance, and non-HTTP scheme rejection.

### 🟡 Should-Fix

| # | Finding | Resolution |
|---|---------|-----------|
| 5 | Exponential backoff unbounded | **Fixed.** Added `MAX_RETRY_DELAY = 60s` cap via `.min(MAX_RETRY_DELAY)` |
| 10 | OAuth2 error body in logs | **Fixed.** Truncated to 256 chars with `"... (truncated)"` suffix |
| 12 | Parallel array coupling | **Fixed.** Introduced `ResolvedEndpoint` struct bundling config + auth |
| 14 | Builder metadata disconnect | **Acknowledged.** This is inherent to the external-metadata pattern (no FFI changes). The builder docs make the coupling requirement explicit. |
| 15 | TemplateEngine per-call | **Fixed.** Moved to provider constructor, stored as field |
| 16 | `try_recv()` racy | **Fixed.** Changed `collect_events()` to use `rx.recv().await` (blocks until sender drops) |
| 17 | Exhausted retries untested | **Fixed.** Added `test_retries_exhausted_returns_error` integration test |
| 18 | OAuth2 caching untested | **Fixed.** Added 3 unit tests: `test_oauth2_token_caching`, `test_oauth2_token_refresh_on_expiry`, `test_oauth2_error_is_truncated` |
| 19 | README backoff description | **Fixed.** Updated to "grows exponentially (delay × 2^(attempt−1)), capped at 60 seconds" |
| 20 | `password_env: null` in example | **Fixed.** Omitted the optional field entirely |

### 🔵 Nits

| # | Finding | Resolution |
|---|---------|-----------|
| 6 | Link header comma-split | **Fixed.** Replaced `split(",")` with bracket-aware `split_link_header()` that only splits outside `<>` |
| 7 | `context_to_json` swallows errors | **Fixed.** Changed to `expect()` with descriptive message |
| 11 | No URL scheme validation | **Fixed.** Added check in `EndpointConfig::validate()` requiring `http://` or `https://` prefix |
| 13 | `Paginator` trait adds indirection | **Won't fix.** The trait keeps each pagination strategy's state machine cleanly encapsulated; the heap allocation is negligible for a bootstrap-time operation. |
| 21 | Use `serde_json_path` crate | **Won't fix for now.** Our subset (field access, array index, negative index) is 40 lines and avoids a new dependency. If we need wildcard/filter expressions later, we'll adopt the crate. |

### Thread 1–2 (copilot-pull-request-reviewer): Unused imports/vars
These were from an older revision and no longer apply to the current code.

### Thread 3 (copilot-pull-request-reviewer): JSON objects stringified
This was fixed in an earlier commit — `json_value_to_element_value` now recursively converts objects to `ElementValue::Object(ElementPropertyMap)`.

---

**Test results:** 46 unit tests + 21 integration tests = **67 total, all passing.**

agentofreality (approved):

components/bootstrappers/http/src/provider.rs
```diff
@@ -0,0 +1,458 @@
...
+        let response = builder.send().await.context("HTTP request failed")?;
+
+        if !response.status().is_success() {
+            let status = response.status();
+            let body = response
+                .text()
+                .await
+                .unwrap_or_else(|_| "Unable to read error response".to_string());
+            let truncated = if body.len() > 256 {
+                format!("{}... (truncated)", &body[..256])
```
ruokun-niu: &body[..256] is byte-indexing into a String. If byte 256 lands in the middle of a multi-byte UTF-8 codepoint (very plausible for non-ASCII error bodies), this panics. Use body.chars().take(256).collect::<String>(), or guard with body.is_char_boundary(256), or use a safe truncation helper.

components/bootstrappers/http/src/auth.rs
```diff
@@ -0,0 +1,419 @@
...
+            .context("Failed to request OAuth2 token")?;
+
+        if !response.status().is_success() {
+            let status = response.status();
+            let body = response
+                .text()
+                .await
+                .unwrap_or_else(|_| "Unable to read response".to_string());
+            let truncated = if body.len() > 256 {
+                format!("{}... (truncated)", &body[..256])
```
ruokun-niu: Similar issue to the previous comment:
&body[..256] is byte-indexing into a String. If byte 256 lands in the middle of a multi-byte UTF-8 codepoint (very plausible for non-ASCII error bodies), this panics. Use body.chars().take(256).collect::<String>(), or guard with body.is_char_boundary(256), or use a safe truncation helper.

components/bootstrappers/http/src/provider.rs
```diff
@@ -0,0 +1,458 @@
...
+
+            // Parse response body using the correct content type
+            let parsed_body = match content_parser::parse_body(&response_text, &ct) {
+                Ok(body) => body,
+                Err(e) => {
+                    error!(
+                        "Failed to parse response from '{}' as {ct:?}. Body length: {}, first 200 chars: {:?}",
+                        endpoint.url,
+                        response_text.len(),
+                        &response_text[..response_text.len().min(200)]
```
ruokun-niu: .min(200) can land mid-codepoint and panic. Use chars().take(200) or is_char_boundary check.

components/bootstrappers/http/src/pagination.rs
```diff
@@ -0,0 +1,735 @@
...
+/// Parse the Link header to find the URL with rel="next".
+/// Uses bracket-aware splitting to handle commas inside URL angle brackets.
+fn parse_link_header_next(headers: &HeaderMap) -> Option<String> {
+    let link_header = headers.get("link")?.to_str().ok()?;
+
+    // Split on commas that are outside angle brackets
+    for part in split_link_header(link_header) {
+        let part = part.trim();
+        // Check if this part has rel="next"
+        if part.contains("rel=\"next\"") || part.contains("rel='next'") {
```
ruokun-niu: nit:  Substring matching will also match rel="nextpage" if any server emits that. RFC 5988 separates parameters by ; — parse the params properly (split on ;, trim, exact-match rel="next").

components/bootstrappers/http/src/provider.rs
```diff
@@ -0,0 +1,458 @@
...
+
+use crate::auth::{self, ResolvedAuth};
+use crate::config::{EndpointConfig, HttpBootstrapConfig, HttpMethod};
+use crate::content_parser::{self, ContentType};
+use crate::pagination::{self, NextPage, Paginator};
+use crate::response;
+use crate::template_engine::TemplateEngine;
+
+/// Maximum number of pages to fetch per endpoint to prevent infinite loops.
+const MAX_PAGES: u64 = 10_000;
```
ruokun-niu: Hardcoded. For very large initial loads this is a silent ceiling — consider exposing it via HttpBootstrapConfig with the current value as default, and emitting an error! (not warn!) when reached, so it doesn’t look like clean termination in logs.

components/bootstrappers/http/src/provider.rs
danielgerlag: Fixed — added a `safe_truncate` helper that finds the nearest valid char boundary before slicing. Applied to all truncation points in `provider.rs`.

components/bootstrappers/http/src/auth.rs
danielgerlag: Fixed — added a `find_char_boundary` helper in `auth.rs` that walks backward to the nearest valid UTF-8 boundary before slicing.

components/bootstrappers/http/src/provider.rs
danielgerlag: Fixed — now uses the same `safe_truncate` helper which respects char boundaries.

components/bootstrappers/http/src/pagination.rs
danielgerlag: Good catch — replaced substring `contains` with proper RFC 5988 parsing: splits on `;`, trims each param, and does an exact match against `rel="next"`. A `rel="nextpage"` will no longer match.

ruokun-niu (approved):

PR #468, opened 2026-05-21
fix: add publish resilience for dev-dependency ordering issues

danielgerlag:
## Problem

The release workflow fails when `release-plz` attempts to publish a crate before its dev-dependencies are available on crates.io. For example, `drasi-source-open511` failed because `drasi-reaction-application` v0.3.2 hadn't been published yet:

> failed to select a version for the requirement `drasi-reaction-application = "^0.3.2"`
> candidate versions found which didn't match: 0.3.1, 0.3.0, 0.2.13, ...

This affects many source plugins that dev-depend on `drasi-reaction-application` for integration tests.

See: https://github.com/drasi-project/drasi-core/actions/runs/26258577688/job/77290459757

## Root Cause

`release-plz` determines publish order from the dependency graph but does **not** consider dev-dependencies. Since sources only depend on `drasi-reaction-application` as a dev-dep, release-plz may attempt to publish them first.

## Fix

1. **`publish_no_verify = true`** in `release-plz.toml` — skips the `cargo publish` build/dependency-resolution check during packaging. CI already validates that everything builds, so this check is redundant and was the direct cause of the failure.

2. **Retry loop (3 attempts, 30s delay)** in the workflow — provides resilience against transient failures (network issues, crates.io rate limits, index propagation delays). Since release-plz skips already-published crates via git tags, retries are idempotent and safe.


ruokun-niu (approved):
Approved. Just a quick question: What would happen if a crate is published but a git tag fails to get created?

agentofreality (approved):

PR #517, opened 2026-06-05
docs+ci: enforce shared System allocator for FFI plugins (#378)

### LINKED ISSUE #378: Cross-cdylib `Box` ownership relies on undocumented allocator uniformity
### Context
The FFI plugin system transfers many rich Rust types across the cdylib boundary via `Box::into_raw` / `Box::from_raw` pairs. This requires the allocator that produced a pointer on one side to be ABI-compatible with `dealloc` on the other side. Today this holds by coincidence: no crate in the workspace sets `#[global_allocator]`, so every allocation routes through the default System allocator (libc `malloc`/`free` on all supported targets).

This was identified as "Bug A" in #376 and deferred because the System-allocator default makes current behavior safe in practice. On investigation the pattern is broader than a single envelope type: the cross-cdylib `Box` contract applies to every FFI type that crosses the boundary, in both directions (plugin-allocated types freed by the host, and host-allocated types freed by the plugin), including event envelopes, subscription response structures, vtables themselves, and plugin registration structs.

### Risk
Today: safe. No custom allocator is configured anywhere in the workspace.

If any plugin, or the host, sets `#[global_allocator]` (jemalloc, mimalloc, tcmalloc, and others are common in perf-sensitive Rust), every cross-boundary `Box::from_raw` becomes heap corruption. The failure mode is silent; symptoms are arbitrary corruption, not a clean crash. The constraint is not stated anywhere in the plugin architecture doc, the plugin-sdk crate docs, or `PluginMetadata` validation, so a third-party plugin author has no way to know it exists.

### Proposed fixes, in order of effort

1. **Document the constraint.** Call out in the plugin architecture doc and the plugin-sdk crate docs that plugins and the host must share the System allocator.
2. **Enforce at build time.** A workspace `deny.toml` rule or CI check that rejects any crate pulling in a non-System allocator transitively.
3. **Enforce at load time.** An allocator-identity probe on plugin load: plugin publishes a canary allocation that the host tries to free, reject the plugin if the free fails. Additive `PluginMetadata` field, no ABI break.
4. **Structural fix (ABI break).** Change ownership contract so whichever side allocates an FFI envelope also frees it; the other side treats the pointer as a borrow. Requires paired `drop_fn` pointers on every FFI envelope and a `PluginMetadata` core-version bump. This is the long-term plan referenced in #376.

### Severity
Latent today, P0 the day a plugin author ships with a custom allocator and nobody notices. Silent corruption is worse than a panic. Suggested path: #1 and #2 immediately, #3 before the next plugin-sdk minor release, #4 at the next ABI-break window.

### Related
#376 defers this as Bug A.

danielgerlag:
Closes #378 (scope #1 + #2).

## Context

The cdylib plugin system transfers rich Rust types across the FFI boundary as owned `Box`es — `Box::into_raw` on one side, `Box::from_raw` on the other, in **both** directions (plugin-allocated types freed by the host and vice versa): event envelopes, subscription responses, the vtables, and registration structs. This is only sound because no crate sets `#[global_allocator]`, so every allocation routes through the default **System** allocator (process-global libc `malloc`/`free`).

If any plugin or the host installs a custom allocator (jemalloc, mimalloc, tcmalloc, …), every cross-boundary `Box::from_raw` becomes **silent heap corruption**. The constraint was previously undocumented and unenforced.

Per the issue's recommended path, this PR implements **#1 (document)** and **#2 (enforce at build time)**. #3 (load-time allocator probe) and #4 (structural ABI fix) are deferred.

## #1 — Document the constraint
- `components/plugin-architecture.md` — new *Cross-cdylib Ownership Contract (Shared System Allocator)* subsection under FFI Boundary Design.
- `components/plugin-sdk/README.md` — note under *Compatibility requirements*.
- `components/plugin-sdk/src/lib.rs` — crate-doc warning block (surfaces on docs.rs).
- `components/plugin-sdk/src/ffi/metadata.rs` — module-doc note anchored near version validation.

## #2 — Enforce at build time
- **`deny.toml`** — `cargo-deny` `[bans]` denylist of known custom-allocator crates (`jemallocator`, `tikv-jemallocator`, `mimalloc`, `snmalloc-rs`, `tcmalloc`, `rpmalloc`, `wee_alloc`, `talc`, + their `-sys` backings), each with a reason citing #378. Resolves the full dependency graph so **transitive** allocator deps are caught. Scoped to the `bans` check so it does not overlap the existing `cargo-audit.yml` (advisories/licenses).
- **`.github/workflows/cargo-deny.yml`** — runs `cargo deny check bans` on PRs to `main` + a weekly cron.

## Notes
- A simple grep for `#[global_allocator]` would miss the real risk (a transitive dependency). `cargo-deny [bans]` is the right tool.
- No `Cargo.toml`/dependency changes; docs + a lint gate only.

## Validation
- `cargo deny check bans` → `bans ok`, 0 warnings (config fully parsed, nothing silently ignored).
- Negative test: temporarily banning a present crate failed with exit 2 (gate fires), then reverted to green.
- `cargo check -p drasi-plugin-sdk` compiles cleanly with the new doc comments.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


danielgerlag:
Thanks for the review — addressed all 6 comments in c1641296:

- **"license scanning" claim** (deny.toml, cargo-deny.yml): removed. cargo-audit only runs RustSec advisory checks, so the comments now say "RustSec advisory scanning" only.
- **`[graph]` all-targets** (deny.toml): reworded to make the intent explicit — `targets` is intentionally left unset so cargo-deny resolves dependencies for *every* target triple (default behavior), catching target-gated allocators. No behavior change.
- **"enforced at build time" too strong** (lib.rs, README.md, metadata.rs): softened in all three. The docs now state that *drasi-core CI* enforces this via the cargo-deny rule, and that third-party plugin authors building outside this workspace should avoid a custom global allocator and may add the same check to their own CI.

Re-verified: `cargo deny check bans` → `bans ok`; `cargo check -p drasi-plugin-sdk` clean.

danielgerlag:
Thanks — addressed all three design-review findings in 926eb4b0:

- **Should-fix (direct workspace vector):** added a complementary step to `cargo-deny.yml` that fails if a custom `#[global_allocator]` is declared directly in workspace `*.rs` source — the vector cargo-deny can't see. Comment/doc-comment mentions (lines starting with `//`) are excluded so the prose reference in `lib.rs` doesn't trip the gate; verified locally that a real attribute is caught and the current tree is clean.
- **Should-fix (metadata.rs placement):** removed the allocator note from `metadata.rs`. The struct now documents only constraints it actually validates; the allocator requirement stays in `plugin-architecture.md`, the SDK README, and the `lib.rs` crate-level warning.
- **Nit (action pinning):** pinned `EmbarkStudios/cargo-deny-action` to a full commit SHA (`bb137d7…`, v2.0.20) instead of the floating `@v2`. Happy to migrate to an org-owned shared workflow if/when `drasi-project/.github` gains a cargo-deny reusable workflow.

Re-verified: `cargo deny check bans` → `bans ok`, workflow YAML parses, `cargo check -p drasi-plugin-sdk` clean.

danielgerlag:
Thanks for the correctness/security/testing passes. Addressed the two remaining nits in 5d2223bf (the testing review needed no changes):

- **Security nit (mutable action tag):** `cargo-deny-action` was already pinned to a SHA; also pinned `actions/checkout@v4` to a commit SHA so neither action rides a mutable tag.
- **Correctness nit (local `cargo deny check` confusion):** added a note near `[bans]` in `deny.toml` clarifying the file is run as `cargo deny check bans`, and that a bare `cargo deny check` additionally runs the intentionally-unconfigured advisories/licenses/sources checks.

Re-verified: workflow YAML parses, `cargo deny check bans` → `bans ok`.

agentofreality (approved):

.github/workflows/cargo-deny.yml
```diff
@@ -0,0 +1,48 @@
+name: Cargo Deny
+
+on:
+  pull_request:
+    branches:
+      - main
+      - feature/*
+      - feature-lib
+      - release/*
```
ruokun-niu, line 9: Do we need feature-lib and release/*?

ruokun-niu (approved):

PR #626, opened 2026-07-14
Bring storedproc-mssql reaction into conformance with the Reaction Developer Guide

ruokun-niu:
# Description

Brings the MS SQL Server stored-procedure reaction into full conformance with the Drasi Reaction Plugin Developer Guide (`components/reactions/README.md`, sections 11/14/15), and fixes a data-fidelity/security bug plus a TLS portability gap discovered while exercising the reaction end to end.

**Why:** the reaction previously used a bespoke `@field` / `EXEC proc(@a, @b)` substitution syntax that interpolated row values directly into the SQL text. That is both a SQL-injection / value-corruption risk and a divergence from the guide's shared Handlebars templating contract that every other reaction follows.

**What changed:**

- **Safe parameter binding.** Command templates are now Handlebars templates using the guide's `{{param <path>}}` helper, which binds each row value as a positional SQL parameter (`@P1`, `@P2`, ...) through the tiberius driver instead of interpolating it into the statement. Values (including strings with quotes/semicolons) can no longer alter the SQL. Added a `{{json <path>}}` helper so a whole object can be bound as a JSON string (e.g. `{{param (json after)}}`).
- **Guide-compliant templating.** Re-uses the shared `TemplateSpec` / `QueryConfig` / `TemplateRouting` types from `drasi_lib::reactions::common`, populates all standard context keys (`query_name`, `query_id`, `operation`, `timestamp`, `metadata`, `before`, `after`, `data`), renders in strict mode, and resolves templates per-query -> last dotted segment -> default. Every template is compiled at construction time and route keys are validated against the subscribed query ids, so misconfiguration fails at startup rather than on the first event.
- **Correctness fixes.** Integers are bound consistently as `BIGINT` (no silent 32-bit narrowing); placeholders use tiberius's uppercase `@P{n}` form (the old lowercase `@p{n}` was never exercised end to end).
- **TLS backend features.** The crate depended on `tiberius` with `default-features = false` and no TLS backend, so it could only establish an encrypted connection when built in-workspace alongside another crate that pulled tiberius's TLS backend via Cargo feature unification. A standalone consumer of just this reaction could not connect over TLS to SQL Server. Added `native-tls` (default) and `rustls` Cargo features so the packaged plugin is self-contained.
- Replaced the removed `parser` module with an internal `render` module, added `with_added_command` / `with_updated_command` / `with_deleted_command` builder setters, and refreshed the README and CHANGELOG.

**Testing:** unit tests cover context building, positional binding order, the json helper, strict-mode render failure, and config/route validation. Real-SQL-Server integration tests (testcontainers) drive the full build -> start -> enqueue -> assert -> stop path, including a regression test proving special characters in row values are bound safely. `cargo test`, `cargo fmt --check`, and `cargo clippy --all-targets` all pass. Both TLS backends were verified to compile, and the reaction was exercised end to end against SQL Server 2022 in Docker (ADD/UPDATE/DELETE plus a custom per-query route, with value fidelity confirmed).

**Note for reviewers:** the templating syntax change is **breaking** for anyone using the old `@field` command format, and `MsSqlStoredProcReactionConfig::validate` now takes the subscribed query ids. Both are documented under `[Unreleased]` in the crate CHANGELOG.

## Type of change

Fixes: #issue_number


danielgerlag (approved):
We should add a task to update drasi.io docs

ruokun-niu:
> We should add a task to update drasi.io docs

@danielgerlag I have created an issue https://github.com/drasi-project/docs/issues/253

agentofreality:
@ruokun-niu Did you go through the checklist: https://github.com/drasi-project/drasi-core/blob/main/components/reactions/README.md#14-conformance-checklist

Also, need to get into habit of docs being part of acceptance criteria for this kind of issue. So the docs should be ready and reviewed before we approve this issue.

agentofreality (approved):

PR #628, opened 2026-07-20
Release persistent index-backend handles on DrasiLib::shutdown() (#627)

### LINKED ISSUE #627: DrasiLib::shutdown() should release persistent index-backend handles (RocksDB lock not freed until drop)
_Filed from downstream `drasi-project/drasi-nodejs` (see drasi-nodejs#22). Verified against `drasi-lib` 0.8.9 + `drasi-index-rocksdb` 0.5.8._

### Problem

`DrasiLib` never releases a persistent **index-backend** handle for the life of the `DrasiLib` value, so a persistent backend that takes an exclusive OS resource — notably RocksDB, which holds a **process-exclusive lock** on its data directory — keeps that resource locked until the whole `DrasiLib` is dropped (in practice, process exit). There is no non-destructive way to release it while keeping the on-disk data.

Concretely, with a RocksDB index provider:

```rust
let core = DrasiLib::builder()
    .with_default_index_provider("rocksdb", Arc::new(RocksDbIndexProvider::new(path, false, false)))
    .build().await?;
// ... use ...
core.shutdown().await?;      // stops components, aborts the graph loop
// `core` still holds the open RocksDB DB(s); the LOCK is NOT released.

let core2 = DrasiLib::builder()  // same `path`
    .with_default_index_provider("rocksdb", Arc::new(RocksDbIndexProvider::new(path, false, false)))
    .build().await?;
core2.add_query(...).await?;
// => Err: "Failed to build indexes"
//    IO error: lock hold by current process ... <path>/<queryId>/LOCK: No locks available
```

Downstream this surfaces to `@drasi/lib` users as: after `close()`, a new engine cannot reopen the same rocksdb `indexStore` path in the same process. Cross-*process* restart is fine (the lock releases on process exit).

### Root cause (from the 0.8.9 source)

- `DrasiLib::stop()` and `DrasiLib::shutdown()` (`src/lib_core.rs`) stop components and abort the graph-update loop, but do **not** drop the per-query index sets. The query runtimes (each owning an `OptimisticTransactionDB` via `RocksDbElementIndex`/`ResultIndex`/`FutureQueue`/checkpoint/outbox/live-results) remain owned by `self.query_manager` / the component graph, i.e. by the still-alive `DrasiLib`. RocksDB releases the OS lock only when the DB handle is dropped.
- The only path that drops a persistent index handle is `QueryManager::teardown_query` (`src/queries/manager.rs`), but it is **destructive**: after teardown it clears the persistent indexes + checkpoints "so a future query with the same ID starts fresh". So it can't be used to release the lock while preserving data.
- The provision path already knows the handle must be dropped before re-opening (it sets `checkpoint_store`/`outbox_writer`/`live_results_writer`/`element_index`/… to `None` before re-`build`ing indexes for a backend "like RocksDB that hold[s] an exclusive lock"), which confirms drop-to-release is the intended mechanism — there just isn't a non-destructive engine-level teardown that does it on shutdown.

### Proposed solution

On `shutdown()` (which is already documented as a permanent, non-restartable teardown), **drop the per-query index sets / index-provider-created handles non-destructively** so persistent backends release their OS locks while leaving on-disk data intact. Equivalently, add a method that releases index handles without the destructive `teardown_query` cleanup.

Acceptance:
- After `core.shutdown().await`, a new `DrasiLib` can open the same RocksDB `path` in the **same process** and recover the prior on-disk query state (no wipe).
- Non-persistent/in-memory backends are unaffected.

### Alternatives considered

- Downstream drop of `Arc<DrasiLib>`: not reliable — `shutdown()`/`close()` take `&self`, and internal Arcs (spawned tasks, callback/bridge contexts) can keep the engine alive, so the lock lingers past `close()`.
- Downstream `teardown_query` on close: releases the lock but deletes persisted state — defeats persistence.
- Documentation-only (done downstream): the rocksdb `indexStore` docs note the lock is released on process exit, not `close()`. Acceptable interim, but same-process reopen stays unsupported until this is fixed.

_Downstream tracking: drasi-project/drasi-nodejs#22._

danielgerlag:
# Description

`DrasiLib` never released the persistent index-backend handles a query retains, so a backend that takes an exclusive OS resource -- notably RocksDB, which holds a process-exclusive lock on its data directory -- kept that resource pinned until the whole `DrasiLib` value was dropped (in practice, process exit). After `shutdown()`, a new `DrasiLib` could not reopen the same rocksdb path in the same process, failing with `LOCK: No locks available`. This blocks the downstream `@drasi/lib` binding (drasi-project/drasi-nodejs#22), whose `close()` cannot free the lock on its own.

**Root cause:** RocksDB's `create_indexes()` shares one `OptimisticTransactionDB` across all eight handles, so the OS lock is freed only when the last clone drops. After a normal `stop()`, the processor task and `FutureQueueSource` are reaped, releasing five of them -- but `DrasiQuery` keeps `checkpoint_store`, `outbox_writer`, and `live_results_writer` for its whole lifetime, and neither `stop()` nor `shutdown()` ever cleared them.

**Approach:** Add a non-destructive release that drops only the in-memory `Arc` handles, leaving all on-disk state intact:

- `Query::release_persistent_handles()` -- a defaulted no-op; `DrasiQuery` clears the three retained handles and drops any lingering `FutureQueueSource`.
- `QueryManager::release_all_persistent_handles()` -- invokes it across all query runtimes.
- `DrasiLib::shutdown()` -- calls it after stopping components and aborting the graph loop.

This runs only on `shutdown()` (the documented permanent teardown), so the in-memory `stop()`/`start()` checkpoint-reuse path is unaffected. Because it clears the inner `Option` fields rather than relying on the `DrasiQuery` itself being dropped, it is robust even when other components (e.g. reactions) still hold an `Arc<dyn Query>` clone.

Adds an e2e test proving same-process shutdown -> reopen -> recover with a brand-new provider on the same RocksDB path. The test was verified to fail with `No locks available` when the fix is disabled, and the first engine is deliberately kept alive across the reopen to prove the lock is freed by `shutdown()` itself rather than by dropping `DrasiLib`.

## Type of change

Fixes: #627

### DIFF
--- a/lib/src/queries/manager.rs
+++ b/lib/src/queries/manager.rs
@@ -2555,7 +2555,10 @@ impl Query for DrasiQuery {
         // Defensive: if the query was never stopped (e.g. left in a terminal Error
         // state), the FutureQueueSource may still hold a clone of the backend future
         // queue. `stop()` normally takes it already, in which case this is a no-op.
-        if let Some(fq) = self.future_queue_source.write().await.take() {
+        // Take the value out first so the write-lock guard is dropped before the
+        // `.await` below (never hold a lock across an await point).
+        let future_queue_source = self.future_queue_source.write().await.take();
+        if let Some(fq) = future_queue_source {
             fq.stop().await;
         }
 
@@ -2943,8 +2946,12 @@ impl QueryManager {
             .collect();
 
         for id in query_ids {
-            if let Ok(query) = self.get_query_instance(&id).await {
-                query.release_persistent_handles().await;
+            match self.get_query_instance(&id).await {
+                Ok(query) => query.release_persistent_handles().await,
+                Err(e) => warn!(
+                    "Failed to release persistent index handles for query '{id}' during \
+                     shutdown: {e}. A persistent backend may keep its exclusive lock held."
+                ),
             }
         }
     }
--- a/lib/src/queries/manager.rs
+++ b/lib/src/queries/manager.rs
@@ -242,6 +242,19 @@ pub trait Query: Send + Sync {
     fn output_metrics(&self) -> Option<Arc<QueryOutputMetrics>> {
         None
     }
+
+    /// Release the persistent index-backend handles this query retains, **without**
+    /// deleting any on-disk data.
+    ///
+    /// Persistent backends (e.g. RocksDB, which holds a process-exclusive lock on its
+    /// data directory) keep that resource pinned until every clone of their shared
+    /// handle is dropped. A query retains some of those handles for its whole lifetime,
+    /// so they are only freed when the whole `DrasiLib` is dropped. This method drops
+    /// those in-memory handles so the backend can release its lock during a permanent
+    /// shutdown, while leaving persisted state intact for a future reopen.
+    ///
+    /// Default: no-op — volatile/in-memory queries hold no such handles.
+    async fn release_persistent_handles(&self) {}
 }
 
 /// Bootstrap phase tracking for each source
@@ -2530,6 +2543,26 @@ impl Query for DrasiQuery {
     fn output_metrics(&self) -> Option<Arc<QueryOutputMetrics>> {
         Some(self.output_metrics.clone())
     }
+
+    async fn release_persistent_handles(&self) {
+        // Drop the persistent handles created by the index backend. For a shared
+        // backend like RocksDB (one `OptimisticTransactionDB` cloned into every
+        // index/store/writer), these are the only backend clones the query still
+        // retains after `stop()`; dropping them lets the backend release its
+        // exclusive lock. On-disk data is left untouched, so a future reopen of the
+        // same path recovers the prior state.
+
+        // Defensive: if the query was never stopped (e.g. left in a terminal Error
+        // state), the FutureQueueSource may still hold a clone of the backend future
+        // queue. `stop()` normally takes it already, in which case this is a no-op.
+        if let Some(fq) = self.future_queue_source.write().await.take() {
+            fq.stop().await;
+        }
+
+        *self.checkpoint_store.write().await = None;
+        *self.outbox_writer.write().await = None;
+        *self.live_results_writer.write().await = None;
+    }
 }
 
 pub struct QueryManager {
@@ -2887,6 +2920,33 @@ impl QueryManager {
         }
 
         Ok(())
+    }
+
+    /// Release persistent index-backend handles for every query runtime, without
+    /// deleting any on-disk data.
+    ///
+    /// Intended for permanent shutdown ([`crate::DrasiLib::shutdown`]). Persistent
+    /// backends such as RocksDB hold a process-exclusive lock on their data directory
+    /// until every clone of their shared handle is dropped; queries retain some of
+    /// those handles for their whole lifetime, so without this they are only freed
+    /// when the entire `DrasiLib` is dropped. Clearing them here lets the backend
+    /// release its lock while leaving persisted state intact for a future reopen.
+    ///
+    /// Callers should stop components first so the transient handles held by
+    /// per-query tasks are already dropped by the time this runs.
+    pub async fn release_all_persistent_handles(&self) {
+        let query_ids: Vec<String> = self
+            .list_queries()
+            .await
+            .into_iter()
+            .map(|(id, _status)| id)
+            .collect();
+
+        for id in query_ids {
+            if let Ok(query) = self.get_query_instance(&id).await {
+                query.release_persistent_handles().await;
+            }
+        }
     }
 
     /// List all registered queries with their current lifecycle status.
--- a/lib/tests/static_rocksdb_index_e2e.rs
+++ b/lib/tests/static_rocksdb_index_e2e.rs
@@ -242,7 +242,116 @@ async fn static_rocksdb_index_persists_across_restart() -> Result<()> {
     Ok(())
 }
 
-/// Proves the `with_default_index_provider` path that `drasi-server` uses for its
+/// Proves the fix for issue #627: after `DrasiLib::shutdown()`, the RocksDB
+/// process-exclusive lock on the data directory is released, so a **brand-new**
+/// `DrasiLib` + provider can reopen the **same** on-disk path **within the same
+/// process** and recover the prior query state — without re-ingesting any source
+/// events.
+///
+/// This is exactly the scenario the sibling `..._persists_across_restart` test
+/// documents as previously unsupported: before the fix, the query runtime kept the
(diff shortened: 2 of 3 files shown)


lib/src/queries/manager.rs
```diff
@@ -2530,6 +2543,26 @@ impl Query for DrasiQuery {
...
+        // retains after `stop()`; dropping them lets the backend release its
+        // exclusive lock. On-disk data is left untouched, so a future reopen of the
+        // same path recovers the prior state.
+
+        // Defensive: if the query was never stopped (e.g. left in a terminal Error
+        // state), the FutureQueueSource may still hold a clone of the backend future
+        // queue. `stop()` normally takes it already, in which case this is a no-op.
+        if let Some(fq) = self.future_queue_source.write().await.take() {
+            fq.stop().await;
+        }
```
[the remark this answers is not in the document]
danielgerlag: Good catch. Fixed in 94c73f4: I now take the value out of the `RwLock` before awaiting, so the write guard is dropped before `fq.stop().await`.

lib/src/queries/manager.rs
```diff
@@ -2889,6 +2922,33 @@ impl QueryManager {
...
+            .await
+            .into_iter()
+            .map(|(id, _status)| id)
+            .collect();
+
+        for id in query_ids {
+            if let Ok(query) = self.get_query_instance(&id).await {
+                query.release_persistent_handles().await;
+            }
+        }
```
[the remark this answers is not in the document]
danielgerlag: Agreed. Fixed in 94c73f4: it now logs a warning (naming the query and error) when a runtime can't be retrieved, so a failure to release a persistent backend lock is diagnosable rather than silent.

agentofreality (approved):

ruokun-niu (approved):

lib/tests/static_rocksdb_index_e2e.rs
```diff
@@ -242,7 +242,116 @@ async fn static_rocksdb_index_persists_across_restart() -> Result<()> {
...
+            .build()
+            .await?,
+    );
+
+    // If the lock were still held, `start()` would fail while building the indexes
+    // ("Failed to build indexes" / "No locks available").
+    core2
+        .start()
+        .await
+        .expect("second engine should open the same RocksDB path after shutdown released the lock");
```
amansinghoriginal, line 331: This expect() can't fail. DrasiLib::start() doesn't propagate query start failures; it logs "Some queries failed to start (they are now in Error state)" and returns Ok(()).

I reverted the two src files to the base commit and kept this test: start() passed, and the test failed 10s later at line 339 with `got: []`. So it does catch the regression, just via a generic "no rows" message that an unrelated persistence bug would produce too.

Maybe assert the query actually came up?

```rust
core2.start().await?;
assert_eq!(
    core2.get_query_status("people").await?,
    ComponentStatus::Running,
    "a held RocksDB lock leaves the query in Error",
);
```

lib/src/queries/manager.rs
```diff
@@ -2530,6 +2543,29 @@ impl Query for DrasiQuery {
...
+
+        // Defensive: if the query was never stopped (e.g. left in a terminal Error
+        // state), the FutureQueueSource may still hold a clone of the backend future
+        // queue. `stop()` normally takes it already, in which case this is a no-op.
+        // Take the value out first so the write-lock guard is dropped before the
+        // `.await` below (never hold a lock across an await point).
+        let future_queue_source = self.future_queue_source.write().await.take();
+        if let Some(fq) = future_queue_source {
+            fq.stop().await;
+        }
```
amansinghoriginal, line 2563: I don't think this defensive case holds. stop_all_components (lifecycle.rs:200) skips any component not in Running|Starting, so an Error-state query never gets stop_query(), its shutdown_tx never fires, and the processor task stays parked in select! still holding continuous_query_for_processor (element, archive, result, future_queue, session_control) plus its own Arc<FutureQueueSource> clone.

So taking the field here doesn't drop the FutureQueueSource, and the lock stays held anyway. Bootstrap failure reaches this state: manager.rs:1901 sets Error after the processor spawns.

Worth aborting the task here too, or dropping the branch and stating the precondition?

lib/src/queries/manager.rs
```diff
@@ -2889,6 +2925,37 @@ impl QueryManager {
...
+    /// Intended for permanent shutdown ([`crate::DrasiLib::shutdown`]). Persistent
+    /// backends such as RocksDB hold a process-exclusive lock on their data directory
+    /// until every clone of their shared handle is dropped; queries retain some of
+    /// those handles for their whole lifetime, so without this they are only freed
+    /// when the entire `DrasiLib` is dropped. Clearing them here lets the backend
+    /// release its lock while leaving persisted state intact for a future reopen.
+    ///
+    /// Callers should stop components first so the transient handles held by
+    /// per-query tasks are already dropped by the time this runs.
+    pub async fn release_all_persistent_handles(&self) {
```
amansinghoriginal, line 2940: lib.rs:80 is `pub mod queries` and query_manager() is pub, so embedders can reach this on a running engine. That nulls checkpoint_store/outbox_writer/live_results_writer under a live query, and fetch_snapshot()'s persisted fallback (line 2470) then quietly returns empty.

shutdown() is in-crate, so would pub(crate) work?

lib/tests/static_rocksdb_index_e2e.rs
```diff
@@ -242,7 +242,116 @@ async fn static_rocksdb_index_persists_across_restart() -> Result<()> {
...
+    assert!(names.contains(&"Bob"), "missing Bob in {recovered:?}");
+    assert!(names.contains(&"Carol"), "missing Carol in {recovered:?}");
+
+    core2.shutdown().await?;
+
+    // Keep the first engine alive until the very end so the test proves the lock was
+    // freed by `shutdown()` rather than by dropping `core1`.
+    drop(core1);
+    Ok(())
+}
```
amansinghoriginal, line 354: Looks like the insert ate the first line of the next test's doc comment, so it now starts mid-sentence. The old line was:

```rust
/// Proves the `with_default_index_provider` path that `drasi-server` uses for its
```

Also missing a blank line after this `}`.

amansinghoriginal (commented):
Traced the mechanism and it matches the issue: one Arc<OptimisticTransactionDB> cloned into all eight handles, and the three that DrasiQuery retains outlive stop(). Scoping the release to shutdown() looks right, since start() already nulls those three before rebuilding.

Ran the suite both ways. All 4 tests pass with the fix, and the new one does fail without it. cargo fmt is clean.

A few comments inline, mostly small. The test one is the only thing I'd want looked at before merge; the rest are take or leave.

PR #653, opened 2026-07-28
Prune emptied set_id from InMemoryResultIndex.sorted_sets

### LINKED ISSUE #643: Memory leak: InMemoryResultIndex.sorted_sets retains empty BTreeMap per set_id (unbounded growth)
### Steps to reproduce

Run a continuous query that maintains a sorted set / ordered aggregation (e.g. `min`/`max`/percentile-style ops backed by `LazySortedSetStore`) over the default `InMemoryResultIndex` (`core/src/in_memory_index/in_memory_result_index.rs`), grouped by a high-cardinality key. Drive counts up and back down to zero for many distinct `set_id`s. Observe `sorted_sets` growing without bound.

Root cause — `increment_value_count()` removes a value when its count reaches 0 but never removes the now-empty inner `BTreeMap`:

```rust
// L159-173
let set = data.entry(set_id).or_insert(BTreeMap::new()); // creates outer entry
let entry = set.entry(value).and_modify(|v| *v += delta).or_insert(delta);
if entry.is_negative() { return Err(IndexError::CorruptedData); }
if *entry == 0 {
    set.remove(&value);   // inner BTreeMap may now be empty, but the set_id entry stays
}
```

`sorted_sets: HashMap<u64, BTreeMap<OrderedFloat<f64>, isize>>`. Emptied `set_id` entries are retained forever; only `clear()` (L96–97) reclaims them. Growth is proportional to the number of distinct `set_id`s ever seen.

### Observed behavior

`sorted_sets` accumulates empty `BTreeMap`s keyed by `set_id` after all their values drop to count 0.

### Desired behavior

After `set.remove(&value)`, remove the `set_id` entry from `sorted_sets` when its inner `BTreeMap` becomes empty.

### Workaround

None short of periodically clearing query state.

### Drasi Version

N/A — identified via source code review of `main` @ `72e88a74`.

### Operating system

N/A — platform-independent source-level defect.

### Additional context

Part of the "emptied containers never pruned" pattern in the in-memory indexes (see sibling issues for `InMemoryElementIndex`). File: `core/src/in_memory_index/in_memory_result_index.rs` (`increment_value_count` L149–176).

- [ ] Yes, I would like to support you

danielgerlag:
# Description

`InMemoryResultIndex.sorted_sets` grew without bound over time. `increment_value_count()` removed a value once its count reached 0 but left the now-empty inner `BTreeMap` (keyed by `set_id`) in place. For continuous queries backed by `LazySortedSetStore` (min/max) grouped by a high-cardinality key, every distinct `set_id` ever seen was retained forever, so memory usage tracked the total number of groups rather than the number of currently-active ones. Previously only `clear()` reclaimed those entries.

The fix removes the `set_id` entry from `sorted_sets` once its inner `BTreeMap` becomes empty:

```rust
if set.is_empty() {
    data.remove(&set_id);
}
```

Added three tests: an emptied `set_id` is pruned (with `get_value_count` / `get_next` still behaving), a `set_id` with remaining values is retained, and draining one `set_id` leaves other active ones intact. `cargo test`, `clippy`, and `fmt` all pass.

## Type of change

Fixes: #643


core/src/in_memory_index/in_memory_result_index.rs
```diff
@@ -196,3 +200,55 @@ impl ResultSequenceCounter for InMemoryResultIndex {
...
+        let data = index.sorted_sets.read().await;
+        assert!(
+            !data.contains_key(&set_id),
+            "emptied set_id should be pruned from sorted_sets"
+        );
+        assert!(data.is_empty());
+    }
+
+    #[tokio::test]
+    async fn increment_value_count_retains_nonempty_set_id() {
```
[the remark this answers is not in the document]
danielgerlag: Good call. Added `increment_value_count_prunes_only_the_drained_set_id`, which sets up two active set_ids, drains one, and asserts only that entry is pruned while the other survives with its count (plus `data.len() == 1`). This guards against a future regression that wipes all entries on the first prune.

agentofreality (approved):

core/src/in_memory_index/in_memory_result_index.rs
```diff
@@ -196,3 +200,94 @@ impl ResultSequenceCounter for InMemoryResultIndex {
         Ok(data.clone())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ordered_float::OrderedFloat;
```
amansinghoriginal, line 207:
```suggestion
    use super::*;
```
Redundant: `OrderedFloat` is already in scope from `use super::*;` (parent
imports it at line 27). Clippy stays quiet because an explicit import shadows
a glob. Non-blocking, could fold into the rebase.

core/src/in_memory_index/in_memory_result_index.rs
```diff
@@ -196,3 +200,94 @@ impl ResultSequenceCounter for InMemoryResultIndex {
...
+        let data = index.sorted_sets.read().await;
+        assert!(
+            !data.contains_key(&set_id),
+            "emptied set_id should be pruned from sorted_sets"
+        );
+        assert!(data.is_empty());
+    }
+
+    #[tokio::test]
+    async fn increment_value_count_retains_nonempty_set_id() {
```
amansinghoriginal, line 236: This one passes without the fix too, so it is an over-pruning guard rather
than a regression test. Worth keeping, just not coverage for #643.

amansinghoriginal (approved):
LGTM

Verified by deleting the fix and rerunning: both pruning tests fail, so they pin the bug. Also ran min() over 10k churned groups end to end, retained set_ids go from 10,000 to 0.

PR #665, opened 2026-07-30
Fix #664: restore source sequence counter on resume so restart doesn't drop the first K changes

### LINKED ISSUE #664: Restart with a persistent index silently drops the first K changes when the source has no durable position
## Summary

When a query has a persistent index store, the first *K* source changes pushed after a restart are silently discarded, where *K* is the number of changes dispatched before the restart. No error is raised and `push_change` returns successfully.

This happens for any source that does not report a durable `source_position`. The query restores its dedup high-water mark from the checkpoint unconditionally, but only asks the source to resume when a `source_position` exists, and the source's sequence counter always restarts at 1.

This is distinct from #640. That issue is about concurrent dispatch reordering within one process; this one reproduces with entirely sequential, single-threaded pushes and is about state carried *across* a restart.

## Reproduction

Two processes against the same store paths (a single process cannot do this, because redb holds its lock for the process lifetime).

Process 1 — write and exit cleanly:

```python
engine = await Drasi.create(
    "recovery-demo",
    state_store={"kind": "redb", "path": ".../state"},
    index_store={"kind": "rocksdb", "path": ".../index"},
)
await engine.start()
await engine.add_python_source("orders")
await engine.add_query("open-orders", QUERY, ["orders"])
for i in range(5):                      # K = 5
    await engine.push_change("orders", insert(f"o{i}"))
assert len(await engine.get_query_results("open-orders")) == 5
await engine.close()
```

Process 2 — same paths, push 15 *new* rows `o100`..`o114`:

```
[probe]   recovered 5 rows
[probe]   pushed 15 new rows, 10 landed: ['o105', ..., 'o114']
```

`o100`..`o104` — exactly the first 5 — are gone. Re-running with K=3 drops exactly `o100`..`o102`. The number of discarded post-restart writes always equals the number of pre-restart writes.

Recovery of the existing 5 rows itself works correctly, including after `SIGABRT`. Deleting the rocksdb index yields 0 recovered rows, confirming the rows genuinely come from disk.

## Cause

With `RUST_LOG=drasi_lib=debug`, the resuming query subscribes as:

```
Query 'open-orders' subscribing to python source 'orders'
  (bootstrap: true, resume_from: None, request_handle: true)
```

In `src/queries/manager.rs`:

```rust
if let Some(cp) = checkpoints.get(&settings.source_id) {
    checkpoint_sequences_per_source
        .insert(settings.source_id.clone(), cp.sequence);   // 1106 — always
    settings.request_position_handle = true;
    if let Some(pos) = &cp.source_position {
        settings.resume_from = Some(pos.clone());           // 1110 — conditional
    }
}
```

- `manager.rs:1106` records `cp.sequence` (5) **unconditionally**.
- `manager.rs:1110` sets `resume_from` **only if `cp.source_position` is `Some`**.
- `manager.rs:2147` seeds `SequenceDedup::new(checkpoint_sequences_per_source)` with that 5.
- `manager.rs:2195` then drops everything with `sequence <= 5` via `should_skip`.
- `sources/base.rs:339` constructs `next_sequence: AtomicU64::new(1)` — the counter restarts at 1 in every process.

So the dedup watermark survives the restart but the counter it is compared against does not.

`SourceBase::set_next_sequence` (`sources/base.rs:580`) exists for exactly this, and its doc says "Reset the sequence counter, typically after recovering from a checkpoint." But it has **no production callers** in drasi-lib 0.8.9 — the only reference is a doc comment in `sources/traits.rs:248`, which makes recovery the source implementation's responsibility:

> Implementations that maintain a sequence counter must recover its monotonicity after a restart from their durable position (e.g. the WAL head via `SourceBase::set_next_sequence`, or the native `resume_from` position).

A source with no durable position of its own cannot satisfy that contract, because it is never handed the checkpoint sequence the query is about to dedup against.

## Suggested fix

The checkpoint sequence is already known at subscribe time, so the gap can be closed without a durable position. Options:

1. Pass `cp.sequence` to the source on subscribe (independently of `source_position`) so `SourceBase` can call `set_next_sequence` itself. This fixes every source at once and needs no plugin changes.
2. Alternatively, only seed `SequenceDedup` for sources that were actually asked to resume — i.e. move the `manager.rs:1106` insert inside the `source_position.is_some()` branch. This trades the silent drop for possible duplicate delivery, so option 1 seems preferable.

Either way, a source whose sequence cannot be restored should not have its first *K* events silently discarded.

## Environment

- `drasi-lib` 0.8.9, `drasi-core` 0.5.7, via the Python bindings (`drasi-lib` 0.1.3 on PyPI)
- macOS arm64, rocksdb index store + redb state store
- Reproduced deterministically at K=3 and K=5, after both clean shutdown and `SIGABRT`

danielgerlag:
## Summary

Fixes #664.

When a query has a persistent index/checkpoint store, the first *K* source changes pushed after a **process restart** were silently discarded, where *K* is the checkpoint sequence reached before the restart. This affected any source that does not report a durable `source_position`.

### Root cause

On resume, the query records `cp.sequence` into its dedup high-water mark **unconditionally** and drops events with `sequence <= K`. But `resume_from` is only set when `cp.source_position` is `Some`, and `SourceBase.next_sequence` is constructed as `AtomicU64::new(1)` — the counter restarts at 1 in every process. So the dedup mark (K) survives the restart, but the counter it is compared against does not: post-restart events get seq `1..=K` and are all silently deduplicated away.

### Fix (issue's recommended Option 1)

Thread the checkpoint sequence to the source on subscribe (independent of `source_position`) so `SourceBase` advances its own counter above the dedup mark:

- Add `resume_sequence: Option<u64>` to `SourceSubscriptionSettings`.
- Populate it from `cp.sequence` **unconditionally** on resume in the query manager (and clear it on the auto-reset retry path).
- Add `SourceBase::ensure_next_sequence_above` (uses `fetch_max`, so it never lowers the counter — safe with multiple subscribers and already-advanced counters) and call it in `subscribe_with_bootstrap_context` when `resume_sequence` is set.

This fixes every source that delegates to the base helper with **no plugin-author changes**, and keeps dedup seeding unconditional (still required so durable/WAL sources filter genuinely-replayed events). The WAL replay path is intentionally left untouched — those sources restore their own counter.

### FFI-loaded plugin sources

Sources loaded as dynamic cdylibs cross the FFI subscribe boundary. This also carries `resume_sequence` so they get the same counter bump — **including durable DB sources** (e.g. postgres): they resume to the correct LSN via `resume_from`, but their `next_sequence` still restarts at 1 in a new process, so without this they'd lose the first K post-restart changes too (thanks @amansinghoriginal for catching this).

- `SourceVtable::subscribe_fn` gains `resume_sequence` + `resume_sequence_present` params (mirroring how `resume_from` is passed as ptr+len).
- The host proxy encodes `settings.resume_sequence`; both plugin-side reconstruction sites decode it back so the plugin's `subscribe_with_bootstrap` runs `ensure_next_sequence_above`.
- The `(value, present)` encoding of the `Option<u64>` lives in one shared, unit-tested helper in `ffi-primitives` (`encode_optional_u64` / `decode_optional_u64`) so host and plugin can't drift.

### Notes

- Adding a field to the public `SourceSubscriptionSettings` is a breaking literal change; all construction sites across the workspace are updated (mechanical `resume_sequence: None`).

## Tests

- **Unit** (`sources/base.rs`): `resume_sequence: Some(K)` makes the next dispatched event seq `K+1`; without it the first event is seq `1`; `fetch_max` never lowers across subscribers.
- **E2E regression** (`queries/checkpoint_tests.rs`): persistent backend + a source with no durable position; push K events, simulate a process restart (reset the source counter), push K new events, assert all are processed (checkpoint advances to `2K`). Verified to **fail before** the fix (checkpoint stuck at K) and **pass after**.
- **FFI marshalling**: round-trip unit tests for `encode_optional_u64`/`decode_optional_u64` in `ffi-primitives`, plus a cdylib integration test (`test_ffi_resume_sequence_marshalled`) exercising the full host→plugin subscribe path at runtime.

All 879 `drasi-lib` lib tests pass; `cargo clippy --workspace --all-targets` and `cargo fmt --check` are clean; the full workspace builds with `--all-targets`.

### DIFF
--- a/components/host-sdk/src/proxies/source.rs
+++ b/components/host-sdk/src/proxies/source.rs
@@ -212,24 +212,13 @@ impl Source for SourceProxy {
             None => (std::ptr::null(), 0u32),
         };
 
-        // The FFI subscribe ABI does not yet marshal `resume_sequence`, so a
-        // dynamically-loaded plugin source cannot advance its in-memory sequence
-        // counter above the query's dedup high-water mark on resume. When the
-        // source also has no durable `resume_from` position of its own, the first
-        // `resume_sequence` post-restart events would be stamped 1..=seq and
-        // silently deduplicated away (issue #664). Surface the gap in logs until
-        // the ABI is extended to carry the checkpoint sequence.
-        if let Some(seq) = settings.resume_sequence {
-            if settings.resume_from.is_none() {
-                log::warn!(
-                    "SourceProxy::subscribe('{}') for query '{}': FFI ABI does not marshal \
-                     resume_sequence ({seq}); this dynamically-loaded source may silently drop \
-                     the first {seq} post-restart events. Pending an ABI extension.",
-                    settings.source_id,
-                    settings.query_id,
-                );
-            }
-        }
+        // Marshal the optional checkpoint sequence as (value, present) since the
+        // ABI cannot carry `Option<u64>` directly. This lets a resuming plugin
+        // source advance its in-memory sequence counter above the query's dedup
+        // high-water mark, so the first K post-restart events are not silently
+        // deduplicated away — closing issue #664 for FFI-loaded sources too.
+        let (resume_sequence, resume_sequence_present) =
+            drasi_ffi_primitives::encode_optional_u64(settings.resume_sequence);
 
         let resp_ptr = (self.vtable.subscribe_fn)(
             self.vtable.state,
@@ -241,6 +230,8 @@ impl Source for SourceProxy {
             resume_from_ptr,
             resume_from_len,
             settings.request_position_handle,
+            resume_sequence,
+            resume_sequence_present,
         );
 
         if resp_ptr.is_null() {
--- a/components/host-sdk/src/proxies/source.rs
+++ b/components/host-sdk/src/proxies/source.rs
@@ -212,6 +212,25 @@ impl Source for SourceProxy {
             None => (std::ptr::null(), 0u32),
         };
 
+        // The FFI subscribe ABI does not yet marshal `resume_sequence`, so a
+        // dynamically-loaded plugin source cannot advance its in-memory sequence
+        // counter above the query's dedup high-water mark on resume. When the
+        // source also has no durable `resume_from` position of its own, the first
+        // `resume_sequence` post-restart events would be stamped 1..=seq and
+        // silently deduplicated away (issue #664). Surface the gap in logs until
+        // the ABI is extended to carry the checkpoint sequence.
+        if let Some(seq) = settings.resume_sequence {
+            if settings.resume_from.is_none() {
+                log::warn!(
+                    "SourceProxy::subscribe('{}') for query '{}': FFI ABI does not marshal \
+                     resume_sequence ({seq}); this dynamically-loaded source may silently drop \
+                     the first {seq} post-restart events. Pending an ABI extension.",
+                    settings.source_id,
+                    settings.query_id,
+                );
+            }
+        }
+
         let resp_ptr = (self.vtable.subscribe_fn)(
             self.vtable.state,
             source_id_ffi,
--- a/components/plugin-sdk/src/ffi/vtable_gen.rs
+++ b/components/plugin-sdk/src/ffi/vtable_gen.rs
@@ -748,6 +748,8 @@ pub fn build_source_vtable<T: Source + 'static>(
         resume_from_ptr: *const u8,
         resume_from_len: u32,
         request_position_handle: bool,
+        resume_sequence: u64,
+        resume_sequence_present: bool,
     ) -> *mut FfiSubscriptionResponse {
         let w = unsafe { &*(state as *const SourceWrapper<T>) };
         let source_id_str = unsafe { source_id.to_string() };
@@ -780,12 +782,14 @@ pub fn build_source_vtable<T: Source + 'static>(
             relations,
             resume_from,
             request_position_handle,
-            // The FFI vtable does not (yet) marshal the checkpoint sequence, so
-            // dynamically-loaded plugins default to None here. In-process sources
-            // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
-            // via the native path. The host proxy logs a warning when it has to
-            // drop a `resume_sequence` here (see host-sdk SourceProxy::subscribe).
-            resume_sequence: None,
+            // Reconstruct the optional checkpoint sequence from its (value,
+            // present) marshalling so a resuming source delegating to
+            // `SourceBase::subscribe_with_bootstrap` advances its counter above
+            // the query's dedup high-water mark (issue #664).
+            resume_sequence: drasi_ffi_primitives::decode_optional_u64(
+                resume_sequence,
+                resume_sequence_present,
+            ),
         };
 
         let handle = (w.runtime_handle)().handle().clone();
@@ -1144,6 +1148,8 @@ pub fn build_source_vtable_from_boxed(
         resume_from_ptr: *const u8,
         resume_from_len: u32,
         request_position_handle: bool,
+        resume_sequence: u64,
+        resume_sequence_present: bool,
     ) -> *mut FfiSubscriptionResponse {
         let w = unsafe { &*(state as *const DynSourceWrapper) };
         let source_id_str = unsafe { source_id.to_string() };
@@ -1176,12 +1182,14 @@ pub fn build_source_vtable_from_boxed(
             relations,
             resume_from,
             request_position_handle,
-            // The FFI vtable does not (yet) marshal the checkpoint sequence, so
-            // dynamically-loaded plugins default to None here. In-process sources
-            // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
-            // via the native path. The host proxy logs a warning when it has to
-            // drop a `resume_sequence` here (see host-sdk SourceProxy::subscribe).
-            resume_sequence: None,
+            // Reconstruct the optional checkpoint sequence from its (value,
+            // present) marshalling so a resuming source delegating to
+            // `SourceBase::subscribe_with_bootstrap` advances its counter above
+            // the query's dedup high-water mark (issue #664).
+            resume_sequence: drasi_ffi_primitives::decode_optional_u64(
+                resume_sequence,
+                resume_sequence_present,
+            ),
         };
 
         let handle = (w.runtime_handle)().handle().clone();
--- a/components/plugin-sdk/src/ffi/vtable_gen.rs
+++ b/components/plugin-sdk/src/ffi/vtable_gen.rs
@@ -780,6 +780,11 @@ pub fn build_source_vtable<T: Source + 'static>(
             relations,
             resume_from,
             request_position_handle,
+            // The FFI vtable does not (yet) marshal the checkpoint sequence, so
+            // dynamically-loaded plugins default to None here. In-process sources
+            // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
+            // via the native path.
+            resume_sequence: None,
         };
 
         let handle = (w.runtime_handle)().handle().clone();
@@ -1170,6 +1175,11 @@ pub fn build_source_vtable_from_boxed(
             relations,
             resume_from,
             request_position_handle,
+            // The FFI vtable does not (yet) marshal the checkpoint sequence, so
+            // dynamically-loaded plugins default to None here. In-process sources
+            // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
+            // via the native path.
+            resume_sequence: None,
         };
 
         let handle = (w.runtime_handle)().handle().clone();
--- a/components/plugin-sdk/src/ffi/vtable_gen.rs
+++ b/components/plugin-sdk/src/ffi/vtable_gen.rs
@@ -783,7 +783,8 @@ pub fn build_source_vtable<T: Source + 'static>(
             // The FFI vtable does not (yet) marshal the checkpoint sequence, so
             // dynamically-loaded plugins default to None here. In-process sources
             // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
-            // via the native path.
+            // via the native path. The host proxy logs a warning when it has to
+            // drop a `resume_sequence` here (see host-sdk SourceProxy::subscribe).
             resume_sequence: None,
         };
 
@@ -1178,7 +1179,8 @@ pub fn build_source_vtable_from_boxed(
             // The FFI vtable does not (yet) marshal the checkpoint sequence, so
             // dynamically-loaded plugins default to None here. In-process sources
             // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
-            // via the native path.
+            // via the native path. The host proxy logs a warning when it has to
+            // drop a `resume_sequence` here (see host-sdk SourceProxy::subscribe).
             resume_sequence: None,
         };
 
--- a/components/plugin-sdk/src/ffi/vtables.rs
+++ b/components/plugin-sdk/src/ffi/vtables.rs
@@ -339,9 +339,14 @@ drasi_ffi_primitives::ffi_vtable! {
         fn initialize_fn(state: *mut, ctx: *const FfiRuntimeContext),
 
         // Subscriptions
-        /// Subscribe with query_id, node_labels JSON, relation_labels JSON, and
-        /// optional resume_from position bytes.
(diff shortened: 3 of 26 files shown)


components/plugin-sdk/src/ffi/vtable_gen.rs
```diff
@@ -780,6 +780,11 @@ pub fn build_source_vtable<T: Source + 'static>(
             relations,
             resume_from,
             request_position_handle,
+            // The FFI vtable does not (yet) marshal the checkpoint sequence, so
+            // dynamically-loaded plugins default to None here. In-process sources
+            // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
+            // via the native path.
+            resume_sequence: None,
```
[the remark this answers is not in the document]
danielgerlag: Good catch — addressed in 21b741de. I put the `warn!` on the host side (`SourceProxy::subscribe` in host-sdk) rather than the plugin-side reconstruction here, because by the time we reach this vtable code the sequence was never marshalled across the boundary, so `resume_sequence` is already gone. The host proxy is the point where the value is still known and being dropped. It logs only when `resume_sequence.is_some() && resume_from.is_none()` — i.e. the exact non-durable case that reproduces #664 — so a durable FFI source resuming via `resume_from` stays quiet. Comments here now cross-reference that site.

lib/src/config/schema.rs
```diff
@@ -176,6 +177,15 @@ pub struct SourceSubscriptionSettings {
...
     pub request_position_handle: bool,
+    /// If set, the resuming query has a dedup high-water mark at this sequence and
+    /// will discard any event with `sequence <= resume_sequence`. Sources that
+    /// restart their monotonic sequence counter from scratch in a new process
+    /// (e.g. those with no durable `source_position` of their own) must advance
+    /// their counter above this value so post-restart events are not silently
+    /// deduplicated away. `SourceBase::subscribe_with_bootstrap_context` honours
+    /// this automatically; unlike `resume_from`, it is populated whenever a
+    /// checkpoint exists, independent of whether a durable position was stored.
+    pub resume_sequence: Option<u64>,
```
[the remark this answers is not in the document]
danielgerlag: Agreed this struct's churn is a real cost, but we're going to leave `#[non_exhaustive]` out of this PR and not pursue it as a follow-up right now. Two reasons: it's a breaking API change that's out of scope for a targeted recovery bugfix, and the suggested `..Default::default()` escape hatch doesn't actually compile — `#[non_exhaustive]` forbids struct-literal *and* functional-update construction from other crates, so every external construction site (the FFI reconstruction plus ~20 test literals across the source crates) would need to move to a real builder/constructor. That's a deliberate API-design decision better made on its own rather than bundled here. Keeping the mechanical `resume_sequence: None` additions for now.

components/host-sdk/src/proxies/source.rs
```diff
@@ -212,6 +212,25 @@ impl Source for SourceProxy {
...
 
+        // The FFI subscribe ABI does not yet marshal `resume_sequence`, so a
+        // dynamically-loaded plugin source cannot advance its in-memory sequence
+        // counter above the query's dedup high-water mark on resume. When the
+        // source also has no durable `resume_from` position of its own, the first
+        // `resume_sequence` post-restart events would be stamped 1..=seq and
+        // silently deduplicated away (issue #664). Surface the gap in logs until
+        // the ABI is extended to carry the checkpoint sequence.
+        if let Some(seq) = settings.resume_sequence {
+            if settings.resume_from.is_none() {
```
amansinghoriginal: Only the WAL-backed sources restore their own counter (http:1130, grpc:327, application:581). A cdylib postgres/kafka/mysql/mssql source resumes with `resume_from = Some(lsn)` but still starts at 1, loses its first `seq` events, and never trips this. Drop the `resume_from.is_none()` check so it covers them? The comment at 218 assumes the same thing.

components/plugin-sdk/src/ffi/vtable_gen.rs
```diff
@@ -780,6 +780,12 @@ pub fn build_source_vtable<T: Source + 'static>(
             relations,
             resume_from,
             request_position_handle,
+            // The FFI vtable does not (yet) marshal the checkpoint sequence, so
+            // dynamically-loaded plugins default to None here. In-process sources
+            // that delegate to `SourceBase::subscribe_with_bootstrap` receive it
+            // via the native path. The host proxy logs a warning when it has to
+            // drop a `resume_sequence` here (see host-sdk SourceProxy::subscribe).
+            resume_sequence: None,
```
amansinghoriginal: `subscribe_fn` has no slot for the sequence, so #664 stays live for cdylib sources that don't restore their own counter. With no released consumers, worth extending the ABI now instead? `lib/src/config/AGENTS.md` lists the four touchpoints, and it would make the host-side warning unnecessary. Same at 1184.

lib/src/sources/base.rs
```diff
@@ -819,6 +838,17 @@ impl SourceBase {
...
 
+        // Close the checkpoint-restore sequence gap. A resuming subscriber's dedup
+        // high-water mark survives the restart, but this source's monotonic
+        // sequence counter restarts at 1 in every process. Bump the counter above
+        // the checkpoint sequence so post-restart events are stamped with
+        // sequences greater than the mark and are not silently deduplicated away.
+        // `fetch_max` keeps this safe across multiple subscribers and never
+        // rewinds a counter the source has already advanced (e.g. a durable source
+        // that restored its own position).
+        if let Some(resume_seq) = settings.resume_sequence {
```
amansinghoriginal, line 849: The `info!` at 831 fires just before this and prints `resume_from: None`, so the bump leaves no info-level trace. Add `resume_sequence` to that line?

lib/src/queries/manager.rs
```diff
@@ -1106,6 +1106,13 @@ impl Query for DrasiQuery {
                                 checkpoint_sequences_per_source
                                     .insert(settings.source_id.clone(), cp.sequence);
                                 settings.request_position_handle = true;
+                                // Always hand the source the checkpoint sequence so a
+                                // source that restarts its counter from scratch in a new
+                                // process (no durable `source_position` of its own) can
+                                // advance it above the query's dedup high-water mark.
+                                // Without this the first `cp.sequence` post-restart events
+                                // get sequences `1..=cp.sequence` and are silently dropped.
+                                settings.resume_sequence = Some(cp.sequence);
```
amansinghoriginal, line 1115: Non-blocking: nothing enforces this, and both the FFI seam and `subscribe_with_replay` ignore it. Since the mark is consumed here, would warning when a first post-resume event arrives with `sequence <= cp.sequence` be worth a follow-up?

amansinghoriginal (changes requested):
Fix looks right to me. I ran the new test against the parent commit with the fix removed and it fails there, "checkpoint should advance to 6, got seq 3", so it's pinning the actual bug rather than passing by accident.

Small thing in the description: it says the durable DB sources loaded over FFI "already resume via `resume_from`". They do resume from an LSN, but their `next_sequence` still restarts at 1, so they lose the first K changes as well.

A few pre-existing bugs turned up while reading around this: the replay gap check, re-bootstrap double counting aggregates, and the sequence being stamped outside the dispatchers lock. None of them block this PR, so I'll file them on their own.

danielgerlag:
@amansinghoriginal thanks for the careful review, and for running the test against the parent commit to confirm it pins the bug.

You're right about the FFI durable-source case, and I've pushed a fix (3af87491). To confirm your diagnosis: postgres (and the other DB sources) dispatch via `dispatch_event`, which auto-stamps from `SourceBase::next_sequence` (initialised to 1), and none of them call `set_next_sequence`. So over FFI they resume to the correct LSN via `resume_from` but their sequence counter still restarts at 1, and the query's dedup mark K drops the first K post-restart events — exactly the same failure, just reached through the FFI boundary. My earlier warn!-only mitigation didn't even cover them since it was gated on `resume_from.is_none()`.

Rather than just warn, the new commit threads `resume_sequence` through the subscribe ABI so FFI sources get the same counter bump as in-process ones:
- `subscribe_fn` now carries `resume_sequence` + `resume_sequence_present` (mirroring how `resume_from` is passed as ptr+len);
- the host proxy encodes `settings.resume_sequence` and the two plugin-side reconstruction sites decode it back, so the plugin's `subscribe_with_bootstrap` runs `ensure_next_sequence_above`;
- the `(value, present)` encoding lives in one shared, unit-tested helper in `ffi-primitives` (`encode_optional_u64`/`decode_optional_u64`) so host and plugin can't drift.

Tests: round-trip unit tests plus a new cdylib FFI integration test (`test_ffi_resume_sequence_marshalled`) that exercises the full host→plugin path at runtime. I've also corrected the PR description, which wrongly implied FFI durable sources were already safe.

For the three pre-existing issues you spotted (replay gap check, re-bootstrap aggregate double-counting, sequence stamped outside the dispatchers lock) — agreed they're out of scope here; thanks for filing them separately.

components/plugin-sdk/src/ffi/vtables.rs
```diff
@@ -339,9 +339,14 @@ drasi_ffi_primitives::ffi_vtable! {
...
-        /// optional resume_from position bytes.
-        fn subscribe_fn(state: *mut, source_id: FfiStr, enable_bootstrap: bool, query_id: FfiStr, nodes_json: FfiStr, relations_json: FfiStr, resume_from_ptr: *const u8, resume_from_len: u32, request_position_handle: bool) -> *mut FfiSubscriptionResponse,
+        /// Subscribe with query_id, node_labels JSON, relation_labels JSON,
+        /// optional resume_from position bytes, and the optional checkpoint
+        /// `resume_sequence` (carried as a value + present flag, since the ABI
+        /// cannot pass `Option<u64>` directly). `resume_sequence` lets a resuming
+        /// source advance its monotonic sequence counter above the query's dedup
+        /// high-water mark so the first post-restart events are not silently
+        /// dropped (issue #664).
+        fn subscribe_fn(state: *mut, source_id: FfiStr, enable_bootstrap: bool, query_id: FfiStr, nodes_json: FfiStr, relations_json: FfiStr, resume_from_ptr: *const u8, resume_from_len: u32, request_position_handle: bool, resume_sequence: u64, resume_sequence_present: bool) -> *mut FfiSubscriptionResponse,
```
amansinghoriginal, line 349: `subscribe_fn` changed arity but `FFI_SDK_VERSION` is still 0.11.0, and the loader only compares major.minor. An old cdylib loads fine and gets called through this wider pointer; on SysV the extra args are ignored, so the plugin sees no `resume_sequence` and silently returns to the #664 behaviour. Bump to 0.12.0?

amansinghoriginal (changes requested):
All the earlier points are addressed, thanks. The only thing left from my side is the `FFI_SDK_VERSION` bump on the new `subscribe_fn` arity. Happy to approve once that lands.

amansinghoriginal (changes requested):
Could we bump `FFI_SDK_VERSION` to `0.12.0` for the new `subscribe_fn` arity? Happy to approve afterward.

agentofreality (approved):

components/plugin-sdk/src/ffi/vtables.rs
danielgerlag: Good catch — bumped `FFI_SDK_VERSION` to 0.12.0 in 81db7204. You're right that leaving it at 0.11.0 was unsafe: since the loader only rejects on a major.minor mismatch, an old 0.11.0 cdylib would still load into the new host and be called through the widened `subscribe_fn` pointer — on SysV the two extra args are ignored, so the plugin never sees `resume_sequence` and silently reverts to the #664 behaviour (and the reverse pairing is worse, reading uninitialised arg slots). With the bump, any version-mismatched plugin is now rejected at load time before any unsafe call. Documented the reason in the metadata history block and rebuilt the mock cdylib at 0.12.0; the FFI subscribe tests still pass.

danielgerlag:
Closing this as superseded. Issue #664 was independently fixed on `main` by #833 (#827, "Restore per-source sequence counter on resubscribe to preserve restart monotonicity"), which is an equivalent but more polished implementation of the same approach: the `resume_sequence: Option<u64>` subscription field populated from the checkpoint and applied with `fetch_max`, a single `raise_next_sequence` choke point with `set_next_sequence` reworked to never lower the counter, the FFI `subscribe_fn` widened to carry `resume_sequence` (already at `FFI_SDK_VERSION` 0.14.0), and comprehensive unit/e2e/WAL-replay tests.

Because the two changes touch the same struct field and the same single FFI signature, they can't coexist, and resolving the merge conflicts in favour of the canonical `main` version would leave this PR with an effectively empty diff. The fix is already in `main`, so there's nothing left for this PR to add. Thanks @amansinghoriginal for the reviews along the way.

Issue #685, opened 2026-08-01
Remove the plugin log capture store and stop forwarding filtered-out log records over the FFI


amansinghoriginal:
The host-sdk keeps a global store of plugin log records (`captured_logs` and `captured_lifecycles` in callbacks.rs). Every record a plugin sends across the FFI is copied into it, it is unbounded, and nothing in production ever reads it. Its only consumers are the host-sdk integration tests. On a 20-query Postgres CDC workload, heap profiling attributed about 4 GB of live memory to this store. It grows with data volume for the life of the process, so it behaves like a leak.

The store is also a symptom of a deeper problem: plugins forward every log record across the FFI regardless of level. The plugin side installs LogTracer at Trace and its FfiTracingLayer has no filter, so on a host running at info, every debug and trace record is still formatted in the plugin, crosses the FFI, and is copied on the host side, only to be discarded by the host's EnvFilter. Sources log per change event (the Postgres decoder logs per column), so this cost also scales with data volume.

We should fix this properly rather than just bounding the store:

1. Remove the store from the SDK. The log callback is already a parameter to plugin loading, so the integration tests can pass their own capture callback and keep a store they own. The production callbacks then do only their real jobs: forward to the host log framework and route to the component log registry. This removes the global mutex and all retained copies from the hot path. `captured_lifecycles` gets the same treatment.

2. Filter at the producer. Extend the FFI log setup so the host passes its effective level to the plugin, and the plugin's tracing layer drops events below that level before formatting or forwarding them. Deriving that level needs care: `LevelFilter::current()` reads `OFF` both when no subscriber is installed (forward everything) and when the subscriber filters everything out (forward nothing). A value set once at callback installation is enough to start; runtime updates can come later.

Both changes break API and ABI. That is acceptable now and is exactly the kind of change to make before we commit to stability.

PR #687, opened 2026-08-01
Bound the plugin log capture store (#685)

amansinghoriginal:
Interim mitigation for #685, which now tracks the full fix (removing the store from the SDK and filtering at the producer).

Bounds the global plugin log capture store to a drop-oldest ring of 1000 entries, so retained memory no longer scales with traffic.

Capture is also gated on the log crate's max level, and it is worth being precise about what that does and does not do. drasi-server initializes logging through tracing, which pins `log::max_level()` at Trace and filters with EnvFilter downstream, so in that host the gate passes everything and the ring alone provides the bound. The gate filters in hosts whose log max level reflects their configured filter, for example an embedder that initializes env_logger before creating a DrasiLib. When no logger is initialized the level is Off and everything is captured, which preserves existing test behavior.

Heap profiling on a 20-query CDC workload attributed about 4 GB of live memory to the unbounded store.

Touches only the capture path in host-sdk.

### DIFF
--- a/components/host-sdk/src/callbacks.rs
+++ b/components/host-sdk/src/callbacks.rs
@@ -117,10 +117,46 @@ pub struct CapturedLifecycle {
     pub message: String,
 }
 
+/// Maximum entries retained in the captured log store. Every plugin log
+/// record is copied here; without a bound an event-heavy workload retains
+/// gigabytes of log copies for the life of the process (measured: 2.7 GB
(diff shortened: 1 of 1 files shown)


amansinghoriginal:
Closing in favor of a single PR that implements the full fix for #685 (removing the store and filtering at the producer) instead of this interim bound.

PR #688, opened 2026-08-01
Restore backpressure on the bootstrap FFI bridge (#686)

### LINKED ISSUE #686: Bootstrap provider FFI: unbounded buffering and a blocking call on async workers
The bootstrap provider crossing is the last FFI path built on a synchronous run-to-completion call: the host invokes the plugin's `bootstrap_fn` inline from an async task, and it only returns once the whole snapshot has been handed over. Two problems follow.

1. Unbounded memory. The adapters on both sides of the call use unbounded std channels (the host bridge in `host-sdk/src/proxies/bootstrap_provider.rs`, the plugin shim in `plugin-sdk/src/ffi/vtable_gen.rs`). The query-side channel is bounded at 1000, so a slow consumer pushes the entire snapshot into those links. Measured: 50k events fully buffered in-process with the consumer stalled, about 67 MB of RSS. Bounding either link alone does not help, the backlog just moves to the other one.

2. Worker starvation. The inline call pins an async worker for the whole bootstrap. Two concurrent bootstraps on a 2-worker runtime stalled unrelated tasks for 2 seconds. Bounding the bridge without also moving the call off the workers deadlocks the runtime once concurrent bootstraps reach the worker count.

Rather than patching this shape (bounded channels plus blocking pool), replace the contract with the push pattern the codebase already uses for CDC and for source-plugin bootstrap: `FfiBootstrapReceiver` for events, `FfiBootstrapResultReceiver` for completion. `bootstrap_fn` returns immediately, events flow through a bounded sync channel whose callback blocks when full, and the result arrives through the existing callback with its error semantics. This removes the blocking call, the per-bootstrap forwarding threads, and every unbounded link, and cancellation becomes dropping the receiver.

Memory expectation, to be precise: this bounds what the transport can buffer. In-flight events per bootstrap stream become a small constant set by the channel capacities, regardless of snapshot size or consumer speed, so a stalled query no longer pulls the whole snapshot into memory. Peak usage can still spike with data size inside providers that materialize their snapshot before sending (the postgres bootstrapper buffers each full table, oracle the whole snapshot). That is provider work, tracked in #690.

Out of scope: BootstrapContext properties are legacy and unused - oracle source has a bug in using them. These do not cross the FFI boundary even without the change, so they will remain unused. Cleaning them up is not in scope of this issue.

amansinghoriginal:
Closes #686.

Bounds the bootstrap FFI bridge with a small sync channel, mirroring the CDC change bridge, so a full bridge blocks the plugin's send and backpressure reaches the producer. The synchronous bootstrap FFI call moves to the blocking pool: with the bound in place its duration is set by the consumer, and on an async worker enough concurrent bootstraps deadlock the runtime.

Verified on a 20-query workload bootstrapping all queries concurrently: every stream completes with the API responsive throughout.

Independent of the other open memory PRs; touches only the bootstrap proxy in host-sdk.

### DIFF
--- a/components/host-sdk/src/proxies/bootstrap_provider.rs
+++ b/components/host-sdk/src/proxies/bootstrap_provider.rs
@@ -61,70 +61,76 @@ impl BootstrapProvider for BootstrapProviderProxy {
         event_tx: BootstrapEventSender,
         _settings: Option<&SourceSubscriptionSettings>,
     ) -> anyhow::Result<BootstrapResult> {
-        // Build FfiStr arrays for node/relation labels
         let node_label_strs: Vec<String> = request.node_labels.clone();
         let rel_label_strs: Vec<String> = request.relation_labels.clone();
-        let node_ffi: Vec<FfiStr> = node_label_strs
-            .iter()
-            .map(|s| FfiStr::from_str(s))
-            .collect();
-        let rel_ffi: Vec<FfiStr> = rel_label_strs.iter().map(|s| FfiStr::from_str(s)).collect();
+        let query_id = request.query_id.clone();
+        let request_id = request.request_id.clone();
+        let server_id = context.server_id.clone();
+        let source_id = context.source_id.clone();
 
-        let query_id_ffi = FfiStr::from_str(&request.query_id);
-        let request_id_ffi = FfiStr::from_str(&request.request_id);
-        let server_id_ffi = FfiStr::from_str(&context.server_id);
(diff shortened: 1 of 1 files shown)


amansinghoriginal:
Closing in favor of a rework. After review, #686 was rescoped: bounding the host bridge alone does not bound memory, the backlog just moves to the plugin-side shim, and the issue now asks for the push-based receiver contract instead of a blocking call plus blocking pool. Replacement PR to follow.

PR #694, opened 2026-08-03
fix(postgres): unify bootstrap/CDC type mapping and stop silent row drops

### LINKED ISSUE #669: Postgres source (CDC): entire change events silently dropped for rows with uuid, date, time, or jsonb columns
### Steps to reproduce

1. Create a table containing a `uuid` (or `date`, `time`, or `jsonb`) column, set `REPLICA IDENTITY FULL`, and add it to a publication:
   ```sql
   CREATE TABLE t (id int PRIMARY KEY, val uuid);
   ALTER TABLE t REPLICA IDENTITY FULL;
   CREATE PUBLICATION p FOR TABLE t;
   ```
2. Configure the Drasi Postgres source (pgoutput logical replication) against the table.
3. After the source is running, `INSERT INTO t VALUES (1, '550e8400-e29b-41d4-a716-446655440000');`
4. Observe that **no `SourceChange` is emitted at all** for that row.

This was verified end-to-end against PostgreSQL 16 (testcontainers). Single-column tables for `uuid`, `date`, `time`, and `jsonb` each produced **zero** change events, while `bool`/`int`/`float`/`text` worked. When a `uuid` column was mixed into an all-types table, the *entire* insert (every column) was dropped.

### Observed behavior

pgoutput streams column values in **text** format, but `decode_column_value` in `components/sources/postgres/src/decoder.rs` assumes **binary** wire format for several types. The decode error propagates up to `stream.rs`, which catches it and discards the **entire WAL message** — so all columns of the row are lost, not just the offending column. Log line observed:

```
Failed to decode WAL message type I (73): Invalid UUID length: 36
```

Specific root causes in `decode_column_value`:

- **uuid (OID 2950)** — `Uuid::from_slice(data)` expects 16 binary bytes, but receives the 36-char text UUID → `Invalid UUID length: 36`.
- **date (1082)** — `cursor.read_i32()` is applied to ASCII text bytes (e.g. `"2024-06-15"`) → nonsensical/overflowing day count → error.
- **time (1083)** — `cursor.read_i64()` over ASCII text (e.g. `"10:30:45"`) → invalid micros → `NaiveTime` construction fails.
- **jsonb (3802)** — the decoder strips a binary "version byte" via `data[1..]`, but text mode has no such byte, so it removes the leading `{` → invalid JSON → parse error. (Plain `json` OID 114 happens to work.)

Because a single failed column aborts the whole message, **any table with a UUID primary key, or a `date`/`time`/`jsonb` column, loses all live CDC changes silently.**

### Desired behavior

1. Decode `uuid`, `date`, `time`, and `jsonb` from pgoutput's **text** representation (the format actually on the wire), consistent with how `bool` is already handled.
2. A single column that fails to decode must **never** drop the entire row/transaction. At minimum, degrade that one column to a string (or Null) and still emit the change, and log a warning.

### Workaround

None at the source level. Avoid `uuid`/`date`/`time`/`jsonb` columns in replicated tables, or cast/duplicate them into `text`/`timestamp` columns for replication.

### Drasi Version

`drasi-core` repo, branch `main`. `drasi-source-postgres` v0.2.7.

### Operating system

Reproduced on macOS (arm64) with PostgreSQL 16-alpine via Docker/testcontainers.

### Additional context

Empirically measured `ElementValue` emitted by each path (live PostgreSQL 16), showing the CDC drops:

| PG type | Bootstrap | CDC |
|---|---|---|
| uuid | `Null` | **entire row DROPPED** |
| date | `Null` | **entire row DROPPED** |
| time | `Null` | **entire row DROPPED** |
| jsonb | `Null` | **entire row DROPPED** |
| json | `Null` | `String("{\"k\":1}")` |

Relevant code: `components/sources/postgres/src/decoder.rs` (`decode_column_value`, OID arms 2950/1082/1083/3802) and the message-level error handling in `components/sources/postgres/src/stream.rs`.

Related: bootstrap data-loss for the same types, and broader bootstrap↔CDC type inconsistency (filed as separate issues).

### LINKED ISSUE #670: Postgres bootstrap: uuid/date/time/json/jsonb/bytea/array columns silently snapshotted as Null (data loss)
### Steps to reproduce

1. Create a table with columns of types `uuid`, `date`, `time`, `json`, `jsonb`, `bytea`, and/or an array (e.g. `integer[]`), and insert a row with non-null values.
2. Configure the Drasi Postgres source with the Postgres bootstrap provider and subscribe with bootstrap enabled.
3. Inspect the bootstrapped element properties.
4. Observe that every one of those columns is delivered as `ElementValue::Null`, even though the row has real values.

Verified end-to-end against PostgreSQL 16 (testcontainers): a per-type table seeded before the replication slot and drained from the bootstrap channel produced `Null` for `uuid`, `date`, `time`, `json`, `jsonb`, `bytea`, and `integer[]`.

### Observed behavior

`row_to_source_change` in `components/bootstrappers/postgres/src/postgres.rs` only has OID match arms for `bool (16)`, `int2/4/8 (21/23/20)`, `float4/8 (700/701)`, `numeric (1700)`, `text/varchar/name (25/1043/19)`, and `timestamp/timestamptz (1114/1184)`. Everything else falls through to the `_` arm, which does `row.try_get::<_, Option<String>>(idx)`. The Postgres driver rejects a `String` deserialization for `uuid`/`date`/`time`/`json`/`jsonb`/`bytea`/array OIDs, the `Err` is swallowed by `if let Ok(Some(...))`, and the value becomes `ElementValue::Null` — silent data loss.

Contributing factors:
- `get_table_columns` maps any unrecognized `data_type` to OID `25` (text), so e.g. `time`/`bytea`/arrays are treated as text.
- The bootstrapper's `tokio-postgres` dependency enables only `with-chrono-0_4` (`components/bootstrappers/postgres/Cargo.toml`) — there is no `with-uuid` or `with-serde_json`, so those types cannot be deserialized even if arms were added.

### Desired behavior

The bootstrap snapshot should emit real, typed values for `uuid`, `date`, `time`, `json`, `jsonb`, `bytea`, and arrays — and, critically, should produce the **same `ElementValue`** the CDC path produces for the same stored value. A column that genuinely cannot be converted should surface an error/warning rather than silently becoming `Null`.

### Workaround

Cast unsupported columns to `text` in a view/generated column and bootstrap that, or avoid these types in bootstrapped tables.

### Drasi Version

`drasi-core` repo, branch `main`. `drasi-bootstrap-postgres` v0.2.10.

### Operating system

Reproduced on macOS (arm64) with PostgreSQL 16-alpine via Docker/testcontainers.

### Additional context

Measured bootstrap output (live PostgreSQL 16):

| PG type | Bootstrap `ElementValue` |
|---|---|
| uuid | `Null` |
| date | `Null` |
| time | `Null` |
| json | `Null` |
| jsonb | `Null` |
| bytea | `Null` |
| integer[] | `Null` |

Relevant code: `components/bootstrappers/postgres/src/postgres.rs` (`row_to_source_change`, `get_table_columns`) and `components/bootstrappers/postgres/Cargo.toml` (`tokio-postgres` features).

Note the interaction with the CDC-side bug: for `uuid`/`date`/`time`/`jsonb`, bootstrap loses the value **and** CDC drops the entire row — so those columns are effectively unusable today.

### LINKED ISSUE #672: Postgres bootstrap and CDC emit inconsistent ElementValue for the same column (timestamp, numeric, char, bytea, NULL handling)
### Steps to reproduce

For each type below, seed one row **before** the replication slot is created (bootstrap path) and insert an identical row **after** the slot boundary (CDC path), then compare the `ElementValue` emitted by each path. Verified end-to-end against PostgreSQL 16 (testcontainers).

### Observed behavior

Even for the types both paths *can* deliver, the bootstrap path (`components/bootstrappers/postgres/src/postgres.rs::row_to_source_change`, typed `try_get` by OID) and the CDC path (`components/sources/postgres/src/stream.rs` → `PostgresValue::to_json()` → `drasi_lib::sources::manager::convert_json_to_element_value`) are two independent conversions that disagree:

| PG type / value | Bootstrap `ElementValue` | CDC `ElementValue` |
|---|---|---|
| `timestamp '2024-06-15 10:30:45.123456'` | `LocalDateTime(...)` | `String("2024-06-15 10:30:45.123456")` |
| `timestamptz '...+02'` | `ZonedDateTime(...)` | `String("2024-06-15T08:30:45.123456+00:00")` |
| `numeric(12,0)` value `4200` | `Float(4200)` | `Integer(4200)` |
| `char(8)` value `'abc'` | `String("abc     ")` (blank-padded) | `String("abc")` (trimmed) |
| `bytea '\xdeadbeef'` | `Null` | `String("XHhkZWFkYmVlZg==")` (base64 of the **text** `\xdeadbeef`, not the bytes) |
| a `NULL` column | property present as `ElementValue::Null` | property **omitted** from the element entirely |

Details:

- **Temporal types**: bootstrap preserves proper temporal `ElementValue`s; CDC stringifies them (`to_json()` → string, and `convert_json_to_element_value` cannot recover the type). Cypher temporal predicates (`datetime()`, ordering, `duration`) behave differently for the initial snapshot vs. subsequent live updates of the same row.
- **numeric**: a whole-number `numeric` becomes `Float` via bootstrap but `Integer` via CDC (serde_json parses `"4200"` as an integer and `convert_json_to_element_value` tries `as_i64()` first). Same column, different type depending on the value.
- **char(n)**: bootstrap keeps blank padding; CDC trims it (`decoder.rs` OID 1042 `trim_end`).
- **bytea**: CDC base64-encodes the pgoutput **text** representation (`\xdeadbeef`) instead of the raw bytes, so the value is wrong even where it isn't lost; bootstrap loses it (see the bootstrap data-loss issue).
- **NULL handling**: CDC skips null columns (`if !json_value.is_null()` in `stream.rs`), so the property key is absent; bootstrap always inserts the key with `Null` (`postgres.rs`). This changes `IS NULL` / key-existence semantics between snapshot and live rows.

### Desired behavior

Bootstrap and CDC should produce **identical** `ElementValue` (type and value) for the same stored value. Pick a canonical mapping and apply it in both paths:

- temporal → `LocalDateTime`/`ZonedDateTime` in both;
- `numeric` → consistently `Float` (or a lossless decimal string) regardless of scale;
- `char(n)` → consistent padding/trimming;
- `bytea` → base64 of the **actual bytes** in both;
- NULL → consistent property presence (either both omit or both include).

The cleanest structural fix is to share a single conversion function between bootstrap and CDC rather than maintaining two implementations.

### Workaround

Cast columns to types that both paths agree on (e.g. cast timestamps to text) so snapshot and live values match, at the cost of losing temporal typing.

### Drasi Version

`drasi-core` repo, branch `main`. `drasi-source-postgres` v0.2.7, `drasi-bootstrap-postgres` v0.2.10.

### Operating system

Reproduced on macOS (arm64) with PostgreSQL 16-alpine via Docker/testcontainers.

### Additional context

Fully consistent types (for reference): `bool`, `int2/int4/int8`, `real/double`, `numeric` with a fractional part, `text`, `varchar`. Relevant code: `components/sources/postgres/src/types.rs` (`to_json`), `lib/src/sources/manager.rs` (`convert_json_to_element_value`), and `components/bootstrappers/postgres/src/postgres.rs` (`row_to_source_change`).

danielgerlag:
# Description

Postgres bootstrap and CDC previously converted column values on separate code paths, which caused three related failures: CDC could drop entire WAL rows when binary decoding was applied to pgoutput text payloads (#669), bootstrap turned SQL NULL into missing properties via a catch-all `_` arm (#670), and bootstrap vs CDC produced inconsistent `ElementValue` shapes for the same Postgres types (#672).

This change introduces `drasi-postgres-common` with a canonical `PostgresValue::to_element_value` mapping and wires both the bootstrapper and CDC source through it. CDC decoding prefers text when the payload looks like UTF-8 (including ambiguous 8-byte values such as `"10:30:45"`), falls back per column instead of aborting a tuple, and always emits Null property keys. Bootstrap resolves OIDs from `pg_attribute.atttypid` and uses the same typed conversion path.

**Canonical mapping highlights**
- Temporals stay typed (`LocalDateTime` / `ZonedDateTime`)
- Numeric types map to `Float`
- `char` is right-trimmed; uuid/date/time/json/jsonb map to `String`
- `bytea` maps to base64 of the raw bytes
- Arrays map to `List`
- NULL columns keep present keys with `ElementValue::Null` on both paths

**Versions**
- `drasi-postgres-common` 0.1.0 (new)
- `drasi-source-postgres` 0.2.8
- `drasi-bootstrap-postgres` 0.2.11

## Type of change

Fixes: #669
Fixes: #670
Fixes: #672


components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -785,6 +726,131 @@ struct ColumnInfo {
...
+            }
+        }
+    }
+}
+
+fn try_array_as_text(row: &Row, idx: usize) -> PostgresValue {
+    match row.try_get::<_, Option<String>>(idx) {
+        Ok(Some(s)) => match drasi_postgres_common::decode_text_to_postgres_value(&s, 1009) {
+            Ok(v) => v,
+            Err(_) => PostgresValue::Text(s),
```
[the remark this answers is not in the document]
danielgerlag: Fixed: `try_array_as_text` now takes the real `array_oid` and passes it to `decode_text_to_postgres_value`, so int/bool/text array fallbacks keep element types.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,332 @@
...
+}
+
+fn parse_array_element(raw: &str, element_oid: u32) -> Result<PostgresValue> {
+    let t = raw.trim();
+    if t.eq_ignore_ascii_case("NULL") && !t.starts_with('"') {
+        return Ok(PostgresValue::Null);
+    }
+    // Strip surrounding quotes if present (already removed by parser for quoted)
+    decode_text_to_postgres_value(t, element_oid)
+}
```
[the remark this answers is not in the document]
danielgerlag: Fixed: the array text parser now tracks whether an element used quote delimiters, so unquoted `NULL` stays SQL null and quoted `"NULL"` remains the literal string. Added unit coverage for both cases.

components/sources/postgres/tests/postgres_helpers.rs
```diff
@@ -401,3 +401,57 @@ pub async fn insert_decimal_test_row(
...
+            c char(8),
+            ts timestamp,
+            tstz timestamptz,
+            maybe_null text
+        )"
+    );
+    execute_sql(client, &create_sql).await?;
+    let replica_sql = format!("ALTER TABLE {table_name} REPLICA IDENTITY FULL");
+    execute_sql(client, &replica_sql).await?;
+    Ok(())
```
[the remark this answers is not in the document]
danielgerlag: Fixed: `create_type_parity_table` now uses `quote_ident` for CREATE/ALTER, matching `insert_type_parity_row`.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,367 @@
...
+                was_quoted = true;
+            }
+            ',' if !in_quotes => {
+                elements.push(parse_array_element(&cur, element_oid, was_quoted)?);
+                cur.clear();
+                was_quoted = false;
+            }
+            _ => cur.push(c),
+        }
+    }
```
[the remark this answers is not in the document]
danielgerlag: Already addressed in the previous commit: `parse_array_text` tracks `was_quoted` and passes it into `parse_array_element`, with unit coverage for `{NULL,"NULL","a"}`.

components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -785,6 +726,134 @@ struct ColumnInfo {
...
+                }
+            }
+        }
+    }
+}
+
+fn try_array_as_text(row: &Row, idx: usize, array_oid: u32) -> PostgresValue {
+    match row.try_get::<_, Option<String>>(idx) {
+        Ok(Some(s)) => match drasi_postgres_common::decode_text_to_postgres_value(&s, array_oid) {
+            Ok(v) => v,
```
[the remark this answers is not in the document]
danielgerlag: Already fixed to pass the real **array** OID (e.g. 1007) into `decode_text_to_postgres_value`. Passing the element OID would be incorrect here: the shared decoder expects the column/array OID and uses `array_element_oid` internally when parsing `{...}` literals.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,367 @@
...
+            let value = match trimmed {
+                "t" | "true" => true,
+                "f" | "false" => false,
+                _ => return Err(anyhow!("Invalid boolean value: {text}")),
+            };
+            Ok(PostgresValue::Bool(value))
+        }
+        oid::INT2 => {
+            Ok(PostgresValue::Int2(trimmed.parse::<i16>().map_err(
+                |e| anyhow!("Failed to parse int2 from '{text}': {e}"),
```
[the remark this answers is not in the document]
danielgerlag: Fixed: decode error messages now include the OID and parse error only (no raw column text).

components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -785,6 +726,134 @@ struct ColumnInfo {
     type_oid: i32,
 }
 
+/// Read a single column from a tokio-postgres `Row` into the shared `PostgresValue`.
+fn row_column_to_postgres_value(row: &Row, idx: usize, type_oid: u32) -> PostgresValue {
+    // Helper: Option try_get that maps None → Null
+    macro_rules! opt {
```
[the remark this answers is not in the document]
danielgerlag: Acknowledged. Leaving the tokio-postgres `Row` adapter in the bootstrapper for now to avoid a feature-gated dependency on `tokio-postgres` in `postgres-common` (CDC does not need it). Happy to revisit if a second consumer shows up.

components/postgres-common/src/value.rs
```diff
@@ -0,0 +1,266 @@
...
+    Timestamp(NaiveDateTime),
+    TimestampTz(DateTime<Utc>),
+    Date(NaiveDate),
+    Time(NaiveTime),
+    Json(JsonValue),
+    Jsonb(JsonValue),
+    Array(Vec<PostgresValue>),
+    Composite(HashMap<String, PostgresValue>),
+    Bytea(Vec<u8>),
+}
```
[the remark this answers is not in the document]
danielgerlag: Fixed: removed the unused public `Composite` variant and its conversion arms until composite support is implemented end-to-end.

components/postgres-common/src/oid.rs
```diff
@@ -0,0 +1,160 @@
...
+    // Arrays report data_type = ARRAY and the element type in udt_name (e.g. `_int4`).
+    if data_type.eq_ignore_ascii_case("ARRAY") {
+        if let Some(udt) = udt_name {
+            return match udt {
+                "_bool" => BOOL_ARRAY,
+                "_bytea" => BYTEA_ARRAY,
+                "_int2" => INT2_ARRAY,
+                "_int4" => INT4_ARRAY,
+                "_int8" => INT8_ARRAY,
+                "_text" => TEXT_ARRAY,
```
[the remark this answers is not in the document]
danielgerlag: Fixed: added unit tests in `oid.rs` for integer, ARRAY/_int4, unknown array, USER-DEFINED uuid, and unknown type fallbacks.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,367 @@
...
+        oid::BYTEA => {
+            let bytes = parse_bytea_text(trimmed)?;
+            Ok(PostgresValue::Bytea(bytes))
+        }
+        _ => Ok(PostgresValue::Text(text.to_string())),
+    }
+}
+
+fn decode_timestamptz_text(trimmed: &str, original: &str) -> Result<PostgresValue> {
+    // RFC3339 first
```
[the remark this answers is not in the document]
danielgerlag: Fixed: added unit tests for RFC3339, Postgres offset with fractional seconds, no-offset UTC fallback, and instant parity across formats.

components/postgres-common/src/value.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /// Stable string form for element ID key parts.
+    pub fn to_key_string(&self) -> Option<String> {
+        match self {
+            PostgresValue::Null => None,
+            PostgresValue::Bool(b) => Some(b.to_string()),
+            PostgresValue::Int2(i) => Some(i.to_string()),
+            PostgresValue::Int4(i) => Some(i.to_string()),
+            PostgresValue::Int8(i) => Some(i.to_string()),
```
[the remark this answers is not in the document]
danielgerlag: Fixed: added unit coverage for `Null`, `Json`/`Jsonb`, and `Array` → `to_element_value()` (Composite removed).

components/postgres-common/src/value.rs
```diff
@@ -0,0 +1,266 @@
...
+                ElementValue::List(arr.iter().map(|v| v.to_element_value()).collect())
+            }
+            PostgresValue::Composite(_) => {
+                ElementValue::String(Arc::from(self.to_json().to_string()))
+            }
+            PostgresValue::Bytea(bytes) => ElementValue::String(Arc::from(encode_base64(bytes))),
+        }
+    }
+
+    pub fn is_null(&self) -> bool {
```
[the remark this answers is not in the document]
danielgerlag: Fixed: added a doc comment on `is_null`.

components/postgres-common/src/value.rs
```diff
@@ -0,0 +1,266 @@
...
+    if let Some(hex) = trimmed
+        .strip_prefix("\\x")
+        .or_else(|| trimmed.strip_prefix(r"\x"))
+    {
+        if hex.len() % 2 != 0 {
+            anyhow::bail!("Odd-length bytea hex string");
+        }
+        let mut out = Vec::with_capacity(hex.len() / 2);
+        for i in (0..hex.len()).step_by(2) {
+            let byte = u8::from_str_radix(&hex[i..i + 2], 16)
```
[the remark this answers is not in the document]
danielgerlag: Acknowledged. Leaving the small manual hex loop for now to avoid an extra dependency for a hot-path-light helper; happy to switch to `hex` in a follow-up if preferred.

components/sources/postgres/src/stream.rs
```diff
@@ -537,16 +532,10 @@ impl ReplicationStream {
...
         for (i, column) in relation.columns.iter().enumerate() {
             if let Some(value) = new_tuple.get(i) {
-                let json_value = value.to_json();
-                if !json_value.is_null() {
-                    after_properties.insert(
-                        &column.name,
-                        drasi_lib::sources::manager::convert_json_to_element_value(&json_value),
-                    );
-                }
+                after_properties.insert(&column.name, value.to_element_value());
```
amansinghoriginal, line 571: decoder.rs:360 maps both b'n' and b'u' to PostgresValue::Null, so a TOASTed column this UPDATE did not modify arrives here as an explicit Null. That defeats merge_missing_properties (continuous_query.rs:316), which merges vacant keys only, so the stored value is overwritten.

Repro: 4096 char column, UPDATE a different column. base keeps the value, this branch returns null.

A PostgresValue::UnchangedToast variant that this loop skips fixes it without touching the NULL parity above.

components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -785,6 +726,134 @@ struct ColumnInfo {
...
+            Ok(Some(v)) => PostgresValue::Array(v.into_iter().map(PostgresValue::Int2).collect()),
+            Ok(None) => PostgresValue::Null,
+            Err(_) => try_array_as_text(row, idx, type_oid),
+        },
+        1000 /* bool[] */ => match row.try_get::<_, Option<Vec<bool>>>(idx) {
+            Ok(Some(v)) => PostgresValue::Array(v.into_iter().map(PostgresValue::Bool).collect()),
+            Ok(None) => PostgresValue::Null,
+            Err(_) => try_array_as_text(row, idx, type_oid),
+        },
+        _ => {
```
amansinghoriginal, line 847: try_get::<Option<String>> can never succeed here. tokio-postgres checks T::accepts before decoding and String accepts only VARCHAR, TEXT, BPCHAR, NAME and UNKNOWN. Every array OID not listed above (uuid[], varchar[], float4[], float8[], numeric[], json[], jsonb[], timestamp[], timestamptz[], date[], time[], bytea[]) therefore becomes Null while CDC returns a List.

try_array_as_text has the same problem, so routing arrays to it does not help. Typed Vec arms are needed.

components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -785,6 +726,134 @@ struct ColumnInfo {
...
+                        Ok(b) => PostgresValue::Bytea(b),
+                        Err(_) => PostgresValue::Text(s),
+                    },
+                    Ok(None) => PostgresValue::Null,
+                    Err(_) => PostgresValue::Null,
+                }
+            }
+        },
+        // Common 1-D arrays
+        1007 /* int4[] */ => match row.try_get::<_, Option<Vec<i32>>>(idx) {
```
amansinghoriginal: Vec<i32> cannot represent a NULL element, so integer[] '{1,NULL,3}' fails try_get and falls through to Null while CDC returns List[1,Null,3]. Same for the text[], int8[], int2[] and bool[] arms below.

Vec<Option<T>> fixes all five.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,420 @@
...
+        return Ok(PostgresValue::Array(vec![]));
+    }
+
+    let mut elements = Vec::new();
+    let mut cur = String::new();
+    let mut in_quotes = false;
+    let mut escape = false;
+    // True if this element used PG array double-quote delimiters.
+    let mut was_quoted = false;
+    for c in inner.chars() {
```
amansinghoriginal, line 194: No brace depth tracking, so a multi-dimensional array is split on its inner commas. text[][] '{{a,b},{c,d}}' decodes to List["{a", "b}", "{c", "d}"]. Element types that never fail to parse make this silent; numeric element types error out and degrade to a string instead.

Multi-dim arrays share the OID of their 1-D form, so any array column can hit this. Detecting and rejecting them is enough here; full nested support could be a follow-up issue.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,420 @@
...
+            _ => cur.push(c),
+        }
+    }
+    elements.push(parse_array_element(&cur, element_oid, was_quoted)?);
+
+    Ok(PostgresValue::Array(elements))
+}
+
+fn parse_array_element(raw: &str, element_oid: u32, was_quoted: bool) -> Result<PostgresValue> {
+    let t = raw.trim();
```
amansinghoriginal, line 222: trim() also strips quoted elements, but PostgreSQL quotes precisely to preserve that whitespace. text[] '{"  padded  ","x"}' decodes to ["padded", "x"], while bootstrap reads the binary array and keeps the padding.

was_quoted is already threaded in, but `if was_quoted { raw }` is not quite the rule: characters after a closing quote land in the same buffer, so '{"a b" , x}' would keep the trailing space. Strip only whitespace accumulated while not in quotes.

components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -505,27 +510,39 @@ impl PostgresBootstrapHandler {
         transaction: &Transaction<'_>,
         table_name: &str,
     ) -> Result<Vec<ColumnInfo>> {
+        // Prefer pg_attribute.atttypid for accurate OIDs (arrays, domains, etc.).
```
amansinghoriginal: atttypid is the domain's own OID, not the base type, so domain columns match no arm and fall to the catch-all. information_schema.data_type reported the base type, so this regresses: a column of CREATE DOMAIN posint_dom AS integer bootstraps as Null here and as Integer(42) on base.

Resolving through pg_type.typbasetype when typtype = 'd' keeps the catalog query and fixes domains.

components/sources/postgres/src/decoder.rs
```diff
@@ -454,12 +482,17 @@ impl PgOutputDecoder {
             }
             20 => {
                 // int8
-                if data.len() == 8 {
-                    // Binary format
+                if looks_like_utf8_text(data) {
```
amansinghoriginal, line 485: looks_like_utf8_text was added to int2, int4, int8, timestamp, timestamptz, date and time, but not to float4 (700) and float8 (701) just below, which still choose binary on data.len() == 4 or == 8. Any float whose text form is exactly that long is read as IEEE bits: 3.141593 decodes to 3.669658951197582e-62, and real 1.25 to 2.5348872e-9. Bootstrap returns the correct value, so this is also a parity break.

Pre-existing, but it is the same heuristic this PR is fixing and the arms are a few lines apart.

components/postgres-common/src/value.rs
```diff
@@ -0,0 +1,282 @@
...
```
amansinghoriginal, line 188: &hex[i..i + 2] slices at byte offsets while the % 2 guard counts bytes, so a multi-byte char panics with "end byte index 2 is not a char boundary". Reachable from decoder.rs:691, which builds the input with from_utf8_lossy. Iterating over bytes avoids it.

components/postgres-common/src/value.rs
```diff
@@ -0,0 +1,282 @@
...
+pub fn encode_base64(input: &[u8]) -> String {
+    use base64::Engine;
+    base64::engine::general_purpose::STANDARD.encode(input)
+}
+
+/// Parse PostgreSQL text-format bytea (`\xdeadbeef` or escaped).
+pub fn parse_bytea_text(text: &str) -> anyhow::Result<Vec<u8>> {
+    let trimmed = text.trim();
+    if let Some(hex) = trimmed
+        .strip_prefix("\\x")
```
amansinghoriginal, line 180: "\\x" and r"\x" are the same literal, so the or_else branch is dead. The doc comment says escaped format is handled, but escape output such as \336\255 falls through to the raw-bytes fallback and is stored as the bytes of the escape text. Same duplicate at decoder.rs:693.

components/sources/postgres/src/types.rs
```diff
@@ -363,178 +181,64 @@ mod tests {
...
-        let bootstrap = decode_column_value_text("9.87654321", 701).unwrap();
-        assert_eq!(cdc, bootstrap, "float8: CDC vs bootstrap mismatch");
+        assert_eq!(cdc, bootstrap);
     }
 
     #[test]
     fn parity_numeric() {
         let dec = Decimal::from_str_exact("123.45").unwrap();
         let cdc = PostgresValue::Numeric(dec).to_element_value();
         let bootstrap = decode_column_value_text("123.45", 1700).unwrap();
```
amansinghoriginal, line 202: Both sides of these parity assertions end in to_element_value(), and decode_column_value_text is the shared text decoder rather than the bootstrap read path (row_column_to_postgres_value / try_get). So they cannot detect a bootstrap/CDC divergence. Every bootstrap-side gap noted elsewhere in this review passes these tests.

components/sources/postgres/src/decoder.rs
```diff
@@ -910,10 +914,13 @@ mod tests {
...
         let ev = decode_column_value_text("2024-06-15 10:30:45.123456+0200", 1184).unwrap();
         match ev {
             ElementValue::ZonedDateTime(dt) => {
-                assert_eq!(dt.offset().local_minus_utc(), 7200); // +02:00
+                assert_eq!(dt.offset().local_minus_utc(), 0);
+                // 10:30:45.123456+02:00 → 08:30:45.123456 UTC
+                assert_eq!(dt.naive_utc().to_string(), "2024-06-15 08:30:45.123456");
             }
             other => panic!("Expected ZonedDateTime, got {other:?}"),
         }
```
amansinghoriginal, line 937: test_decode_timestamptz_text_short_offset (line 929) and test_decode_timestamptz_text_short_negative_offset (line 976) have no #[test] attribute and never run. They are the only coverage for the %#z short offset branches this PR reworked.

components/bootstrappers/postgres/src/postgres.rs
```diff
@@ -505,27 +510,39 @@ impl PostgresBootstrapHandler {
...
+                 WHERE n.nspname = 'public'
+                   AND c.relname = $1
+                   AND a.attnum > 0
+                   AND NOT a.attisdropped
+                 ORDER BY a.attnum",
+                &[&table_name],
+            )
+            .await;
+
+        if let Ok(rows) = catalog_result {
```
amansinghoriginal, line 536: If the pg_attribute query errors, the enclosing transaction is already aborted, so this fallback fails too with 25P02 and oid_from_information_schema is unreachable in practice. Either drop the fallback or run it in its own transaction.

components/postgres-common/src/oid.rs
```diff
@@ -0,0 +1,199 @@
...
+        "smallint" => INT2,
+        "integer" => INT4,
+        "text" => TEXT,
+        "json" => JSON,
+        "real" => FLOAT4,
+        "double precision" => FLOAT8,
+        "character" => CHAR,
+        "character varying" => VARCHAR,
+        "date" => DATE,
+        "time without time zone" | "time with time zone" => TIME,
```
amansinghoriginal, line 100: timetz is OID 1266, not 1083, and nothing handles 1266. Folding it into TIME means the bootstrap does try_get::<NaiveTime> against a TIMETZ column, and on the CDC side the text "10:30:45+02" matches neither NaiveTime pattern in decode.rs.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,420 @@
...
+}
+
+/// Parse a PostgreSQL 1-D array text literal like `{1,2,NULL,"a"}`.
+///
+/// Quote markers are tracked separately from element content so a quoted
+/// `"NULL"` string is preserved while an unquoted `NULL` becomes
+/// [`PostgresValue::Null`].
+fn parse_array_text(text: &str, element_oid: u32) -> Result<PostgresValue> {
+    let s = text.trim();
+    let inner = if let Some(body) = s.strip_prefix('{').and_then(|x| x.strip_suffix('}')) {
```
amansinghoriginal, line 177: PostgreSQL prints an explicit dimension prefix when a lower bound is not 1, for example [3:5]={1,2,3}. strip_prefix('{') fails on that, so the value degrades to a string here.

components/postgres-common/src/decode.rs
```diff
@@ -0,0 +1,420 @@
...
+                |e| anyhow!("Failed to parse float4 (OID {type_oid}): {e}"),
+            )?))
+        }
+        oid::FLOAT8 => {
+            Ok(PostgresValue::Float8(trimmed.parse::<f64>().map_err(
+                |e| anyhow!("Failed to parse float8 (OID {type_oid}): {e}"),
+            )?))
+        }
+        oid::NUMERIC => {
+            let value = Decimal::from_str_exact(trimmed)
```
amansinghoriginal, line 76: Decimal cannot represent PostgreSQL NaN, Infinity, or magnitudes above 7.9228e28, so those legal numeric values return Err here and degrade to String on CDC, while the bootstrap NUMERIC arm ends at Null. Worth a follow-up issue to settle on one representation, either Float or a lossless decimal string.

components/sources/postgres/src/stream.rs
```diff
@@ -634,12 +623,8 @@ impl ReplicationStream {
             for (i, column) in relation.columns.iter().enumerate() {
                 if keys.contains(&column.name) {
                     if let Some(value) = tuple.get(i) {
-                        let json_val = value.to_json();
-                        if !json_val.is_null() {
-                            // Remove quotes from JSON string representation
-                            let val_str = json_val.to_string();
-                            let cleaned = val_str.trim_matches('"');
-                            key_parts.push(cleaned.to_string());
+                        if let Some(part) = value.to_key_string() {
```
amansinghoriginal, line 659: table_primary_keys is initialized empty and never written (see the note at line 86), so key_columns is None whenever config.table_keys has no entry for the table and the ID falls back to Uuid::new_v4() at line 642, a fresh identity for every change. Bootstrap meanwhile detects real primary keys from the catalog, so the IDs never match and each CDC change creates a new element.

Pre-existing and outside this PR, but it makes the bootstrap/CDC parity established here conditional on table_keys being set by hand. Worth a follow-up issue.

components/sources/postgres/tests/integration_tests.rs
```diff
@@ -1974,3 +1974,260 @@ async fn test_postgres_stop_restart_recovers() -> Result<()> {
...
+/// Bootstrap ↔ CDC ElementValue parity for previously broken Postgres types
+/// (issues #669, #670, #672).
+///
+/// - Bootstrap must not Null-out uuid/date/time/json/jsonb/bytea/arrays
+/// - CDC must emit the row (not drop it) for uuid/date/time/jsonb
+/// - Both paths must produce identical ElementValues for the same stored values
+/// - NULL columns are present as ElementValue::Null on both paths
+#[tokio::test]
+#[serial]
+#[ignore]
```
amansinghoriginal, line 2158: This is the only test covering the central claim of the PR, and CI runs plain cargo test --workspace with no --ignored, so it never executes. Consistent with the rest of the file, so worth a follow-up issue to run these somewhere rather than a change here.

amansinghoriginal (changes requested):
Verified end to end against PostgreSQL 16 on both base and this branch, using a per-type table seeded before the replication slot (bootstrap) and an identical row inserted after it (CDC), diffing the emitted ElementValue.

Confirmed fixed: bootstrap and CDC now agree for uuid, date, time, json, jsonb, bytea, integer[], timestamp, timestamptz, numeric, char and NULL. bytea is base64 of the real bytes on both paths. Element IDs match across handover when table_keys is set.

Two behaviours regress against base:

1. An UPDATE that leaves a TOASTed column untouched writes NULL over the stored value. For a 4096 char column, base keeps it and this branch returns null.
2. A column typed as a domain over a non-text base type bootstraps as Null. base returned Integer(42) for a domain over integer.

Three pre-existing problems are out of scope here and want their own issues: CDC element IDs fall back to a random UUID per change when table_keys is unset while bootstrap uses catalog primary keys; an UPDATE that changes a primary key orphans the element under the old ID; a NULL member of a composite key is dropped from the ID so distinct rows can collide.

components/sources/postgres/src/stream.rs
danielgerlag: Fixed and verified locally against PG 16.

- Decoder maps `b'u'` → `PostgresValue::UnchangedToast` (not Null)
- Stream omits those properties on insert/update so `merge_missing_properties` can keep the stored value
- SQL NULL (`b'n'`) still emits an explicit Null key for bootstrap/CDC parity

Regression test: `test_toast_unchanged_omitted_on_update` (EXTERNAL storage + 16KB payload; UPDATE only a non-TOAST column). Confirmed `big` is absent on the CDC update, not Null.

components/bootstrappers/postgres/src/postgres.rs
danielgerlag: Fixed and verified locally against PG 16.

Bootstrap column OID query now joins `pg_type` and uses `typbasetype` when `typtype = 'd'`, so a domain over integer is read as int4.

Regression test: `test_domain_over_integer_bootstrap` with `CREATE DOMAIN posint_dom AS integer` — bootstrap now emits `Integer(42)` instead of Null.

danielgerlag:
Addressed the two regressions from the review (verified locally on PG 16):

1. **Unchanged TOAST** — `b'u'` → `PostgresValue::UnchangedToast`; property omitted so UPDATE does not clobber stored large values with Null. Test: `test_toast_unchanged_omitted_on_update`.
2. **Domains** — bootstrap resolves `pg_type.typbasetype` when `typtype = 'd'`. Test: `test_domain_over_integer_bootstrap`.

Other items from the review are acknowledged as follow-ups / out of scope for this PR.

components/bootstrappers/postgres/src/postgres.rs
danielgerlag: Fixed: bootstrap array arms now use `Vec<Option<T>>` so elements like `{1,NULL,3}` map to `List[1, Null, 3]` instead of failing `try_get` and becoming Null.

components/sources/postgres/src/decoder.rs
danielgerlag: Fixed: float4/float8 now use the same `looks_like_utf8_text` preference as int/time before falling back to binary length. Added unit tests for 4-byte `"1.25"` and 8-byte `"1.234567"` text payloads.

components/sources/postgres/src/decoder.rs
danielgerlag: Fixed: added the missing `#[test]` attributes. Both short-offset tests now run and pass.

components/bootstrappers/postgres/src/postgres.rs
danielgerlag: Acknowledged, deferring. Typed `Vec`/`Vec<Option<T>>` arms cover the common array OIDs on the happy path; expanding binary/text fallbacks for every array OID is follow-up scope beyond this PR.

components/postgres-common/src/decode.rs
danielgerlag: Acknowledged, deferring. Multi-dimensional array text parsing is out of scope for the #669/#670/#672 fixes; we can reject multi-dim or add nested List support in a follow-up.

components/postgres-common/src/decode.rs
danielgerlag: Acknowledged, deferring. Preserving quoted array whitespace is a real subtlety but not part of the core type-parity issues this PR targets. Follow-up is fine.

components/postgres-common/src/value.rs
danielgerlag: Acknowledged, deferring. Normal pgoutput `\x` hex is ASCII; hardening the nibble loop against multi-byte input is a low-priority nit for a follow-up.

components/postgres-common/src/value.rs
danielgerlag: Acknowledged, deferring. The duplicate prefix check is dead code and escaped (non-hex) bytea is not a goal of this PR. Doc cleanup / real escape support can be a follow-up.

components/sources/postgres/src/types.rs
danielgerlag: Acknowledged. Those unit checks exercise the shared canonical mapping; bootstrap `try_get` vs CDC is covered by the ignored integration parity tests. Not changing the unit style in this PR.

components/bootstrappers/postgres/src/postgres.rs
danielgerlag: Acknowledged, deferring. Running the information_schema fallback on a fresh transaction (or dropping it) is a reasonable cleanup but not required for the issue fixes here.

components/postgres-common/src/oid.rs
danielgerlag: Acknowledged, deferring. Proper `timetz` (OID 1266) support is niche relative to this PR; tracking as a follow-up is fine.

components/postgres-common/src/decode.rs
danielgerlag: Acknowledged, deferring. Non-default array lower bounds (`[3:5]={...}`) are an edge case outside the claimed issue scope.

components/postgres-common/src/decode.rs
danielgerlag: Acknowledged, deferring. NaN/Infinity/out-of-range numeric needs an explicit product choice (Float vs string). Better as its own issue.

components/sources/postgres/src/stream.rs
danielgerlag: Acknowledged as pre-existing / out of scope (as in the review summary). Populating catalog PKs into `table_primary_keys` deserves a separate issue.

components/sources/postgres/tests/integration_tests.rs
danielgerlag: Acknowledged as pre-existing for this suite. Running ignored Docker integration tests in CI is a separate infrastructure change, not part of this PR.

components/bootstrappers/postgres/src/postgres.rs
amansinghoriginal: row_column_to_postgres_value is new in this PR, so this is not pre-existing scope. #670 names integer[] explicitly, and this fixes integer[], text[], int2[], int8[] and bool[] while varchar[], uuid[], float4[], float8[] and numeric[] still bootstrap as Null against a CDC List.

Five arms in the style you already use:

```suggestion
        1015 /* varchar[] */ => match row.try_get::<_, Option<Vec<Option<String>>>>(idx) {
            Ok(Some(v)) => PostgresValue::Array(
                v.into_iter()
                    .map(|o| o.map(PostgresValue::Varchar).unwrap_or(PostgresValue::Null))
                    .collect(),
            ),
            Ok(None) => PostgresValue::Null,
            Err(_) => try_array_as_text(row, idx, type_oid),
        },
        1021 /* float4[] */ => match row.try_get::<_, Option<Vec<Option<f32>>>>(idx) {
            Ok(Some(v)) => PostgresValue::Array(
                v.into_iter()
                    .map(|o| o.map(PostgresValue::Float4).unwrap_or(PostgresValue::Null))
                    .collect(),
            ),
            Ok(None) => PostgresValue::Null,
            Err(_) => try_array_as_text(row, idx, type_oid),
        },
        1022 /* float8[] */ => match row.try_get::<_, Option<Vec<Option<f64>>>>(idx) {
            Ok(Some(v)) => PostgresValue::Array(
                v.into_iter()
                    .map(|o| o.map(PostgresValue::Float8).unwrap_or(PostgresValue::Null))
                    .collect(),
            ),
            Ok(None) => PostgresValue::Null,
            Err(_) => try_array_as_text(row, idx, type_oid),
        },
        1231 /* numeric[] */ => match row.try_get::<_, Option<Vec<Option<rust_decimal::Decimal>>>>(idx) {
            Ok(Some(v)) => PostgresValue::Array(
                v.into_iter()
                    .map(|o| o.map(PostgresValue::Numeric).unwrap_or(PostgresValue::Null))
                    .collect(),
            ),
            Ok(None) => PostgresValue::Null,
            Err(_) => try_array_as_text(row, idx, type_oid),
        },
        2951 /* uuid[] */ => match row.try_get::<_, Option<Vec<Option<uuid::Uuid>>>>(idx) {
            Ok(Some(v)) => PostgresValue::Array(
                v.into_iter()
                    .map(|o| o.map(PostgresValue::Uuid).unwrap_or(PostgresValue::Null))
                    .collect(),
            ),
            Ok(None) => PostgresValue::Null,
            Err(_) => try_array_as_text(row, idx, type_oid),
        },
        _ => {
```

Verified with this applied: bootstrap and CDC agree for uuid[], varchar[], float8[] and numeric[]. rustfmt and clippy clean. Rarer array OIDs are a fine follow-up.

components/postgres-common/src/decode.rs
amansinghoriginal: parse_array_text is new in this PR, and text[][] currently decodes to List["{a", "b}", "{c", "d}"] on CDC. Base emitted the raw literal, so this is the one place the new code is worse than what it replaced. Multi-dim arrays share the OID of their 1-D form, so no configuration avoids it.

Not asking for nested List support. This restores the base behaviour of keeping the literal:

```suggestion
    // Multi-dimensional arrays share the OID of their 1-D form. A brace outside
    // quotes means this is not a flat list, so keep the literal rather than
    // splitting it into garbage elements.
    let mut nested_q = false;
    let mut nested_esc = false;
    for c in inner.chars() {
        if nested_esc {
            nested_esc = false;
        } else if c == '\\' && nested_q {
            nested_esc = true;
        } else if c == '"' {
            nested_q = !nested_q;
        } else if c == '{' && !nested_q {
            return Ok(PostgresValue::Text(text.to_string()));
        }
    }
    for c in inner.chars() {
```

Verified: text[][] returns String("{{a,b},{c,d}}") on CDC, while {"{a}"} and {"a{b}c","x"} still parse as arrays. rustfmt and clippy clean.

components/postgres-common/src/decode.rs
amansinghoriginal: Fair that this is not a regression. Base returned a raw string here, so there was no List to corrupt.

The narrower point is that base failed loudly and this fails silently, and parse_array_element is new in this PR. A text[] element of '  padded  ' comes back as 'padded' on CDC while bootstrap keeps the padding.

One line. PostgreSQL's array_out never emits whitespace outside quotes, so trimming only unquoted elements covers everything that can arrive on the wire or from a text cast.

```suggestion
    // Quoted elements are verbatim: PostgreSQL quotes precisely to preserve
    // leading/trailing whitespace. Only unquoted elements are safe to trim.
    let t = if was_quoted { raw } else { raw.trim() };
```

Verified with this applied: 31 postgres-common unit tests pass, and bootstrap and CDC both return ["  padded  ", "x"] for a text[] column.

Not a blocker either way.

components/postgres-common/src/oid.rs
amansinghoriginal: Agreed, unchanged from base.

components/postgres-common/src/decode.rs
amansinghoriginal: Agreed, improved over base.

components/postgres-common/src/value.rs
amansinghoriginal: Agreed, cosmetic.

components/bootstrappers/postgres/src/postgres.rs
amansinghoriginal: Agreed, not blocking.

components/postgres-common/src/value.rs
amansinghoriginal: Agreed. pgoutput hex is ASCII, so the slice is not reachable in practice.

components/postgres-common/src/decode.rs
amansinghoriginal: Agreed, this needs a product decision rather than a patch here. Please open the follow-up so the choice is recorded against the new crate.

amansinghoriginal (approved):
Re-reviewed the current head. The regressions from my previous review are fixed, and the remaining pre-release Postgres parity work is tracked in #751. Approving.

agentofreality (approved):

PR #696, opened 2026-08-03
fix(mssql): unify bootstrap/CDC ElementValue conversion

### LINKED ISSUE #673: MSSQL bootstrap and CDC emit inconsistent ElementValue (uniqueidentifier, varbinary, datetime format); bootstrap does not use the shared converter
### Steps to reproduce

1. Create a SQL Server table with `UNIQUEIDENTIFIER`, `VARBINARY`, `DATETIME`, and `DATETIME2` columns; enable CDC.
2. Insert a row, configure the Drasi MSSQL source + bootstrap provider, subscribe with bootstrap enabled.
3. Compare the bootstrap snapshot values with live CDC values for the same row.

Verified end-to-end against SQL Server. Bootstrap values from the real `MsSqlBootstrapProvider`; CDC values from the production CDC conversion function `drasi_source_mssql::types::extract_properties_from_cdc_row`.

### Observed behavior

The MSSQL bootstrapper uses its **own** `convert_column_value` (`components/bootstrappers/mssql/src/mssql.rs`) instead of the **shared** `drasi_mssql_common::types::extract_column_value` that the CDC/source path uses (`components/sources/mssql/src/stream.rs` → `extract_properties_from_cdc_row`). The two disagree:

| Column type | Bootstrap | CDC |
|---|---|---|
| `uniqueidentifier` | `Null` (logs "Unsupported column type Guid…") | `String("550e8400-e29b-41d4-a716-446655440000")` |
| `varbinary` | `Null` | `String("0xdeadbeef")` |
| `datetime` | `String("2024-06-15 10:30:45.123333333")` | `String("2024-06-15T10:30:45.123")` |
| `datetime2` | `String("2024-06-15 10:30:45.123456700")` | `String("2024-06-15T10:30:45.123")` |

Details:

- **uniqueidentifier**: bootstrap's `convert_column_value` has no `ColumnType::Guid` arm, so it falls through to the `_` default (`try_get::<&str>` → fails → `Null`). The shared function has a `Guid` arm returning `String(uuid.to_string())`.
- **varbinary**: bootstrap folds `BigVarBin`/`BigBinary` into the **string** arm (`try_get::<&str>` on binary → `Null`). The shared function has a dedicated binary arm returning `String("0x" + hex)`.
- **datetime/datetime2**: both emit strings, but bootstrap uses `NaiveDateTime::to_string()` (space separator, up to 9 fractional digits) while the shared function uses `format("%Y-%m-%dT%H:%M:%S%.3f")` (`T` separator, exactly 3 fractional digits). Same instant, different serialization.

So a `uniqueidentifier` or `varbinary` column has data on the CDC path but `Null` from the snapshot, and datetime values are string-formatted differently between snapshot and live rows — breaking any Cypher predicate that parses or string-compares them.

Also note several types are lost by **both** paths (consistent, but still data loss): `money`, `date`, `time`, `datetimeoffset` all resolve to `Null` on both bootstrap and CDC.

### Desired behavior

Bootstrap and CDC should produce identical `ElementValue` for the same stored value. The cleanest fix is to **delete the bootstrapper's private `convert_column_value` and call the shared `drasi_mssql_common::types::extract_column_value`**, then extend the shared function to also cover the currently-lost types (`money`, `date`, `time`, `datetimeoffset`) and pick a canonical temporal representation (ideally `LocalDateTime`/`ZonedDateTime` rather than a string).

### Workaround

`CAST` unsupported columns to `NVARCHAR` in a view and bootstrap that, matching formats manually.

### Drasi Version

`drasi-core` repo, branch `main`. `drasi-bootstrap-mssql` v0.2.10, `drasi-source-mssql` v0.2.7, `drasi-mssql-common` v0.2.8.

### Operating system

Reproduced on macOS (arm64) with SQL Server (azure-sql-edge / mssql) via Docker.

### Additional context

Relevant code: `components/bootstrappers/mssql/src/mssql.rs` (`convert_column_value`), `components/mssql-common/src/types.rs` (`extract_column_value`), `components/sources/mssql/src/types.rs` (`extract_properties_from_cdc_row`). Related: DECIMAL bootstrap loss (separate issue), ARM-disabled type-mapping test (separate issue).

Contrast with Postgres: MSSQL does **not** drop whole rows and NULL-property presence is consistent, so this is less severe than the Postgres CDC issues — but the bootstrap↔CDC divergence is the same class of bug.

### LINKED ISSUE #674: MSSQL bootstrap: DECIMAL/NUMERIC columns silently snapshotted as Null (CDC preserves them)
### Steps to reproduce

1. Create a SQL Server table with a `DECIMAL`/`NUMERIC` column and enable CDC on it:
   ```sql
   CREATE TABLE dbo.T (Id INT PRIMARY KEY, Price DECIMAL(12,4));
   INSERT INTO dbo.T VALUES (1, 123.4500);
   ```
2. Configure the Drasi MSSQL source with the MSSQL bootstrap provider and subscribe with bootstrap enabled.
3. Compare the `Price` property produced by the **bootstrap snapshot** vs. a **live CDC** change to the same row.

Verified end-to-end against SQL Server (values obtained from the real `MsSqlBootstrapProvider` snapshot and from the production CDC conversion function `drasi_source_mssql::types::extract_properties_from_cdc_row` run on a real row).

### Observed behavior

`DECIMAL`/`NUMERIC` columns are delivered as `ElementValue::Null` by the **bootstrap** path, while the **CDC** path correctly delivers them as `Float`:

| Column | Bootstrap | CDC |
|---|---|---|
| `DECIMAL(12,4)` = `123.45` | `Null` | `Float(123.45)` |
| `DECIMAL(12,0)` = `4200` | `Null` | `Float(4200)` |

Root cause is in `components/bootstrappers/mssql/src/mssql.rs::convert_column_value`. `ColumnType::Numericn | ColumnType::Decimaln` are grouped with the float arm, which only tries:

```rust
if let Ok(Some(val)) = row.try_get::<f32, _>(idx) { ... }
else if let Ok(Some(val)) = row.try_get::<f64, _>(idx) { ... }
else { Ok(ElementValue::Null) }
```

tiberius returns a decimal column as `rust_decimal::Decimal`, not `f32`/`f64`, so both `try_get`s fail and the value silently becomes `Null`. The CDC path uses the **shared** `drasi_mssql_common::types::extract_column_value`, which has a dedicated arm: `row.try_get::<rust_decimal::Decimal, _>(idx)` → `Float`.

Net effect: a `DECIMAL` column (very common for prices, quantities, amounts) is **absent/Null after the initial bootstrap snapshot but present after subsequent live changes** to the same row — an inconsistency that also loses data.

### Desired behavior

Bootstrap should produce the same value the CDC path does for `DECIMAL`/`NUMERIC` (`Float`, or a shared canonical mapping). The cleanest fix is to make the bootstrapper reuse the shared `drasi_mssql_common::types::extract_column_value` instead of its own `convert_column_value` (this also fixes the guid/varbinary/datetime issues tracked separately).

### Workaround

`CAST`/`CONVERT` the decimal column to `FLOAT` or `VARCHAR` in a view and bootstrap that, at the cost of precision/typing.

### Drasi Version

`drasi-core` repo, branch `main`. `drasi-bootstrap-mssql` v0.2.10, `drasi-source-mssql` v0.2.7, `drasi-mssql-common` v0.2.8.

### Operating system

Reproduced on macOS (arm64) with SQL Server (azure-sql-edge / mssql) via Docker.

### Additional context

Relevant code: `components/bootstrappers/mssql/src/mssql.rs` (`convert_column_value`, float/decimal arm) and `components/mssql-common/src/types.rs` (`extract_column_value`, `Numericn | Decimaln` arm). Related: guid/varbinary/datetime bootstrap↔CDC inconsistency, and the ARM-disabled type-mapping test (filed separately).

danielgerlag:
# Description

MSSQL bootstrap and CDC were using different column converters, so the same stored row could produce different `ElementValue`s after snapshot vs live change. That caused `DECIMAL`, `UNIQUEIDENTIFIER`, and `VARBINARY` to appear as `Null` on bootstrap while CDC preserved them, and datetime values were serialized differently on each path.

This change makes bootstrap call the shared `drasi_mssql_common::types::extract_column_value` used by CDC, and extends that shared converter so both paths emit the same canonical mapping:

- `decimal`/`numeric` -> `Float` (via `rust_decimal`)
- `uniqueidentifier` -> `String`
- `varbinary`/`binary` -> `String("0x" + hex)`
- `datetime`/`datetime2`/`smalldatetime` -> `LocalDateTime` (breaking vs previous string formatting)
- `datetimeoffset` -> `ZonedDateTime`
- `date`/`time` -> `String` (same convention as Postgres; `ElementValue` has no bare date/time variants)
- `money`/`smallmoney` -> `Float`

Integration coverage in `TypesTest` now seeds a bootstrap row and updates it via CDC to assert the same property shapes/values on both paths.

**Note for reviewers:** switching datetime types from plain strings to `LocalDateTime`/`ZonedDateTime` is intentionally breaking for Cypher that string-compares those properties. It aligns MSSQL with Postgres and with schema `PropertyType::Timestamp`.

## Type of change

Fixes: #673
Fixes: #674


components/mssql-common/src/types.rs
```diff
@@ -70,6 +75,16 @@ pub fn extract_column_value(row: &Row, col_idx: usize) -> Result<ElementValue> {
                 Ok(ElementValue::Null)
             }
         }
+        // money/smallmoney are decoded by tiberius as f64/f32.
+        ColumnType::Money | ColumnType::Money4 => {
+            if let Ok(Some(val)) = row.try_get::<f64, _>(col_idx) {
```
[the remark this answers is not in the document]
danielgerlag: Investigated against tiberius 0.12.3 source: `money.rs` decodes both 4-byte and 8-byte money as `ColumnData::F64` (scaled int / 1e4), and `FromSql for f64` matches that. `FromSql for rust_decimal::Decimal` only matches `ColumnData::Numeric` (decimal/numeric columns), so using `Decimal` here always yields `Null`.

Verified with the integration test: Decimal path failed (`money_val: expected number, got Some(Null)`); f64 path passes for both `MONEY` and `SMALLMONEY`.

Comment on the arm updated to document the tiberius wire mapping. Leaving `try_get::<f64>`.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -939,6 +907,166 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+    }
+    if let Some(s) = value.as_str() {
+        // Some reaction paths stringify temporal values.
+        anyhow::ensure!(
+            s.starts_with("2024-06-15") && s.contains("10:30:45"),
+            "{path}.{field}: unexpected datetime string: {s}"
+        );
+        return Ok(());
+    }
+    anyhow::bail!("{path}.{field}: expected LocalDateTime, got: {value}")
```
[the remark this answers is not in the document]
danielgerlag: Agreed that envelope would be ideal, but this test observes results via `ApplicationReaction`, which converts through `VariableValue`. That path always stringifies temporals:

```rust
VariableValue::LocalDateTime(t) => Value::String(t.to_string()),
```

So the `__drasi_v1_type__` envelope from direct `ElementValue` serde is never visible here.

Tightened the assertion instead to require chrono Display format (`YYYY-MM-DD HH:MM:SS…` with a space) and reject the old CDC ISO-with-`T` format (`%Y-%m-%dT%H:%M:%S%.3f`). That still catches a regression to the previous CDC string mapping while matching what ApplicationReaction actually emits.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -939,6 +907,166 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+            .get("value")
+            .and_then(|v| v.as_str())
+            .with_context(|| format!("{path}.{field}: missing value in envelope"))?;
+        anyhow::ensure!(
+            raw.starts_with("2024-06-15") && raw.contains("10:30:45"),
+            "{path}.{field}: unexpected LocalDateTime value: {raw}"
+        );
+        return Ok(());
+    }
+    if let Some(s) = value.as_str() {
```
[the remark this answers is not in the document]
danielgerlag: Removed the dual envelope/string acceptance. ApplicationReaction always stringifies `LocalDateTime`/`ZonedDateTime` via `VariableValue` (see `core/src/evaluation/variable_value/mod.rs`), so the helpers now require that Display string and reject the old ISO-with-`T` CDC format. Documented in the helper comments.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -669,8 +669,16 @@ async fn prepare_types_database(config: &MssqlConfig) -> Result<MssqlConfig> {
...
             DecimalVal DECIMAL(10,2) NOT NULL,
+            MoneyVal MONEY NOT NULL,
             VarcharVal VARCHAR(100) NOT NULL,
-            NVarcharVal NVARCHAR(100) NOT NULL
+            NVarcharVal NVARCHAR(100) NOT NULL,
+            GuidVal UNIQUEIDENTIFIER NOT NULL,
+            BinaryVal VARBINARY(16) NOT NULL,
+            DateTimeVal DATETIME NOT NULL,
+            DateTime2Val DATETIME2(7) NOT NULL,
+            DateVal DATE NOT NULL,
```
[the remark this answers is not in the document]
danielgerlag: Done — added `SmallDateTimeVal SMALLDATETIME` to the fixture (seeded at minute precision `10:30:00`) and assert it on both bootstrap and CDC paths.

components/mssql-common/src/types.rs
```diff
@@ -70,6 +75,16 @@ pub fn extract_column_value(row: &Row, col_idx: usize) -> Result<ElementValue> {
...
+                Ok(ElementValue::Float(ordered_float::OrderedFloat(val)))
+            } else if let Ok(Some(val)) = row.try_get::<f32, _>(col_idx) {
+                Ok(ElementValue::Float(ordered_float::OrderedFloat(val as f64)))
+            } else {
```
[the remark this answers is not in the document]
danielgerlag: Done — added `SmallMoneyVal SMALLMONEY` (seed `9.99`) and assert it on both paths. Note tiberius decodes both MONEY and SMALLMONEY as `ColumnData::F64`, so there is a single f64 branch rather than separate f32/f64 paths.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -933,12 +900,172 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+}
+
+fn assert_zoned_datetime_field(data: &Value, field: &str, path: &str) -> Result<()> {
+    let value = data
+        .get(field)
+        .with_context(|| format!("{path}.{field}: missing field"))?;
+    if let Some(obj) = value.as_object() {
+        let type_tag = obj
+            .get("__drasi_v1_type__")
+            .and_then(|v| v.as_str())
```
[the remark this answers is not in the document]
danielgerlag: Resolved by removing the dual-path helper: ApplicationReaction only ever produces the string form for temporals, so that is now the sole asserted shape (with a regression guard against old ISO-with-`T` strings).

components/sources/mssql/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 <!-- generated by git-cliff -->
 
+## [0.2.8] - 2026-08-03
+
+### Bug Fixes
+
+- **mssql:** CDC datetime values are now `LocalDateTime`; datetimeoffset/money/date/time supported via shared converter ([#673](https://github.com/drasi-project/drasi-core/issues/673)) ([#674](https://github.com/drasi-project/drasi-core/issues/674))
```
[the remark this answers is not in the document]
danielgerlag: Done — changelog now marks the datetime `String` → `LocalDateTime` change as **BREAKING**.

components/bootstrappers/mssql/src/mssql.rs
```diff
@@ -406,14 +406,10 @@ impl MsSqlBootstrapHandler {
...
-        // Process each column
+        // Process each column via the shared converter (same path as CDC).
         for (idx, column) in row.columns().iter().enumerate() {
             let column_name = column.name();
-
-            // Convert the value
-            let element_value = self.convert_column_value(row, idx, column)?;
-
-            // Add to properties
+            let element_value = extract_column_value(row, idx)?;
```
amansinghoriginal, line 412: Now that every bootstrap column runs through the shared converter, I think this also picks up the decimal arm's panic.

I checked against main with a live SQL Server: `DECIMAL(38,0)` holding 1e30 panics with "Number exceeds maximum value that can be represented", and `DECIMAL(38,30)` panics with "Scale exceeds the maximum precision allowed: 30 > 28". tiberius calls `Decimal::from_i128_with_scale`, which panics rather than returning Err, so the `if let Ok(..)` cannot catch it and the bootstrap task dies. CDC already hits this so it is not new, but bootstrap used to return Null and finish.

Would reading `tiberius::numeric::Numeric` instead work here? It would also keep decimal primary keys exact, since `generate_element_id` stringifies this value and f64 merges keys that differ past about 16 digits.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -809,120 +848,56 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+        let mut cdc_for_assert = cdc_data.clone();
+        if let Some(obj) = cdc_for_assert.as_object_mut() {
+            obj.insert(
+                "varchar_val".to_string(),
+                Value::String("hello".to_string()),
             );
-        } else {
-            anyhow::bail!("Expected ADD change");
         }
+        assert_types_test_row(&cdc_for_assert, "cdc")?;
```
amansinghoriginal, line 905: You build `cdc_for_assert` just above so it lines up with the bootstrap payload. Any reason not to compare them directly here?

An `assert_eq!(bootstrap_data, cdc_for_assert)` would pin the actual invariant this PR is about. Right now each side is checked against the same loose oracle, so if the two paths drift apart again the test still passes.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -933,12 +908,176 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+        format!(
+            "{path}.{field}: expected LocalDateTime Display string from ApplicationReaction, got: {value}"
+        )
+    })?;
+    // Reject old CDC ISO-with-T format and bare non-temporal values.
+    anyhow::ensure!(
+        !raw.contains('T'),
+        "{path}.{field}: got ISO-with-T string (old CDC String mapping), expected LocalDateTime Display: {raw}"
+    );
+    anyhow::ensure!(
```
amansinghoriginal, line 1056: The fixture seeds `DATETIME2(7)` with `.1234567`, but this checks the date prefix plus a whole second substring, so a change that truncates back to milliseconds would still pass.

Since the values are deterministic, could this assert the exact string `2024-06-15 10:30:45.123456700`? Related: the `!contains('T')` guard above does not really separate the new mapping from the old one, chrono's Display and the deleted converter's `to_string()` produce identical bytes.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -933,12 +908,176 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+    assert_local_datetime_field(data, "smalldatetime_val", path, "10:30:00")?;
+
+    anyhow::ensure!(
+        str_field("date_val")? == "2024-06-15",
+        "{path}.date_val mismatch"
+    );
+
+    let time_val = str_field("time_val")?;
+    anyhow::ensure!(
+        time_val.starts_with("10:30:45"),
```
amansinghoriginal, line 1023: Same thought, `starts_with("10:30:45")` passes even if the whole fractional part disappears. Worth pinning `10:30:45.123456700`?

components/sources/mssql/tests/integration_test.rs
```diff
@@ -933,12 +908,176 @@ async fn test_mssql_column_type_mapping() -> Result<()> {
...
+            "{path}.{field}: expected ZonedDateTime Display string from ApplicationReaction, got: {value}"
+        )
+    })?;
+    anyhow::ensure!(
+        raw.contains("2024-06-15") && raw.contains("10:30:45"),
+        "{path}.{field}: unexpected ZonedDateTime Display value: {raw}"
+    );
+    // Offset should be present (+05:00 / +0500 depending on chrono Display).
+    anyhow::ensure!(
+        raw.contains('+') || raw.contains('Z'),
```
amansinghoriginal, line 1080: This checks an offset exists but not what it is, so a regression that zeroed it to `+00:00` would pass even though the seed is `+05:00`. Could it assert the full expected string?

components/mssql-common/Cargo.toml
```diff
@@ -14,7 +14,7 @@
 
 [package]
 name = "drasi-mssql-common"
-version = "0.2.8"
+version = "0.2.9"
```
amansinghoriginal, line 17: Will hand bumping these fight release-plz? My reading is it still generates its own section from the commits, so after merge there would be two `0.2.9` headings, one calling this a routine bug fix and one calling it BREAKING.

Would it be cleaner to drop the manual bumps and changelog edits and let the release PR produce them, with the breaking signal in the commit message instead?

components/mssql-common/src/types.rs
```diff
@@ -92,14 +105,32 @@ pub fn extract_column_value(row: &Row, col_idx: usize) -> Result<ElementValue> {
...
+        }
+        ColumnType::DatetimeOffsetn => {
+            if let Ok(Some(val)) = row.try_get::<DateTime<FixedOffset>, _>(col_idx) {
+                Ok(ElementValue::ZonedDateTime(val))
+            } else {
+                Ok(ElementValue::Null)
+            }
+        }
+        // ElementValue has no bare Date variant; match Postgres and emit a string.
+        ColumnType::Daten => {
```
amansinghoriginal, line 123: Fair given `ElementValue` has no Date variant, so this is a question rather than a request.

With datetime2 becoming a real temporal and date staying a string, do sibling columns end up behaving differently in one query? `WHERE o.created_at > datetime(...)` hits a real comparison arm, but `WHERE o.ship_date > date(...)` looks like it falls through to false.

Happy for this to be a follow up. Just flagging it because core already has `(Date, Date)` ordering, so adding the variants looks like the only missing piece.

components/mssql-common/src/types.rs
```diff
@@ -70,6 +75,14 @@ pub fn extract_column_value(row: &Row, col_idx: usize) -> Result<ElementValue> {
                 Ok(ElementValue::Null)
             }
         }
+        // MONEY/SMALLMONEY decode as ColumnData::F64 in tiberius (scaled int / 1e4).
+        ColumnType::Money | ColumnType::Money4 => {
```
amansinghoriginal, line 79: Not asking for a change, tiberius only gives back an f64 here so I do not think there is another option. Might be worth extending the comment to note money is approximate above roughly 1e11, since this arm and the DECIMAL arm can land 1 ulp apart on the same stored value.

components/sources/mssql/tests/integration_test.rs
```diff
@@ -726,13 +736,12 @@ async fn prepare_types_database(config: &MssqlConfig) -> Result<MssqlConfig> {
...
 #[tokio::test]
 #[ignore]
-#[cfg(not(target_arch = "aarch64"))]
 async fn test_mssql_column_type_mapping() -> Result<()> {
     let _ = env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
         .is_test(true)
         .try_init();
 
-    let result = tokio::time::timeout(Duration::from_secs(300), async {
+    let result = tokio::time::timeout(Duration::from_secs(900), async {
```
amansinghoriginal, line 744: Good to see this one running on Apple Silicon. Since the helper moved everything to real SQL Server 2022, do the other 11 `cfg(not(target_arch = "aarch64"))` gates still need to be there?

If they come off they would likely need this 900s budget too, the rest are still on 300s and container startup alone can eat most of that under emulation.

amansinghoriginal (changes requested):
Nice cleanup, having bootstrap and CDC go through one converter is clearly the right shape.

I left some inline questions that need addressing before merging. Some are nits/optional.

components/bootstrappers/mssql/src/mssql.rs
danielgerlag: Fair catch, but out of scope for this PR.

CDC already used `extract_column_value` with the `rust_decimal` path, so the panic on extreme `DECIMAL(38,*)` values is pre-existing for live updates. Bootstrap now shares that path, which is intentional for parity.

Hardening decimal decode (e.g. `tiberius::Numeric` instead of `Decimal::from_i128_with_scale`) and keeping high-precision decimal PKs exact are real follow-ups, not part of the bootstrap/CDC consistency fix. Happy to track separately.

components/sources/mssql/tests/integration_test.rs
danielgerlag: Good call — added `assert_eq!(bootstrap_data, cdc_for_assert)` so the bootstrap ↔ CDC invariant is pinned directly, in addition to the shared oracle checks.

components/sources/mssql/tests/integration_test.rs
danielgerlag: Declining for this PR.

The assert is intentionally checking the ApplicationReaction `VariableValue` Display form and rejecting the old CDC ISO-with-`T` format. Fractional-second padding from SQL Server/tiberius/chrono can vary slightly (e.g. DATETIME rounding), so pinning an exact `.123456700` string is brittle for the invariant we care about here.

Bootstrap ↔ CDC equality is now asserted directly above, which is the stronger parity check.

components/sources/mssql/tests/integration_test.rs
danielgerlag: Declining — same rationale as the datetime2 comment. Prefix check is enough for type/shape; bootstrap ↔ CDC `assert_eq!` covers drift between paths.

components/sources/mssql/tests/integration_test.rs
danielgerlag: Declining for this PR. Offset presence is checked; full-string pinning is brittle across chrono Display variants. Bootstrap ↔ CDC equality now covers path drift.

components/mssql-common/Cargo.toml
danielgerlag: Out of scope / declining.

Manual plugin version + changelog bumps match how other component PRs in this repo have been landing. If release-plz needs a different convention, that should be a repo-wide process change, not something unique to this fix.

components/mssql-common/src/types.rs
danielgerlag: Agreed this is a follow-up, not a request for this PR.

`date`/`time` → `String` matches the existing Postgres source convention because `ElementValue` has no bare `Date`/`Time` variants. Adding those variants (and comparison arms end-to-end) is a core model change beyond bootstrap/CDC unification.

components/mssql-common/src/types.rs
danielgerlag: Acknowledged — no code change. tiberius only exposes f64 here, so the mapping stands. Optional comment polish not needed for merge.

components/sources/mssql/tests/integration_test.rs
danielgerlag: Out of scope for this PR.

Only the type-mapping test needed the amd64 helper + longer timeout for this change. Dropping the other `aarch64` gates and retuning all timeouts is a separate harness cleanup.

components/bootstrappers/mssql/src/mssql.rs
amansinghoriginal: Agreed on scope, the panic is pre-existing on the CDC path so it is not something this PR introduced.

I have opened #720 to track it, covering both the panicking decode and the f64 narrowing that collides decimal primary keys. Nothing needed from you here.

components/sources/mssql/tests/integration_test.rs
amansinghoriginal: That makes sense on the DATETIME rounding, and I agree pinning `datetime_val` would be brittle given the 1/300 second ticks land on `.123333333`.

The part I am less sure about is `assert_eq!` covering this. Both sides of that comparison now come from the same `extract_column_value`, so if that function ever truncates, both sides truncate together and the assert still passes. It is a good check for the two paths drifting apart, which is what this PR fixes, but it does not look like it catches both paths moving in step.

`DATETIME2(7)` and `TIME(7)` do seem exact rather than rounded, tiberius computes `increments * 10^(9-scale)`, and I measured `2024-06-15 10:30:45.123456700` consistently. Would pinning just those two be reasonable, leaving `datetime_val` on the prefix check where your rounding point applies?

Not blocking from my side either way.

components/sources/mssql/tests/integration_test.rs
amansinghoriginal: Same thought as the datetime2 thread, `TIME(7)` looks exact too. Not blocking.

components/sources/mssql/tests/integration_test.rs
amansinghoriginal: Same point as the datetime2 thread. One addition: this value goes through `to_rfc3339()` rather than chrono Display, so the format looks fixed rather than variable. If full string pinning still feels brittle, would asserting just the `+05:00` offset work? Not blocking.

components/mssql-common/src/types.rs
amansinghoriginal: Agreed, thanks.

Could you open an issue for the `ElementValue::Date` and `LocalTime` variants? Core already has `(Date, Date)` ordering arms, so the model gap looks like the only missing piece, and it seems worth writing down while it is still cheap to change.

components/sources/mssql/tests/integration_test.rs
amansinghoriginal: Makes sense.

Could you open an issue for the harness cleanup so the remaining gates are not forgotten? Mostly so the next person on Apple Silicon knows why 11 of the 12 tests still compile out.

amansinghoriginal (approved):
Thanks for addressing the feedback. The remaining items are non-blocking or tracked separately. Approving.

ruokun-niu (approved):

PR #701, opened 2026-08-03
Fix order-dependent cross-source bootstrap

### LINKED ISSUE #682: Cross-source query bootstrap is order-dependent: a source listed second bootstraps 0 elements (its rows missing until CDC)
## Summary

For a continuous query that subscribes to **two sources** (a cross-source join), whether each source's rows are **bootstrapped** depends on the **order the sources are listed** in the query. When one source is listed second, its bootstrap runs with `0 node labels` and sends `0 elements`, so that source's existing rows are **missing from the query result** until they are later changed (and arrive via CDC). Listing that source first bootstraps it correctly.

Source order should not affect bootstrap completeness — every subscribed source should bootstrap the node labels it owns.

## Environment

Reproduced on both:
- `@drasi/lib` `0.2.0` (npm prebuilt), and
- built from source at `drasi-lib 0.8.9` / `drasi-core 0.5.7` (drasi-nodejs main).

Sources: `source/postgres` + `source/mysql` (+ their bootstrap providers), from `ghcr.io/drasi-project` (nightly-test).

## Minimal reproduction

Two sources, one node label each:
- `retail-ops` (PostgreSQL) → `orders`
- `physical-ops` (MySQL) → `vehicles`

Join query:

```cypher
MATCH (o:orders)-[:PICKUP_BY]->(v:vehicles)
WHERE o.status <> 'ready' AND v.location = 'Curbside'
RETURN o.id AS id
```

with synthetic join `PICKUP_BY` on `vehicles.plate == orders.plate`.

Seed: 3 orders (`preparing`) in Postgres, 3 vehicles (`Parking`) in MySQL. Start the engine, then move one vehicle to `Curbside` (its order stays `preparing`). Expect the row to appear.

**Case A — sources listed `['physical-ops', 'retail-ops']` (MySQL first): BROKEN**

```
[postgres-bootstrap] Starting PostgreSQL bootstrap for query '…' with 0 node labels and 0 relation labels
[mysql-bootstrap]     Bootstrapped 3 rows from table vehicles
[postgres-bootstrap] Bootstrap completed: 0 total elements sent
```

`orders` are never bootstrapped → the join has no orders → moving the vehicle produces nothing (the query result stays empty). The order only appears if it is separately changed (CDC).

**Case B — sources listed `['retail-ops', 'physical-ops']` (PostgreSQL first): WORKS**

```
[postgres-bootstrap] Starting PostgreSQL bootstrap for query '…' with 2 node labels and 0 relation labels
[postgres-bootstrap] Bootstrapped 3 rows from table 'orders'
[mysql-bootstrap]     Bootstrapped 3 rows from table vehicles
[postgres-bootstrap] Bootstrap completed: 3 total elements sent
```

Both sources bootstrap; moving the vehicle correctly emits the joined row. The same query with `drasi.trueFor(v.location='Curbside', duration({seconds: 5}))` also fires exactly 5s after the vehicle arrives — but only in Case B.

## Expected

Bootstrap completeness is independent of the order sources are listed; each subscribed source bootstraps the labels it provides. In Case A, the PostgreSQL source should bootstrap `orders` (it should receive its owned label, not `0 node labels`).

## Actual

Only the **first**-listed source (and any source that happens to own a label routed to it) bootstraps; the label routing assigns node labels to the first source and the later source(s) receive `0` for the labels they actually own. A cross-source (multi-database) join therefore silently starts with one side missing from bootstrap unless the sources happen to be ordered favorably.

## Notes

- Discovered while porting the Curbside Pickup tutorial to `@drasi/lib` (PostgreSQL orders + MySQL vehicles). Single-source queries, CDC change propagation, the cross-source join itself, `drasi.changeDateTime`/`listMax`, and single-source `drasi.trueFor` all work correctly; only multi-source **bootstrap** routing is order-dependent.

danielgerlag:
Cross-source queries could bootstrap incomplete data because unmapped query labels were assigned only to the first configured source. This made results depend on source ordering and left missing rows unavailable until CDC supplied them.

This change broadcasts unmapped node and physical relationship labels to every subscribed source, allowing each provider to filter the labels it owns. Explicit label ownership remains exclusive, duplicate explicit mappings remain errors, and synthetic join relationships are excluded from physical-source requests.

Regression coverage verifies both source permutations produce equivalent subscription settings and preserves mixed explicit/unmapped behavior. PostgreSQL and MSSQL empty-label behavior for fully unlabeled queries remains out of scope.

**Compatibility note:** Sources that previously received an empty label request may have interpreted it as "bootstrap everything." They now receive the query's labels and filter by label name, so configurations where query labels differ from provider table or collection names must explicitly align those names rather than relying on the previous empty-request fallback.

Closes #682


lib/src/queries/subscription_builder.rs
```diff
@@ -138,12 +137,14 @@ impl SubscriptionSettingsBuilder {
...
-                    } else {
+                    if settings_vec.is_empty() {
                         bail!("No sources configured for query");
                     }
+
+                    // Sources filter requested labels against the data they own.
+                    for settings in settings_vec.iter_mut() {
+                        settings.relations.insert(relation_label.clone());
+                    }
                 }
```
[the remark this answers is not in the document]
danielgerlag: Updated relation allocation so synthetic joins are removed from initialized source settings even when explicitly configured, while duplicate explicit ownership still errors. Added a regression test for the configured-join case.

lib/src/queries/subscription_builder.rs
```diff
@@ -72,15 +72,14 @@ impl SubscriptionSettingsBuilder {
...
-                    // (first) source, which is the common single-source case.
-                    if let Some(first_settings) = settings_vec.first_mut() {
-                        first_settings.nodes.insert(node_label.clone());
-                    } else {
+                    if settings_vec.is_empty() {
                         bail!("No sources configured for query");
                     }
+
+                    // Sources filter requested labels against the data they own.
+                    for settings in settings_vec.iter_mut() {
```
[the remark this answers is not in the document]
danielgerlag: Documented the broadcast/filtering contract on both `SourceSubscriptionSettings` label fields and `Source::subscribe()`, including that explicit mappings remain exclusive.

lib/src/queries/subscription_builder.rs
```diff
@@ -395,6 +405,63 @@ mod tests {
...
+                .iter()
+                .find(|settings| settings.source_id == source_id)
+                .unwrap();
+
+            assert_eq!(forward_settings.nodes, reverse_settings.nodes);
+            assert!(forward_settings.nodes.contains("orders"));
+            assert!(forward_settings.nodes.contains("vehicles"));
+            assert!(forward_settings.relations.is_empty());
+            assert_eq!(forward_settings.relations, reverse_settings.relations);
+        }
```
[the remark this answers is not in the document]
danielgerlag: Expanded the source-order regression to include the unmapped physical `CONTAINS` relation. It now verifies equal non-empty relation sets in both orders while the synthetic `PICKUP_BY` relation remains excluded.

lib/src/queries/subscription_builder.rs
```diff
@@ -72,15 +72,14 @@ impl SubscriptionSettingsBuilder {
...
-                    // This is intentional: labels that aren't explicitly mapped in
-                    // source configuration are assumed to belong to the primary
-                    // (first) source, which is the common single-source case.
-                    if let Some(first_settings) = settings_vec.first_mut() {
-                        first_settings.nodes.insert(node_label.clone());
-                    } else {
+                    if settings_vec.is_empty() {
                         bail!("No sources configured for query");
                     }
+
```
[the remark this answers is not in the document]
danielgerlag: Updated the comment to describe the broadcast performed here and the downstream filtering responsibility.

lib/src/queries/subscription_builder.rs
```diff
@@ -138,12 +137,14 @@ impl SubscriptionSettingsBuilder {
                         }
                     }
 
-                    // Not a join relation - add to first source (default)
-                    if let Some(first_settings) = settings_vec.first_mut() {
-                        first_settings.relations.insert(relation_label.clone());
-                    } else {
+                    if settings_vec.is_empty() {
                         bail!("No sources configured for query");
```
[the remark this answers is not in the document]
danielgerlag: Updated the relation-path comment with the same broadcast and downstream filtering wording.

ruokun-niu (approved):

lib/src/sources/traits.rs
```diff
@@ -240,7 +240,9 @@ pub trait Source: Send + Sync {
     ///
     /// This is called by queries to receive data changes from this source.
     /// The source should return a receiver for streaming events and optionally
-    /// a bootstrap receiver for initial data.
+    /// a bootstrap receiver for initial data. Subscription labels without an
```
amansinghoriginal, line 243: Could this note also go on BootstrapRequest.node_labels in lib/src/bootstrap/mod.rs? That struct is what providers read, and FFI plugin authors never implement Source::subscribe. components/bootstrappers/README.md:144 also still says "(from this source)", which now reads as the opposite of this.

lib/src/queries/subscription_builder.rs
```diff
@@ -72,15 +72,14 @@ impl SubscriptionSettingsBuilder {
...
-                    // source configuration are assumed to belong to the primary
-                    // (first) source, which is the common single-source case.
-                    if let Some(first_settings) = settings_vec.first_mut() {
-                        first_settings.nodes.insert(node_label.clone());
-                    } else {
+                    if settings_vec.is_empty() {
                         bail!("No sources configured for query");
                     }
+
+                    // Broadcast to all sources; each source filters out labels it does not own.
```
amansinghoriginal, line 79: This broadcasts to sources that declared an explicit nodes list too, so their config no longer bounds what they contribute. It also sits a bit oddly next to the arm below, where the same label in two configs is an error.

Would skipping sources with a non-empty mapping work? #682 declares no mappings, so I think it would still be fixed. May well be missing a reason you did it this way.

lib/src/queries/subscription_builder.rs
```diff
@@ -114,46 +113,41 @@ impl SubscriptionSettingsBuilder {
...
+                        );
                     }
                 }
-                1 => {
-                    // Found in exactly one source - already in the HashSet from initialization
-                    // Nothing to do, it's already there
+
+                // Synthetic join relations must never be requested from physical sources.
+                for settings in settings_vec.iter_mut() {
+                    settings.relations.remove(relation_label);
```
amansinghoriginal: If a source explicitly configured this relation, it gets dropped here with no diagnostic (it was kept before). Excluding synthetic joins looks right to me, but maybe a warn! so the misconfig is visible?

lib/src/config/schema.rs
```diff
@@ -167,7 +167,13 @@ pub struct SourceSubscriptionSettings {
     pub source_id: String,
     pub enable_bootstrap: bool,
     pub query_id: String,
+    /// Node labels the query is interested in. Explicit mappings are exclusive to their
```
amansinghoriginal, line 170: Worth a line in the PR description? Later sources used to receive an empty list, and mysql/oracle/sqlite/neo4j read empty as "bootstrap everything". That is why Case B in #682 worked. They now filter by label name, so a config where the query label differs from the table name silently drops to 0 rows.

lib/src/queries/subscription_builder.rs
```diff
@@ -395,6 +398,100 @@ mod tests {
...
+
+        let settings =
+            SubscriptionSettingsBuilder::build_subscription_settings(&query_config, &query_labels)
+                .unwrap();
+
+        assert!(!settings[0].relations.contains("CUSTOMER"));
+    }
+
+    #[test]
+    fn test_cross_source_join_settings_are_order_independent() {
```
amansinghoriginal, line 442: Nice repro. Any appetite for a two-source case in lib-integration-tests as well? The unit test pins the routing, but the bug people hit was missing bootstrap rows. Non-blocking.

amansinghoriginal (commented):
The fix looks right to me, and the order-independence test pins the exact failure from #682.

Since the change depends on providers ignoring labels they do not own, I read through the bundled bootstrappers (postgres, mysql, mssql, oracle, sqlite, neo4j, http, scriptfile, application, kubernetes, dataverse, platform). All of them warn and skip; none error. settings.nodes is also only read in one place, sources/base.rs, to build the BootstrapRequest, so live CDC routing is unaffected. Locally, 876 drasi-lib unit tests pass and clippy is clean.

Comments below are mostly about documenting the new contract where implementers will see it, plus a couple of questions on the broadcast scope.

lib/src/sources/traits.rs
danielgerlag: Added the broadcast/filtering contract directly to `BootstrapRequest.node_labels` and `relation_labels`, and updated the bootstrap provider guide so FFI and provider implementers see the same semantics.

lib/src/queries/subscription_builder.rs
danielgerlag: I kept the broadcast to all sources intentionally. An explicit mapping claims ownership of that specific label, but does not imply the source cannot own other unmapped labels; skipping every source with any mapping would make mixed explicit/unmapped configurations incomplete. Providers remain responsible for filtering labels they do not own.

lib/src/queries/subscription_builder.rs
danielgerlag: Agreed that silently dropping the explicit mapping hides a configuration mistake. This now returns a clear error when a synthetic join relation is assigned to a physical source, with the regression test updated accordingly.

lib/src/config/schema.rs
danielgerlag: Added a compatibility note to the PR description explaining that sources which previously treated an empty request as "bootstrap everything" now filter the query labels by name, including the table/collection-name alignment implication.

lib/src/queries/subscription_builder.rs
danielgerlag: I kept this at the deterministic unit boundary for now. The existing join fixtures inject events after subscription rather than exposing provider bootstrap requests, so an end-to-end case would require broader test scaffolding without adding coverage of a different decision point.

agentofreality (approved):

Issue #679, opened 2026-07-30
MySQL type-consistency test is #[ignore]d, blocked by the TLS panic, and excludes/omits the divergent types


danielgerlag:
### Area for Improvement

Testing. The MySQL bootstrap↔CDC type-consistency test is opt-in, currently cannot run against a standard MySQL in this environment, and by design skips the column types that actually diverge — so real inconsistencies (ENUM, SET, TIMESTAMP, fractional DATETIME, JSON) are unguarded.

### Observed behavior

`test_type_mapping_consistency_between_bootstrap_and_cdc` in `components/sources/mysql/tests/integration_test.rs`:

- Is annotated `#[ignore]`, so it does not run in the default test/CI pass.
- **Fails to run at all** against a TLS-capable MySQL because the source uses the default `SslMode::IfAvailable`, which panics when `mysql_async` is built without a TLS backend (see the separate TLS-panic issue). I only got the CDC half working by setting `SslMode::Disabled` in a local copy.
- **Explicitly excludes** `ENUM`, `SET`, and `TIMESTAMP` from its comparison (documented in-code: "the binlog encodes them as raw integers/bitfields without schema context"). These are exactly the divergent types.
- **Omits** several types entirely: `JSON`, `BLOB`, `BIT`, `VARBINARY`, and fractional-second `DATETIME(n)`/`TIME(n)`. Empirically, fractional `DATETIME` (µs dropped) and `JSON` (whitespace) also diverge but are not covered.

Empirically, of the columns the test skips or omits, 5 diverge between bootstrap and CDC: ENUM, SET, TIMESTAMP, fractional DATETIME, JSON.

### Desired behavior

- The consistency test should run in CI (or a documented, enforced integration lane), not be silently `#[ignore]`d.
- It should set `SslMode::Disabled` (or the TLS bug should be fixed) so the CDC half actually runs.
- Coverage should include `ENUM`, `SET`, `TIMESTAMP`, `JSON`, `BIT`, `BLOB`, `VARBINARY`, and fractional-second `DATETIME(n)`/`TIME(n)`, with explicit bootstrap-vs-CDC equality assertions per type — so the divergences are caught rather than excluded by comment.

### Proposed Fix

[...]

PR #702, opened 2026-08-03
Fix MySQL CDC TLS panic and add SSL support to the bootstrapper

### LINKED ISSUE #678: MySQL source: CDC binlog task panics under default SslMode::IfAvailable when mysql_async is built without a TLS backend
### Steps to reproduce

1. Build the MySQL source as-is (its `mysql_async` dependency is declared with `features = ["binlog"]` only — no TLS backend).
2. Point it at a MySQL server that advertises TLS (the default for MySQL 5.7+/8.x).
3. Leave `ssl_mode` at its default (`SslMode::IfAvailable`) and start the binlog stream.
4. The binlog task panics instead of connecting.

Reproduced against MySQL 8.1 (testcontainers). The repo's own `test_type_mapping_consistency_between_bootstrap_and_cdc` fails the same way in this environment.

### Observed behavior

The binlog reader task panics inside `mysql_async`:

```
thread '...' panicked at mysql_async-0.37.0/src/io/tls/no_tls.rs:9:9:
Client had asked for TLS connection but TLS support is disabled.
Please enable one of the following features: ["native-tls-tls", "rustls-tls"]
```

Root cause: the default `SslMode::IfAvailable` path in `components/sources/mysql/src/stream.rs` is *written* to try TLS and gracefully fall back to a plaintext connection on error:

```rust
SslMode::IfAvailable => match self
    .connect_binlog_stream_with_ssl(start_position, Some(self.relaxed_ssl_opts()))
    .await
{
    Ok(stream) => Ok(stream),
    Err(ssl_error) => {
        warn!("SSL connection attempt failed ..., retrying without SSL: {ssl_error}");
        self.connect_binlog_stream_with_ssl(start_position, None).await ...
    }
};
```

But `mysql_async` is compiled **without** any TLS backend, so when SSL opts are supplied it **panics** (`no_tls.rs`) rather than returning `Err`. The panic is not catchable by the `match`, so the intended fallback never runs and the binlog task aborts.

Net effect: with default configuration, **MySQL CDC panics against any TLS-capable MySQL server** (i.e. most real deployments). Bootstrap is unaffected because it connects without requesting SSL. The only way to get CDC working today is to explicitly set `SslMode::Disabled`.

### Desired behavior

The default configuration should connect successfully (with or without TLS) against a standard MySQL server, and `SslMode::IfAvailable` should genuinely fall back to plaintext when TLS is unavailable — without panicking. Options:
1. Enable a TLS backend feature on `mysql_async` (`native-tls-tls` or `rustls-tls`) so TLS actually works and `IfAvailable` can negotiate it; and/or
2. When no TLS backend is compiled in, have `IfAvailable` skip the SSL attempt entirely (connect with `ssl_opts = None`) instead of calling into the code path that panics.

### Workaround

Explicitly configure the source with `ssl_mode = Disabled`. This connects in plaintext and avoids the panic (at the cost of no transport encryption).

### Drasi Version

`drasi-core` repo, branch `main`. `drasi-source-mysql` v0.2.5 (`mysql_async` 0.37, `features = ["binlog"]`).

### Operating system

Reproduced on macOS (arm64) with MySQL 8.1 via Docker; the failure is environment-independent for any TLS-advertising MySQL.

### Additional context

Relevant code: `components/sources/mysql/Cargo.toml` (`mysql_async` features), `components/sources/mysql/src/stream.rs` (`connect_binlog_stream`, `connect_binlog_stream_with_ssl`, `relaxed_ssl_opts`), `components/sources/mysql/src/config.rs` (`SslMode`, default `IfAvailable`). This bug also blocks the CDC half of the MySQL type-consistency integration test (filed separately).

ruokun-niu:
# Description

With the default `SslMode::IfAvailable`, the MySQL source's CDC (binlog) task
**panics** against any TLS-advertising MySQL server (i.e. most real deployments):

    thread '...' panicked at mysql_async-0.37.0/src/io/tls/no_tls.rs:9:9:
    Client had asked for TLS connection but TLS support is disabled.
    Please enable one of the following features: ["native-tls-tls", "rustls-tls"]

`mysql_async` is built with `features = ["binlog"]` and **no TLS backend**, so the
`IfAvailable` code path — which tries TLS and is written to fall back to plaintext on
`Err` — instead hits a `panic!` that the `match` cannot catch. The binlog task aborts,
and the only workaround was to explicitly set `SslMode::Disabled`.

This PR makes the default configuration connect successfully (with or without TLS),
centralizes SSL handling in `drasi-mysql-common`, and extends the same SSL support to
the bootstrap provider (which previously always connected in plaintext).

## Root cause

- `components/sources/mysql/src/stream.rs` (`connect_binlog_stream`) attempts a TLS
  connection for `IfAvailable`, expecting a catchable `Err` on failure.
- Without a compiled TLS backend, `mysql_async` **panics** rather than returning `Err`,
  so the intended plaintext fallback never runs.
- Bootstrap was unaffected only because it never requested SSL — but that also meant it
  could not connect to a TLS-requiring server at all.

### `drasi-mysql-common` (single source of truth)

- Moved the `SslMode` enum here so the source and bootstrap share one definition.
- Added `connect_with_ssl_mode(...)`, a panic-guarded connect helper used by both plugins:
  - **`Disabled`** → plaintext.
  - **`IfAvailable`** → with `tls`: try TLS, gracefully fall back to plaintext on error;
    without `tls`: connect plaintext directly (never touches the panicking path).
  - **`Require` / `RequireVerifyCa` / `RequireVerifyFull`** → with `tls`: negotiate TLS;
    without `tls`: return a clear error instead of panicking.
- Added an opt-in `tls` feature backed by **rustls** (`mysql_async/rustls-tls`), chosen
  to avoid a system OpenSSL build dependency (matters for cross-compilation/docs.rs).

### `drasi-source-mysql`

- `SslMode` is now re-exported from `drasi-mysql-common` (public API unchanged).
- `connect_binlog_stream` builds its `Conn` via the shared helper; removed the local
  panic-prone match and SSL-opt helpers.
- Added a `tls` feature that enables the rustls backend.

### `drasi-bootstrap-mysql` (now SSL-aware)

- Added an `ssl_mode` config field, a `with_ssl_mode(...)` builder method (default
  `IfAvailable`), and a `SslModeDto` in the descriptor/schema.
- `connect()` now uses the shared `connect_with_ssl_mode(...)`.
- Added a matching `tls` feature.

## Behavior

- **Default build (no `tls`):** panic-free. `IfAvailable` connects in plaintext;
  `Require*` returns a descriptive error rather than panicking.
- **`--features tls` build:** real TLS negotiation; `IfAvailable` falls back to plaintext
  on error.

## Validation

- `test_type_mapping_consistency_between_bootstrap_and_cdc` (Docker, MySQL 8.x,
  **default features**) — previously panicked, now **passes**.
- `test_mysql_source_change_detection` (Docker) — **passes**.
- New unit test in `drasi-mysql-common` asserts `Require*` returns an error (not a panic)
  when built without a TLS backend.
- `cargo build` and `cargo build --features tls` for all three crates; `cargo clippy
  --all-targets`; unit tests — all green.

## Scope note

This unblocks the TLS blocker called out in #679 (the type-consistency test can now run
under the default `IfAvailable`). The remaining #679 work — un-`#[ignore]`ing the test in
CI and expanding bootstrap↔CDC type coverage (ENUM/SET/TIMESTAMP/JSON/fractional
DATETIME, etc.) via a shared type converter — is intentionally left as a separate
follow-up.

## Type of change

Fixes: #678


components/mysql-common/src/connect.rs
```diff
@@ -0,0 +1,157 @@
...
+#[cfg(feature = "tls")]
+fn relaxed_ssl_opts() -> mysql_async::SslOpts {
+    mysql_async::SslOpts::default()
+        .with_danger_accept_invalid_certs(true)
+        .with_danger_skip_domain_validation(true)
+}
+
+#[cfg(feature = "tls")]
+fn verify_ca_ssl_opts() -> mysql_async::SslOpts {
+    mysql_async::SslOpts::default().with_danger_skip_domain_validation(true)
```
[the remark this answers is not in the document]
ruokun-niu: RequireVerifyCa intentionally skips hostname validation — that's the definition of MySQL --ssl-mode=VERIFY_CA (verify chain, not hostname), which this variant mirrors by name. Making it validate hostname would make it identical to Require/RequireVerifyFull and defeat the purpose of a CA-only mode. The hostname-skip is opt-in and documented on the variant; users wanting full verification have Require/RequireVerifyFull.

components/bootstrappers/mysql/src/descriptor.rs
```diff
@@ -69,15 +70,60 @@ pub struct MySqlBootstrapConfigDto {
...
 }
 
+/// SSL mode DTO (mirrors [`SslMode`]).
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, utoipa::ToSchema)]
+#[schema(as = bootstrap::mysql::SslMode)]
+#[serde(rename_all = "snake_case")]
+#[derive(Default)]
+pub enum SslModeDto {
+    #[default]
+    Disabled,
```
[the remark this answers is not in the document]
ruokun-niu: already fixed

danielgerlag (approved):

components/mysql-common/Cargo.toml
```diff
@@ -26,9 +26,23 @@ categories = ["database"]
 [lints]
 workspace = true
 
+[features]
+# Enables a TLS backend (rustls) for MySQL connections. Off by default so that
+# builds without a system/vendored TLS backend still work; when disabled, the
+# SslMode::Require* variants return an error instead of panicking.
+tls = ["mysql_async/rustls-tls"]
```
amansinghoriginal: This turns on rustls but not a crypto provider. mysql_async pulls rustls with default-features = false, and its rustls-tls feature is only rustls + tokio-rustls + webpki-roots. Its own default-rustls preset adds aws-lc-rs on top for this reason.

So the first real handshake panics:

```
rustls-0.23.43/src/crypto/mod.rs:249: Could not automatically determine the process-level CryptoProvider
```

`cargo tree --manifest-path components/sources/mysql/Cargo.toml --features tls -e features,no-dev` shows no ring and no aws-lc-rs, same for the bootstrapper. A standalone binary with this exact feature set panics against a live MySQL 8.1, and adding ring makes it connect.

Test builds hide this because testcontainers pulls rustls/ring in as a dev-dep.

Suggest `tls = ["mysql_async/rustls-tls", "mysql_async/ring"]` here and in the two plugin manifests.

components/mysql-common/src/connect.rs
```diff
@@ -0,0 +1,204 @@
...
+mod tls_tests {
+    use super::*;
+
+    /// With the `tls` feature, `IfAvailable` first attempts a TLS connection and
+    /// then falls back to a plaintext connection when that fails. The critical
+    /// invariant (the original bug) is that this path must **not panic**. Against
+    /// a non-listening address both the TLS attempt and the plaintext retry fail,
+    /// so the call returns `Err` from the second attempt rather than aborting.
+    #[tokio::test]
+    async fn if_available_falls_back_without_panicking() {
```
amansinghoriginal, line 133: This connects to a closed port, so TCP fails before any handshake and build_tls_connector never runs. It passes even when TLS is entirely broken (see my note on the tls feature in Cargo.toml). Could it point at a real TLS-advertising server instead?

components/mysql-common/src/dto.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use crate::config::SslMode;
+
+/// SSL mode DTO (mirrors [`SslMode`]).
+///
+/// Single source of truth for the `ssl_mode` config field shared by the MySQL
+/// source and bootstrap plugin descriptors.
+#[derive(
```
amansinghoriginal: cargo fmt --check fails here, the derive fits on one line. Two more spots: dto.rs:92 (the two require_verify_full asserts need wrapping) and sources/mysql/src/stream.rs:559 (double blank line left by the removed helpers). The pre-commit hook runs cargo fmt -- --check.

components/sources/mysql/Cargo.toml
```diff
@@ -51,6 +51,10 @@ bytes = "1.0"
 
 [features]
 dynamic-plugin = []
+# Enables a TLS backend (rustls) so SslMode::Require* and IfAvailable can
+# negotiate TLS. Off by default; without it, IfAvailable connects in plaintext
+# and Require* returns an error instead of panicking.
+tls = ["mysql_async/rustls-tls", "drasi-mysql-common/tls"]
```
amansinghoriginal: Worth noting xtask builds plugin cdylibs with `--features dynamic-plugin` only (xtask/src/main.rs:509), so there is no way to switch this on for a packaged plugin. As shipped, IfAvailable is always plaintext and Require* always errors.

Postgres and mssql do not gate TLS at all: sources/postgres depends on postgres-native-tls unconditionally, and tiberius has native-tls in its defaults. Making it unconditional here would fix this and let both cfg pairs in connect.rs go away. Not blocking, but as it stands the feature does not reach users.

components/mysql-common/src/dto.rs
```diff
@@ -0,0 +1,130 @@
...
+    Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize, utoipa::ToSchema,
+)]
+#[schema(as = mysql::SslMode)]
+#[serde(rename_all = "snake_case")]
+pub enum SslModeDto {
+    Disabled,
+    #[default]
```
amansinghoriginal, line 35: Heads up, this flips the source default. The old SslModeDto in sources/mysql/src/descriptor.rs defaulted to Disabled. With #[serde(default)] and ConfigValue::default() being Static(T::default()), a descriptor that omits ssl_mode now means IfAvailable. Inert today since both are plaintext without the tls feature, but it goes live once TLS is on. Probably worth a line in the PR description.

components/mysql-common/src/connect.rs
```diff
@@ -0,0 +1,204 @@
...
+/// [`SslMode::RequireVerifyFull`]) is used while the crate was built without
+/// the `tls` feature.
+pub async fn connect_with_ssl_mode<F>(build_opts: F, mode: SslMode) -> Result<Conn>
+where
+    F: Fn() -> OptsBuilder,
+{
+    match mode {
+        SslMode::Disabled => Ok(Conn::new(build_opts().ssl_opts(None)).await?),
+        SslMode::IfAvailable => connect_if_available(build_opts).await,
+        SslMode::Require => connect_require(build_opts, RequireMode::VerifyFull).await,
```
amansinghoriginal: Require and RequireVerifyFull both map to VerifyFull, so one is redundant. It also reads backwards against MySQL itself, where --ssl-mode=REQUIRED does not verify the cert and VERIFY_CA/VERIFY_IDENTITY do. Someone writing require would get stricter behavior than they expect and a failure on self-signed certs. Nothing depends on this yet, so maybe drop Require or rename to match MySQL?

components/bootstrappers/mysql/src/descriptor.rs
```diff
@@ -69,15 +70,22 @@ pub struct MySqlBootstrapConfigDto {
...
+    #[serde(default)]
+    #[schema(value_type = SslModeDto)]
+    pub ssl_mode: ConfigValue<SslModeDto>,
+
     #[serde(default)]
     #[schema(value_type = Vec<bootstrap::mysql::MySqlTableKeyConfig>)]
     pub table_keys: Vec<MySqlTableKeyConfigDto>,
 }
 
+/// SSL mode DTO (`SslModeDto`) is shared from `drasi-mysql-common`.
```
amansinghoriginal: Nit: this doc comment has no item under it, so it attaches to MySqlBootstrapSchemas. Probably wants to be // or just removed.

amansinghoriginal (changes requested):
Verified the fix works: I reproduced the original panic on the base commit against MySQL 8.1, and test_type_mapping_consistency_between_bootstrap_and_cdc passes here with default features. Moving the shared pieces into drasi-mysql-common and giving the bootstrapper an ssl_mode both look right to me, and propagating Error::Server instead of falling back is a nice touch.

Two things I think should be fixed before merge, plus a few smaller notes, all inline.

Unrelated to this PR: storedproc-mysql still has the same panic at executor.rs:90. I will file that separately.

components/mysql-common/Cargo.toml
ruokun-niu: Good point. Thanks for pointing this out

components/mysql-common/src/connect.rs
ruokun-niu: Updated

components/sources/mysql/Cargo.toml
ruokun-niu: Made TLS unconditional — mysql_async now pulls rustls-tls + ring directly in all three manifests, so packaged cdylibs (--features dynamic-plugin) negotiate TLS and Require* works for users. Dropped the tls feature and all #[cfg(feature = "tls")] branches in connect.rs (both cfg pairs are gone), and removed the now-impossible "no TLS backend" error path + test. cargo tree -e no-dev confirms ring is present with no feature flags.

components/bootstrappers/mysql/src/descriptor.rs
ruokun-niu: fixed

ruokun-niu:
@amansinghoriginal I have addressed your comments. Please review again. Thanks!

components/mysql-common/Cargo.toml
```diff
@@ -26,9 +26,22 @@ categories = ["database"]
...
+api = ["dep:utoipa"]
+
 [dependencies]
 drasi-core.workspace = true
 anyhow = "1.0"
 log = "0.4"
+# A rustls TLS backend (+ ring crypto provider) is always compiled in so every
+# SslMode works in shipped plugins (which are built without extra features).
+# rustls is pure Rust, so this needs no system OpenSSL.
+mysql_async = { version = "0.37", features = ["rustls-tls", "ring"] }
```
amansinghoriginal: Could we add `tls12` here? Production builds currently offer only TLS 1.3. I reproduced `Require` failing with `ProtocolVersion` against MySQL restricted to TLS 1.2.

components/bootstrappers/mysql/src/config.rs
```diff
@@ -39,6 +39,13 @@ pub struct MySqlBootstrapConfig {
     /// Tables to bootstrap
     pub tables: Vec<String>,
 
+    /// SSL/TLS mode for the MySQL connection.
+    ///
+    /// Defaults to [`SslMode::IfAvailable`]. When built without the `tls`
```
amansinghoriginal: Could we update this and the MySQL READMEs? There is no `tls` feature now because rustls is always enabled.

amansinghoriginal (dismissed):
Thanks for working on this fix.

components/mysql-common/Cargo.toml
ruokun-niu: Good point. Reproduced and fixed — added tls12 to all three manifests. Standalone binary against MySQL 8.1 with --tls-version=TLSv1.2: without tls12 → received fatal alert: ProtocolVersion; with it → connects, Ssl_version=TLSv1.2

agentofreality (approved):

Issue #602, opened 2026-06-25
source/mysql 0.2.4 crashes (double-free / heap corruption) on every binlog row event with binlog_row_metadata=FULL


danielgerlag:
### Summary

The published **`source/mysql`** plugin (v0.2.4, the latest version compatible with `drasi-server` 0.2.0 / plugin-sdk 0.9.0) **crashes the entire server process with a native memory error on every binlog row event** when `binlog_row_metadata=FULL` is configured on the MySQL server. Because `FULL` metadata is *required* for the source to identify rows by key column name, and GTID is *required* by the source's `AUTO_POSITION` binlog streaming, there is no working configuration: the source either crashes (`FULL`) or cannot construct row identities (`MINIMAL`).

This makes the MySQL source unusable for live change capture.

### Environment

- `drasi-server` 0.2.0 (rustc 1.88.0, plugin-sdk 0.9.0)
- Plugins (auto-installed from `ghcr.io/drasi-project`): `source/mysql` **0.2.4**, `bootstrap/mysql` 0.2.7
- MySQL server: `mysql:8.0` (Docker), `binlog-format=ROW`, `binlog-row-image=FULL`, `gtid-mode=ON`, `enforce-gtid-consistency=ON`
- Reproduced on **both Windows (x86_64-pc-windows-msvc)** and **Linux (x86_64-unknown-linux-gnu)**.
- Source deps (from `components/sources/mysql/Cargo.toml`): `mysql_async = "0.34"`, `mysql_common = "0.32"`.

### Steps to reproduce

1. Start MySQL 8.0 with ROW binlog, GTID on, and `binlog-row-metadata=FULL`.
2. Create a table with a non-`id` primary key, e.g.:
   ```sql
   CREATE TABLE vehicles (
       plate VARCHAR(10) PRIMARY KEY,
       location VARCHAR(20) NOT NULL DEFAULT 'Parking'
       -- ...other columns
   );
   ```
3. Configure a `source/mysql` with `tables: [vehicles]` and `tableKeys: [{ table: vehicles, keyColumns: [plate] }]`, plus `bootstrap/mysql`. Bootstrap succeeds and loads the seed rows.
4. Make **any** change to the table (INSERT *or* UPDATE), e.g. `UPDATE vehicles SET location='Curbside' WHERE plate='A1234';`.

### Actual behavior

The server process dies the instant the binlog row event is decoded:

- **Linux:** `double free or corruption (fasttop)` (glibc abort).
[...]

PR #704, opened 2026-08-04
docs: standardize agent docs on AGENTS.md with portable CLAUDE.md stubs

amansinghoriginal:
Six AGENTS.md files at the levels that actually need context: root, core, lib, components, shared-tests, examples. Every line was verified against the current code, Makefile and CI configs.

Each one gets a one line CLAUDE.md beside it containing just `@AGENTS.md`. Claude Code only reads CLAUDE.md, and symlinks break on Windows clones (git defaults to core.symlinks=false and checks them out as plain text files), so the import stub is the portable bridge.

The old CLAUDE.md files and the eight lowercase agents.md files are deleted, about 1,000 lines. Anything true in them is either covered by the new files or already documented in doc comments at the point of use. A lot of it was stale, some dangerously so, like the pre #602 opaque pointer FFI description.

Also updates components/plugin-architecture.md to describe the post #602 serialized payload FFI design, and fixes two stale doc comments in lib/src/error.rs and lib/src/lib.rs.


ruokun-niu:
@amansinghoriginal 
Heads up: you can safely ignore the failing `pr-correctness-reviewer` check on this PR.


It's not failing on anything in our code — it dies in a gh-aw pre-agent security step ("Restore agent config folders from base branch") before the reviewer even runs. That step restores agent config files from `main` and deletes any that don't exist there. Because this PR adds `AGENTS.md` (not yet on `main`) and turns `CLAUDE.md` into a symlink → `AGENTS.md`, the step deletes `AGENTS.md`, which leaves `CLAUDE.md` as a dangling symlink, and the following `cp` fails on it (`cp: not writing through dangling symlink`).


**This is a one-time thing.** Once this PR merges, `AGENTS.md` will exist on `main`, so the step restores it instead of deleting it and the symlink never dangles — future PRs won't hit this. `main` has no required checks, so this red mark doesn't block merging.

lib/src/sources/AGENTS.md
ruokun-niu, line 1: **`lib/src/sources/AGENTS.md` — "What crosses FFI" table is out of date.** The `Source` trait grew but the table wasn't updated. It's missing 4 methods that have real `SourceVtable` fields:

| Trait method | Vtable field |
|---|---|
| `properties()` | `properties_fn` |
| `describe_schema()` | `describe_schema_fn` |
| `supports_replay()` | `supports_replay_fn` |
| `remove_position_handle()` | `remove_position_handle_fn` |

Also the `subscribe()` row is thin — the real `subscribe_fn` takes more args (`enable_bootstrap`, `query_id`, `nodes_json`, `resume_from_ptr`, `request_position_handle`, …). Worth adding the missing rows.

lib/src/reactions/AGENTS.md
ruokun-niu: The `ReactionVtable` has **no `enqueue_query_result_fn`** — it's now `start_result_push_fn` (push-callback model), and the table is missing `properties_fn`, `is_durable_fn`, `needs_snapshot_on_fresh_start_fn`, `default_recovery_policy_fn`, `bootstrap_fn`.

amansinghoriginal:
@ruokun-niu Good catches, thanks.
I went further than the specific rows, because I think the tables are the actual problem.

Both tables were written in March and updated on 6 July, and were still wrong a month later: `SourceVtable` was missing 4 fields, `ReactionVtable` was missing 6 and listed one that does not exist. An enumeration of code in markdown is a second copy with no compiler checking it, so it drifts silently and an agent trusts it anyway.

So instead of adding the missing rows I removed the tables and pointed at `vtables.rs` as the authority, keeping the design level rules and the "what to update when you change X" checklists, which do not rot.

Two things worth flagging while I was in there:

`enqueue_query_result_fn` appears nowhere in the code, only in docs. The reactions and channels docs both described result delivery as an opaque `Box::into_raw` / `Box::from_raw` handoff. That is the pattern #602 removed for causing heap corruption, and `channels/AGENTS.md` warned against it at the top of the file while prescribing it further down. Both now describe the push based model and the envelope vs payload distinction.

The reactions doc also said `QueryProvider` had been removed and should not be re-added. It still exists in `lib/src/reactions/traits.rs`, so I dropped that line.

Also thanks for the note on `pr-correctness-reviewer`, that matches what I saw.

danielgerlag (changes requested):
**Requesting changes** because the current symlink strategy is not portable and is already breaking review automation.

1. **`CLAUDE.md` symlinks are unsafe across supported checkout paths.** On checkouts without symlink support, notably Git for Windows with `core.symlinks=false`, each link becomes a regular file containing only `AGENTS.md`, which is not a Claude import. Source archives also do not reliably retain the intended behavior. More immediately, workflow run `31146231058` removed the PR-only `AGENTS.md`, left `CLAUDE.md` dangling, and six reviewer jobs failed with `cp: not writing through dangling symlink`. Please use regular `CLAUDE.md` files containing `@AGENTS.md` instead.

2. **Nested instructions bypass reviewer sanitization.** `.github/workflows/pr-correctness-reviewer.lock.yml:504` restores only root instruction files. A future PR can modify a nested `AGENTS.md`, which Copilot CLI automatically loads while reviewing files under that directory, allowing the PR to steer its own automated review. Please recursively restore or remove all `AGENTS.md` and `CLAUDE.md` files against the base branch before agent execution, then regenerate the locked workflows.

3. **The ABI bump guidance is incorrect.** `lib/src/sources/AGENTS.md:28-29` and `lib/src/reactions/AGENTS.md:35-36` require a major bump for vtable changes, but the SDK is `0.12.0` and `components/plugin-sdk/src/ffi/metadata.rs:19-27` explicitly requires a minor bump for breaking changes while pre-1.0. Please align the guidance with that policy.

4. **The proxy lifetime guidance contradicts the implementation.** `components/host-sdk/src/proxies/AGENTS.md:68-70` says the cdylib unloads after all proxies are dropped, but `components/host-sdk/src/loader.rs:108-126` intentionally retains the library process-wide to prevent unsafe `dlclose`. Please document the actual lifetime behavior.

The final SHA also currently has a red Plugin Coverage check at `components/sources/http/tests/webhook_integration.rs:749`.

AGENTS.md
```diff
@@ -0,0 +1,61 @@
...
+  `test-ffi.yml` separately re-runs the host-sdk integration test on Linux/macOS/Windows.
+- `make clippy` uses the same feature list as `ci-lint.yml`. `--all-features` is deliberately
+  avoided: `bundled-jq` compiles jq from source and is flaky.
+- libjq is **not** needed to build: `drasi-middleware` is `default = []` and only the opt-in `jq`
+  feature dynamic-links system libjq (set `JQ_LIB_DIR`). Linting is the exception, since
+  `make clippy` enables `drasi-middleware/all`, which includes `jq`.
+- Not hermetic: a plain `cargo test --workspace` starts testcontainers (Redis and Postgres images
+  among others), so Docker must be running. Some plugin builds need `protobuf-compiler`; the kafka
+  crates pull in `cmake` via `rdkafka/cmake-build` under `dynamic-plugin`.
+- No `.unwrap()`, no `print!`/`println!`, enforced by `[workspace.lints.clippy]` in the root
```
danielgerlag: This is too absolute. The repository has intentional production exceptions: `core/src/query/mod.rs` allows `unwrap_used`, while `xtask/src/main.rs` and the log reaction allow printing. Please describe these as default workspace lints with test and explicitly annotated production exceptions; otherwise agents may rewrite intentional APIs/output or misread what CI enforces.

AGENTS.md
```diff
@@ -0,0 +1,61 @@
...
+  `Cargo.toml` with `RUSTFLAGS=-Dwarnings`; `clippy.toml` permits both in tests only. `.rs` files
+  carry the Apache 2.0 header; commits need `Signed-off-by:` (`git commit -s`).
+
+## Before editing
+
+- Storage is trait-injected and the contracts live in **two** crates: engine-side index/checkpoint/
+  session traits in `core/src/interface/`, runtime provider traits (`StateStoreProvider`,
+  `WalProvider`, `SecretStoreProvider`, `IdentityProvider`) in `lib/src/`. Check both.
+- `core/src/index_cache/` wraps `ElementIndex`, `ResultIndex` and `FutureQueue`; changing one of
```
danielgerlag: This collapses several distinct extension seams. `IndexBackendPlugin` also requires fallible `create_indexes(...) -> Result<CreatedIndexes, IndexError>` and `is_volatile()`, which affects query recovery. `DrasiLibBuilder` directly accepts index, state-store, WAL, identity, and secret-store providers as well. Please document the exact index contract and direct builder-injection paths rather than implying everything arrives through only the three component traits or FFI.

lib/AGENTS.md
```diff
@@ -0,0 +1,50 @@
+# AGENTS.md: `drasi-lib`
+
+Orientation for editing `drasi-lib` (Rust path `drasi_lib::`). `lib/README.md` is the long-form
+reference; the per-directory `AGENTS.md` files under `lib/src/` own the FFI-boundary contracts.
+
+## Shape
+- `lib/src/lib.rs` is the API surface: the curated re-export list is what embedders and plugin
```
danielgerlag: `doc(hidden)` does not make these modules test-only. They remain public, production plugins import them directly, and the curated re-exports are incomplete (`QueryResult` and `ResultDiff` are examples). Please describe them as undocumented but externally consumed public APIs; suggesting that only tests may use them risks breaking the plugin ecosystem.

lib/AGENTS.md
```diff
@@ -0,0 +1,50 @@
...
+- The only construction path is `DrasiLib::builder()…build().await?` then `start().await?`.
+  `DrasiLib::new` and `initialize` are `pub(crate)`, and `build()` calls `initialize()` for you.
+- `DrasiLib` is Arc-internal and implements `Clone`, so cloning is cheap; clone it to share.
+- `lib/src/lib_core.rs` holds the struct, lifecycle and public accessors; per-component CRUD is in
+  `lib/src/lib_core_ops/*_ops.rs` and inspection logic in `lib/src/inspection.rs` (the `*_ops`
+  inspection methods just forward). Check all three before concluding a method does not exist.
+  `ComponentGraph` (`lib/src/component_graph/`) is the source of truth for status and relationships.
+
+## Rules that prevent damage
+- Error layering, per the `lib/src/error.rs` module docs: public API returns
```
danielgerlag: Please scope this rule to the high-level `DrasiLib`, `*_ops`, and `InspectionAPI` boundaries. Publicly exposed advanced APIs such as `ComponentGraph` and the manager accessors intentionally expose methods returning `anyhow::Result`. Applying this sentence to every public method would drive incompatible signature changes beyond the contract documented in `lib/src/error.rs`.

lib/src/sources/AGENTS.md
```diff
@@ -0,0 +1,38 @@
...
+rather than reinterpreted as `repr(Rust)` values on the far side. Other pointers in this
+ABI mean other things, so read the signature rather than assuming. See
+`lib/src/channels/AGENTS.md` for why the payload rule exists.
+
+## Changing the `Source` trait
+
+Adding, removing or changing a method means updating all four of these together:
+
+1. `components/plugin-sdk/src/ffi/vtables.rs`, `SourceVtable`: the function pointer field.
+2. `components/plugin-sdk/src/ffi/vtable_gen.rs`, `build_source_vtable()`: the plugin side.
```
danielgerlag: This checklist is incomplete. `build_source_vtable_from_boxed()` has an independent copy of every entry and is what the descriptor/factory plugin path actually uses. The reaction side has the same `build_reaction_vtable_from_boxed()` twin. Please name both builders in the respective guides so a trait change cannot update only the generic path.

lib/src/sources/AGENTS.md
```diff
@@ -0,0 +1,38 @@
...
+   Async methods use the `block_on` wrapping the existing entries use.
```
danielgerlag: Please do not instruct agents to use `block_on` here. `vtable_gen.rs` explicitly documents that nested `plugin_handle.block_on()` panics; the source entries use `dispatch_to_runtime()` instead. This should name that helper and explicitly prohibit `block_on`, since several entry points are not protected by an FFI panic guard.

lib/src/sources/AGENTS.md
```diff
@@ -0,0 +1,38 @@
...
+3. `components/host-sdk/src/proxies/source.rs`, `SourceProxy`: the host side dispatch.
+4. `components/plugin-sdk/src/ffi/metadata.rs`, `FFI_SDK_VERSION`: the vtable layout
+   changed, so this is a major bump.
+
+Missing one does not fail to compile. It fails at plugin load, or worse, at runtime.
```
danielgerlag: This is false as written. Exhaustive vtable struct literals catch a missing field at compile time, and a required proxy trait method also fails to compile. The silent cases are narrower: a new trait method with a default body can bypass proxy dispatch, while an omitted version bump compiles and fails at load/runtime. Please describe those exact cases.

components/plugin-sdk/src/ffi/AGENTS.md
```diff
@@ -46,15 +46,23 @@ If you change `FfiStr`, `FfiResult`, `FfiComponentStatus`, `FfiDispatchMode`, or
...
+and the `0.10.0` entry in `metadata.rs`.
 
-1. Define an `#[repr(C)]` envelope struct in `vtables.rs` (like `FfiSourceEvent`)
-2. Add wrapping logic in `vtable_gen.rs` (box the Rust type, store as `*mut c_void`)
-3. Add unwrapping logic in `components/host-sdk/src/proxies/` (cast back with `Box::from_raw`)
-4. Add a `drop_fn` field to the envelope for safe cleanup
-5. Bump `FFI_SDK_VERSION` in `metadata.rs`
+Instead, follow what `payload.rs` already does:
+
+1. Give the payload `serde` derives so it can be serialized.
```
danielgerlag: This does not match `payload.rs`. `SourceEventWrapper` and `BootstrapEvent` do not have serde derives; the SDK defines separate mirror DTOs (`SourceEventPayload` and `BootstrapEventPayload`) with explicit conversions. The consumer must also go through a new `consume_*` wrapper over `take_ffi_payload()` so the payload-size cap and null-`drop_fn` guard are preserved. Please document that DTO/conversion/hardened-consumer pattern instead of directing agents to derive serde on domain types and hand-roll decoding.

lib/src/channels/AGENTS.md
```diff
@@ -0,0 +1,39 @@
+# AGENTS.md: drasi-core/lib/src/channels
+
+## The payload rule, and why it exists
+
+Event types defined here cross the dynamic plugin FFI boundary. They cross **by value as
```
danielgerlag: This is inaccurate for two of the main event types. `SourceEventWrapper` and `BootstrapEvent` are not serialized directly and have no serde derives; `components/plugin-sdk/src/ffi/payload.rs` maps them through separate wire DTOs. Only `QueryResult` follows the direct serialization model. Please distinguish these paths and require updating both DTO mappings when wrapper fields change.

lib/src/channels/AGENTS.md
```diff
@@ -0,0 +1,39 @@
...
+drop function for the serialized bytes. The distinction matters: seeing a raw pointer in
+the vtable does not mean a Rust value is being handed across.
+
+For which types cross and in what form, read the `Ffi*` envelope definitions in
+`components/plugin-sdk/src/ffi/vtables.rs`. That is the authority and it stays current.
+
+## What to update when changing these types
+
+**Event payload fields.** Changing fields on a serialized payload does not alter the
+envelope layout, but host and plugin must still agree on how to decode it, so bump
```
danielgerlag: Please require a pre-1.0 minor bump explicitly, currently `0.12.x -> 0.13.0`. The loader compares only major/minor, so a patch-only bump is accepted and provides no protection against an incompatible payload schema. The deleted lowercase guide stated the minor-bump requirement correctly.

lib/src/bootstrap/AGENTS.md
```diff
@@ -1,4 +1,4 @@
-# agents.md — drasi-core/lib/src/bootstrap
+# AGENTS.md: drasi-core/lib/src/bootstrap
```
danielgerlag: Activating this guide exposes two critical omissions. The current plugin ABI drops `BootstrapContext.properties` and passes `None` for `SourceSubscriptionSettings`; Oracle relies on the SCN property for consistent snapshots. `BootstrapResult` also crosses through two independent FFI routes but is absent from the type table/checklist. Please document the lossy boundary (or extend it) and add both `BootstrapResult` conversion paths before making this authoritative agent guidance.

lib/src/config/AGENTS.md
```diff
@@ -1,4 +1,4 @@
-# agents.md — drasi-core/lib/src/config
+# AGENTS.md: drasi-core/lib/src/config
```
danielgerlag: The newly activated subscription mapping in this file is inaccurate. Only `nodes` and `relations` use JSON `FfiStr`s; `resume_from` is a raw pointer/length and the booleans are direct fields. The host deconstructs all seven fields, and two plugin reconstruction paths must stay synchronized. Please document that exact mapping so recovery position and position-handle fields are not omitted.

components/plugin-architecture.md
```diff
@@ -67,7 +67,7 @@ entry points only when the `dynamic-plugin` feature is enabled on a plugin crate
 | `drasi-host-sdk` | `drasi-core/components/host-sdk` | Host-side SDK: `PluginLoader`, proxy types (impl Source/Reaction/SourcePlugin), callback wiring, schema merging |
 | `drasi-server` | `drasi-server/` | Application: REST API, config persistence, OpenAPI spec, server lifecycle — uses `drasi-host-sdk` for dynamic loading |
 | `drasi-lib` | `drasi-core/lib/` | Core processing: query engine, routers, channels — no FFI awareness |
-| `drasi-core` | `drasi-core/core/` | Core types: Element, SourceChange, QueryResult — crosses FFI as opaque pointers |
+| `drasi-core` | `drasi-core/core/` | Core types: Element, SourceChange, QueryResult — cross FFI as serialized payloads |
```
danielgerlag: `QueryResult` is not a `drasi-core` type; it is defined in `drasi-lib` alongside the other channel payload types. Please move it to the `drasi-lib` row and qualify the adjacent `no FFI awareness` claim, since these `lib/src/channels/` types define the serialized wire schema consumed by the SDKs.

components/plugin-architecture.md
```diff
@@ -141,30 +141,38 @@ pub extern "C" fn drasi_plugin_init() -> *mut FfiPluginRegistration
...
+// through payload_drop_fn.
 ```
 
+> Earlier versions passed these types as opaque `Box::into_raw` pointers that the other
+> side reclaimed with `Box::from_raw`. That is undefined behaviour: `repr(Rust)` has no
+> stable layout across independently compiled cdylibs, and `bytes::Bytes` carries a
+> `&'static` vtable pointer valid only in the producing module. It caused
+> non-deterministic heap corruption. See issue #602, the module docs in
+> `drasi-core/components/plugin-sdk/src/ffi/payload.rs`, and the `0.10.0` entry in
+> `metadata.rs`. Do not reintroduce that pattern for payloads.
```
danielgerlag: This prohibition is correct, but the ownership section later in this document still says that `many rich Rust types` cross as `Box::into_raw`/`Box::from_raw`. That directly contradicts this new safety rule. Please rewrite the later section to limit cross-boundary boxing to `#[repr(C)]` POD envelopes, responses, vtables, and registration structs, explicitly excluding payloads.

components/plugin-architecture.md
```diff
@@ -477,9 +485,10 @@ self-contained with a stable C ABI boundary.
 
 ### Can plugins use different tokio versions?
 
-Yes. Each cdylib plugin statically links its own tokio. The host and plugins can
-use different tokio versions as long as the DrasiLib trait types (which cross as
-opaque pointers) remain layout-compatible (validated via version metadata).
+Yes. Each cdylib plugin statically links its own tokio. Nothing that crosses the
+boundary depends on tokio's layout: payloads are serialized, and the `#[repr(C)]`
+envelopes and vtables have a stable layout whose compatibility is asserted by the
+`FFI_SDK_VERSION` check at load time.
```
danielgerlag: The compatibility check is not always asserted. `validate_plugin_metadata` logs a warning and skips validation when the metadata symbol is absent or returns null. For appended `FfiPluginRegistration` fields, the host must therefore gate each read using a `MIN_SDK_VERSION_WITH_*` check. Please qualify this statement and document the append-only plus host-side gating protocol.

danielgerlag (changes requested):
Follow-up accuracy review of the newly standardized agent instructions. The inline comments cover findings that can be anchored to changed lines. Three additional inaccuracies are on unchanged lines and cannot be placed precisely in the diff:

- `components/plugin-sdk/src/ffi/AGENTS.md:41-47` says `FfiStr`, `FfiOwnedStr`, and `FfiResult` are defined in the local `types.rs`; they are defined in `components/ffi-primitives/src/types.rs` and only re-exported here.
- `components/plugin-architecture.md:495-497` says every FFI entry point is guarded by `catch_unwind` and converted to `FfiResult::Err`; only initialization is guarded by the macro, and many non-`FfiResult` vtable entries remain unguarded.
- `components/plugin-architecture.md:122` says plugins export exactly two symbols, but `drasi_plugin_shutdown` is a third export that the host deliberately must not call.

These should be corrected before the files become authoritative instructions for coding agents.

AGENTS.md
amansinghoriginal: Agreed the tests-only claim was wrong. Rather than enumerating the exceptions, which is the kind of list that rots, we removed the lint prose entirely: [workspace.lints] and clippy.toml are the authority and make clippy is already documented as the check. The doc now only keeps what no tool enforces locally: the license header and DCO sign-off.

AGENTS.md
amansinghoriginal: Both points are right. Fixed the other way though: the quoted signature was already stale, so instead of documenting the fuller contract we removed signatures from the doc and pointed at core/src/interface/index_backend.rs as the authority. The injection bullet now says instances arrive as trait objects, via the component and provider traits or the FFI seam, without enumerating the trait set.

lib/AGENTS.md
amansinghoriginal: Confirmed: components/ has roughly 490 non-test imports of these module paths, and QueryResult/ResultDiff are only reachable that way. The tests-only wording came from the stale comment in lib.rs itself; fixed both. The doc now calls these modules externally consumed public API.

lib/AGENTS.md
amansinghoriginal: Agreed on scope. Went further: the doc no longer restates the contract at all, just its direction and scope (DrasiLib / *_ops / InspectionAPI vs everything beneath) plus a pointer to error.rs, which we also fixed to carry the scoping in its Rules section. The scoped phrasing covers ComponentGraph implicitly.

lib/src/sources/AGENTS.md
amansinghoriginal: Confirmed: the _from_boxed builders are independent copies (no delegation) and the descriptor path calls them. Both guides now name both builders and the no-shared-code trap. Note a field addition does fail to compile in both; the silent case is behavioral drift between the copies. Identity and secret-store vtables have the same twins.

lib/src/sources/AGENTS.md
amansinghoriginal: That wording came from the old lowercase guide, not this rewrite, but agreed it is wrong. Fixed: dispatch_to_runtime required, block_on prohibited.

lib/src/sources/AGENTS.md
amansinghoriginal: Agreed it was false. Rather than describing what the compiler catches (it teaches that itself), the doc now names only the two omissions nothing catches: default-body trait methods and a forgotten version bump.

components/plugin-sdk/src/ffi/AGENTS.md
amansinghoriginal: Agreed, and we cut deeper: the recipe is gone. payload.rs's module docs already document the DTO and hardened-consumer pattern, so the doc now just points there, with one warning that a parallel codec loses the size cap and null drop_fn guards.

lib/src/channels/AGENTS.md
amansinghoriginal: Right: only QueryResult serializes directly; the derive claim was wrong for the other two. Rather than encoding the two-path split, the doc now points at payload.rs as the mapping and warns that a new event field is not on the wire until the mapping carries it; from_wrapper copies field by field, so nothing fails to compile if you forget.

lib/src/channels/AGENTS.md
amansinghoriginal: The loader semantics are right, but metadata.rs already states the minor-bump rule, bolded, directly above the constant; the edit site teaches it at the moment of the bump. We'd rather keep the single copy there than duplicate it here and let the copies drift.

lib/src/bootstrap/AGENTS.md
amansinghoriginal: Confirmed: the FFI path rebuilds the context with new_minimal (empty properties) and passes None for settings, and the oracle source sets the SCN property its bootstrapper requires, so that pairing is in-process only today. Added a lossy-boundary warning to the doc. Extending the ABI and the BootstrapResult rows feel like a follow-up issue rather than this rename PR: the gap is in the code, not just the doc.

lib/src/config/AGENTS.md
amansinghoriginal: Right on all three encodings, and the file's own detail section omitted exactly the recovery fields. Fixed by naming the mixed encodings, the two easy-to-omit recovery fields, and the twin reconstruction sync, and pointing at subscribe_fn in vtables.rs as the authoritative mapping instead of freezing all seven args in prose.

components/plugin-architecture.md
amansinghoriginal: Right, and went further: the crate rows now describe roles instead of type rosters, so the misfiled QueryResult mention is gone entirely; the lib row says it defines the channel event types that are the FFI wire payloads, with no FFI code itself. Also dropped routers while there: lib/src has no routers directory anymore.

components/plugin-architecture.md
amansinghoriginal: Right, that opener was pre-#602 language sitting above a still-correct contract. Reworded: the boxed set is the #[repr(C)] structures (envelopes, responses, vtables, registration), with payloads explicitly excluded. The allocator requirement and the cargo-deny note are unchanged.

components/plugin-architecture.md
amansinghoriginal: Right, and the same overstatement was stronger in the ffi AGENTS.md. Went further there: the version section was restating metadata.rs's own doc (down to a stale 0.8.x bump example), so it now just points at the constant and keeps the two things no edit site teaches: the skipped check for plugins without metadata, and the MIN_SDK_VERSION_WITH_* gating for appended fields.

ruokun-niu (approved):

amansinghoriginal:
> **Requesting changes** because the current symlink strategy is not portable and is already breaking review automation.

@danielgerlag Since your review the PR was redone from scratch.

- Symlinks are gone. Each AGENTS.md now has a one line CLAUDE.md beside it containing @AGENTS.md, the import Claude Code documents. Survives Windows checkouts.
- The deep AGENTS.md files you flagged for accuracy are deleted, not fixed. Same for the old lowercase agents.md files and both old CLAUDE.md files, about 1000 lines total.
- In their place: six short AGENTS.md files at root, core, lib, components, shared-tests, examples. Every line was verified against the current code, Makefile and CI configs.
- plugin-architecture.md now describes the post #602 serialized payload design instead of opaque pointers.
- Two stale doc comments fixed in lib/src/error.rs and lib/src/lib.rs.
- Rebased on latest main, including the bootstrap FFI rework.

agentofreality (approved):

PR #707, opened 2026-08-04
Return a clear error for temporal reads when archive is disabled (#699)

### LINKED ISSUE #699: Temporal reads on an archive-disabled index panic (RocksDB) or return nothing (Garnet)
Temporal functions (drasi.getVersionByTimestamp, drasi.getVersionsByTimeRange, "past()" in the docs) are served by the ElementArchiveIndex trait. The RocksDB backend resolves the archive column family with .expect("Archive CF not found"), but that CF only exists when the index is opened with enable_archive set. With archive disabled, the index set still hands out the element index as the archive index, so the first temporal read panics inside spawn_blocking. The caller gets IndexError::Other(JoinError): a panic backtrace in the logs and an error that names a RocksDB internal, not the enable_archive setting, and cannot be pattern-matched. drasi-lib then drops the source change, so the query emits nothing.

Current behavior per backend:
- RocksDB: get_element_as_at and get_element_versions panic. clear() returns Ok(()).
- In-memory: returns IndexError::NotSupported from every method.
- Garnet: reads never check archive_enabled and silently return nothing, a wrong answer with no error.

The path is latent today because drasi-server hardcodes enable_archive to true. Archive off by default is planned (drasi-project/drasi-server#151; the archive CF is double-written on every element update). A clear error is a prerequisite for that flip.

Proposal: get_element_as_at and get_element_versions on an archive-disabled index return a new IndexError::ArchiveNotEnabled variant with a message like "archive index not enabled for this query; set enable_archive to use past()", identical across all backends. We have no backward compatibility concerns right now.

clear() keeps returning Ok(()) or NotSupported: drasi-lib's clear_persistent_indexes tolerates only those and would otherwise fail query reset for archive-disabled queries.

One test per backend per read method; no behavior change when archive is enabled.

Returning empty results is rejected as a silently wrong answer; it is Garnet's behavior today. Separate follow-up: fail fast at query creation when a query uses temporal functions without archive (needs a capability flag plus AST inspection). Also out of scope: with no storage backend, the functions are never registered and the same mistake surfaces as UnknownFunction.

amansinghoriginal:
# Description

Temporal reads (drasi.getVersionByTimestamp, drasi.getVersionsByTimeRange) on an index without archive enabled behaved differently per backend: RocksDB panicked on the missing archive column family and surfaced an opaque JoinError, Garnet silently returned nothing, and the in-memory index returned NotSupported.

All backends now return a new IndexError::ArchiveNotEnabled from get_element_as_at and get_element_versions when archive is off. Its message says what to do: "archive index not enabled for this query; set enable_archive to use past()". The RocksDB read paths no longer contain a panic.

clear() is unchanged (Ok for RocksDB, NotSupported for in-memory), since drasi-lib's clear_persistent_indexes treats only those as success and anything else would fail query reset for archive-disabled queries.

Tests: one per backend per read method asserting the error, plus enabled baselines asserting no behavior change. Garnet tests use the existing Redis testcontainer helper.

## Type of change

Fixes: #699

### DIFF
--- a/core/src/in_memory_index/in_memory_element_index.rs
+++ b/core/src/in_memory_index/in_memory_element_index.rs
@@ -594,7 +594,7 @@ impl ElementArchiveIndex for InMemoryElementIndex {
         time: ElementTimestamp,
     ) -> Result<Option<Arc<Element>>, IndexError> {
         if !self.archive_enabled {
-            return Err(IndexError::NotSupported);
+            return Err(IndexError::ArchiveNotEnabled);
         }
 
         let guard = self.element_archive.read().await;
@@ -613,7 +613,7 @@ impl ElementArchiveIndex for InMemoryElementIndex {
         range: TimestampRange<ElementTimestamp>,
     ) -> Result<ElementStream, IndexError> {
         if !self.archive_enabled {
-            return Err(IndexError::NotSupported);
+            return Err(IndexError::ArchiveNotEnabled);
         }
 
         let from = range.from;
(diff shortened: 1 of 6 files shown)


danielgerlag (approved):

core/src/in_memory_index/in_memory_element_index.rs
ruokun-niu, line 636: `clear()` behavior is inconsistent across backends: in-memory returns `NotSupported` when archive is disabled (`in_memory_element_index.rs:635`), RocksDB returns `Ok` (`rocksdb/tests/archive_disabled.rs:64`), and Garnet has no guard at all (`garnet/.../archive_index.rs:241`). I know this is probably a pre-existing issue, but do you think we should address this inconsistency in this PR?

ruokun-niu (approved):
approved with a minor comment

Issue #743, opened 2026-08-16
Add durable reaction recovery conformance tests


amansinghoriginal:
## Summary

Add end-to-end conformance tests for the durable reaction recovery guarantees defined in #393 and the Reaction Checkpoints design.

Existing tests exercise stop and restart within the same process using in-memory query state. They do not verify reconstruction from persistent query storage and durable reaction checkpoints.

## Acceptance criteria

- Run recovery coverage against RocksDB and Garnet query indexes.
- Destroy all runtime objects and construct a fresh `DrasiLib` instance using the same persistent query and reaction stores.
- Verify query indexes, source checkpoints, result sequences, outbox entries, and live results are committed atomically.
- Use fault injection to verify failures between transaction stages cannot advance a source checkpoint without preserving its query output.
- Verify result sequences continue from the persisted high-water mark.
- Verify `fetch_snapshot` returns the complete persisted result set.
- Verify `fetch_outbox` returns retained results and reports unavailable positions.
- Verify retained results missed during downtime are delivered at least once.
- Verify reaction checkpoints advance only after side effects complete and interrupted side effects are replayed.
- Verify query reconfiguration and delete-and-recreate remove all output state from the previous configuration.
- Verify durable reactions cannot subscribe to volatile queries.

The tests must detect reset, partial restoration, stale state, non-atomic persistence, and checkpointing before processing completes.

Issue #393, opened 2026-04-24
Reaction Recovery (Resumable Reactions)


amansinghoriginal:
Implements durable delivery and bootstrap for the Query-to-Reaction pipeline in Drasi Lib.

Queries maintain a sequenced outbox and live result snapshot. Reactions can bootstrap from current state (`fetch_snapshot`), catch up from a checkpoint (`fetch_outbox`), and resume after crashes using durable checkpoints with configurable recovery policies (`Strict`, `AutoReset`, `AutoSkipGap`).

**Design docs:** [Reaction Checkpoints](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints)

### Dependency Graph

```mermaid
graph TD
    I395["#395 QueryResult sequence &\nResultDiff row_signature"] --> I396["#396 QueryOutputState &\nin-memory outbox"]
    I396 --> I397["#397 Query bootstrap APIs\nfetch_snapshot / fetch_outbox"]
    I398["#398 Reaction trait extensions &\nbootstrap hook"] --> I399["#399 ReactionBase checkpoint\n& dedup primitives"]
    I397 --> I400["#400 Reaction startup\norchestration"]
    I398 --> I400
    I399 --> I400
    I400 --> I401["#401 Runtime gap\ndetection"]
    I395 --> I405["#405 Update simple\nreaction plugins"]
    I398 --> I405
    I399 --> I405
    I395 --> I408["#408 Update http & grpc\nreaction plugins"]
    I398 --> I408
    I399 --> I408
    SR["Source Recovery\n#367, #370"] --> I402["#402 Persistent storage\ntraits & transaction"]
    I402 --> I403["#403 RocksDB\nimplementation"]
    I402 --> I404["#404 Garnet\nimplementation"]
    I396 --> I406["#406 Observability\ninstrumentation"]
    I399 --> I406
    I400 --> I406
```

PR #745, opened 2026-08-17
test: add reaction recovery conformance coverage

amansinghoriginal:
## Summary

- Add a RocksDB and Redb full-process recovery test.
- Verify snapshot, outbox, checkpoint, and sequence restoration.
- Verify fresh trigger reactions do not replay historical results.
- Verify durable reactions reject volatile queries.
- Correct the recording fixture to checkpoint after its observable side effect.

## Current result

The existing recovery tests pass. The new tests expose three current failures:

- Persistent query output state is not rehydrated after restart.
- Fresh trigger reactions replay retained history.
- Durable reactions can start against volatile queries.

This PR remains a draft until the recovery implementation satisfies these tests.

## Not yet covered

- Atomic commit fault injection, interrupted side-effect replay, unavailable-position recovery, and reconfiguration cleanup will be added alongside the corresponding remediation work under #393.
- #743 tracks completion of the conformance coverage.
- Garnet coverage can be deferred, if we're not planning to support it long term.

Related to #743

### DIFF
--- a/lib-integration-tests/Cargo.toml
+++ b/lib-integration-tests/Cargo.toml
@@ -14,6 +14,9 @@ drasi-core.workspace = true
 drasi-source-postgres.workspace = true
 drasi-source-application.workspace = true
 drasi-reaction-application.workspace = true
+drasi-index-rocksdb.workspace = true
+drasi-state-store-redb = { path = "../components/state_stores/redb" }
+drasi-wal-redb = { path = "../components/wals/redb" }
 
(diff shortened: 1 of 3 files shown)


agentofreality:
This remains the historical conformance predecessor to current #743 and open PR #825. Preserve its discussion/tests, but use the current durability stack #818-#824 rather than duplicating it. Tracked by #889/#890.

Issue #692, opened 2026-08-03
Reduce the fixed per-CF memory floors in the RocksDB index (write_buffer_size, arena_block_size)


amansinghoriginal:
## Problem

Each persistent query opens a RocksDB database with 15 column families. RocksDB defaults every CF to a 64 MiB write buffer and may allocate arena blocks up to 1 MiB, even though most CFs are small or mostly idle.

Four point-lookup CFs also allocate memtable bloom filters at 2% of the write buffer. Together, these fixed costs consume about 6.9 MiB of RSS per query DB at rest. Under load, the current 128 MiB per-DB cap can provision 2.5 GiB of memtables across 20 queries. Because 15 x 64 MiB far exceeds the 128 MiB DB cap, the cap forces busy CFs to flush early at irregular sizes, increasing SST and compaction churn.

## Proposal

Apply an explicit two-tier policy to every CF, including the mandatory default CF:

- Large, 16 MiB: `elements`, `inbound`, `outbound`, `values`, `archive`, `outbox`, `live_results`
- Small, 8 MiB: all remaining CFs

Derive `arena_block_size` as:

```text
clamp(write_buffer_size / 64, 64 KiB, 1 MiB)
```

Centralize this policy in one sizing module and preserve it whenever a CF is dropped and recreated by `clear()`.

Expose programmatic overrides through `RocksIndexOptions` and `RocksDbIndexProvider::with_write_buffer_sizes`. No descriptor or config-file fields are added here. Shared budgets and budget-derived tier sizes are tracked by #634 and #710.

## Expected result

Measurements on a 20-query workload:

- At-rest RSS per DB: 6.9 MiB -> 2.8 MiB
- Bloom-filter floor per query: 5.2 MiB -> 1.0 MiB
- Peak memtable usage: approximately halved
- Sustained multi-CF ingest: parity or better

A transient point-read slowdown of about 9% was observed immediately after bulk loading, converging as L0 compaction completed.

## Validation

Read RocksDB's generated OPTIONS files to verify:

- Effective sizes for all 15 CFs
- Custom programmatic sizes
- Arena clamp behavior
- Every CF drop/recreate path

## Non-goals

- Sharing memory across query databases
- Adding an operator-facing memory budget
- Deriving sizes dynamically from that budget
- Merging column families
- Changing durability, transactions, or the one-DB-per-query design

amansinghoriginal:
Additional design context:

[...]

Issue #634, opened 2026-07-25
Share block cache and write buffer budgets across query index DBs


amansinghoriginal:
## Problem

Nothing is shared across query index DBs. Each query's RocksDB opens with private block caches (about 145 MiB of cache capacity per query, measured near empty) and a private per-DB memtable cap (db_write_buffer_size, 128 MiB). Total index memory therefore scales linearly with query count regardless of data volume: 20 queries are provisioned for several GiB of cache capacity plus 2.5 GiB of memtables while holding 11 MB of index data on disk. There is no bound an operator can set; the only lever on total index memory is query count.

## Proposal

Keep one DB per query (session atomicity, trivial lifecycle, small blast radius). Share the budgets:

1. RocksDbIndexProvider accepts one shared block Cache used by all its DBs and CFs.
2. RocksDbIndexProvider accepts one shared WriteBufferManager, charged to that Cache, replacing per-DB db_write_buffer_size.

The provider takes both as handles, not sizes, so the sharing scope (provider, instance, or process) stays a caller decision: drasi-core and drasi-lib expose the full programmatic surface, and each consumer curates its own operator knobs. When no handles are passed the provider constructs its own, with proposed defaults of a 256 MiB cache and a 128 MiB write budget shared across that provider's DBs; these defaults are policy for unconfigured embedders and are open to review here. The default manager is constructed with allow_stall false: at the limit, saturation degrades to flush pressure rather than stalled writes, since a stalled write would block its query's whole pipeline (process_source_change holds the per-query change_lock for the session). This choice is open to review here. The IndexBackendPlugin trait is unchanged. Requires the rocksdb crate 0.21 to 0.22 bump for the WriteBufferManager API.

Prerequisite, tracked in #709: optimize_for_point_lookup constructs a private cache internally, so it must first be unrolled into explicit per-CF options that preserve the settings it bundles.

Expected result: total index memory bounded by one caller-chosen budget instead of growing linearly with query count, with the per-query marginal cost reduced to the small fixed floors from #692.

## Validation

Measured on a 20-query, 8-source Postgres CDC workload with a 10k-event replay: the write buffer budget held with zero saturation events across a full run at a 1 GiB budget. Before merge, re-run query-perf sustained ingest against the shared budget to confirm no throughput regression from flush pressure.

## Non-goals

No durability changes (commits intentionally use sync=false). No change to the one-DB-per-query layout or the session transaction model.

[...]

PR #769, opened 2026-08-20
feat(index-rocksdb): scale CF buffers with write budget

### LINKED ISSUE #710: Derive per-CF write buffer sizes from the shared write-buffer budget
After #634, all query DBs share one WriteBufferManager budget; after #692, every CF has a fixed write_buffer_size (16 MiB large, 8 MiB small). The two are unconnected: with a large budget and a small query fleet, elements still flushes every 16 MiB, producing small SSTs and compaction churn while the budget sits idle. A larger budget should buy bigger self-triggered flushes, not only more cache.

Proposal: at DB open, derive the tier sizes as a pure function of the injected budget, read via WriteBufferManager::get_buffer_size() (rust-rocksdb 0.22): large = clamp(budget / 8, 16 MiB, 64 MiB), small = large / 2. At the #634 default budget of 128 MiB this yields exactly the #692 constants, so unconfigured deployments are unchanged. Arena block sizes and memtable bloom floors follow the existing derivations. The ceiling does double duty: it caps flush and compaction burst size, and it bounds the eager memtable bloom floors (2% of write_buffer_size on the four point-lookup CFs), which otherwise scale back up with budget.

Deliberately not addressed: small budgets. When the sum of active buffer ceilings exceeds the budget, WriteBufferManager arbitration is the designed mechanism, not a pathology: the flush triggers in whichever DB is actively writing and picks its oldest memtable, so pressure lands on busy queries while idle ones stay cheap. Derived sizes never go below the #692 floors. #693's saturation logging makes this regime observable.

Caveats and non-goals: sizes are derived once at open. set_buffer_size() can change the budget at runtime and derived sizes will not follow. Live re-derivation is not possible today: rust-rocksdb exposes set_options_cf only for non-transactional DBs, and TransactionDB lacks it in both 0.21 and 0.22, so it would need an upstream contribution first. Whether the manager stalls writes at the limit (allow_stall) is decided in #634; this issue inherits that choice. Note that a stalled write blocks its query's whole pipeline, since process_source_change holds the per-query change_lock for the session.

Validation: flush counts and SST size distributions at a small and a large budget versus fixed tiers, query-perf sustained ingest, effective values verified from the OPTIONS files.

amansinghoriginal:
## Summary

- derive per-CF write buffer sizes from the shared write-buffer budget
- preserve the existing 16/8 MiB tiers at the default 128 MiB write budget
- cap large and small tiers at 64/32 MiB
- retain resolved sizing when column families are recreated
- document how shared manager pressure can preempt per-CF thresholds
- update sizing documentation and regression coverage

## Validation

- `cargo test -p drasi-index-rocksdb`
- `cargo clippy -p drasi-index-rocksdb --all-targets --features plugin-descriptor -- -D warnings`
- `cargo fmt --all -- --check`

Two reversed-order, 600k-event query-perf runs:

| Write budget | Throughput | Flushes | Peak RSS |
|---:|---:|---:|---:|
| 128 MiB | 9,434 events/s | 12 | 72 MiB |
| 512 MiB | 8,779 events/s | 1 | 145 MiB |

The larger tier substantially reduced self-triggered flushes, with the expected memory tradeoff and lower throughput for this repeated-update workload.

Per-CF sizes are flush thresholds, not memory reservations. The provider-wide `WriteBufferManager` remains authoritative across CFs and query databases and can trigger a flush before an individual threshold under aggregate multi-query pressure. The measurement above uses a single query; multi-query fleets may reach shared pressure sooner.

Closes #710

### DIFF
--- a/components/indexes/rocksdb/src/options.rs
+++ b/components/indexes/rocksdb/src/options.rs
@@ -26,9 +26,9 @@ pub struct RocksIndexOptions {
 impl RocksIndexOptions {
     /// Create options with the memory resources used to open and maintain the index.
     ///
-    /// Per-column-family write-buffer sizes are captured from the shared
-    /// manager's current budget and remain fixed for the lifetime of these
-    /// options.
+    /// Per-column-family write-buffer sizes are computed once from the shared
+    /// manager's budget when `new` is called. They remain fixed for the
+    /// lifetime of these options, even if the manager's budget changes later.
     pub fn new(archive_enabled: bool, direct_io: bool, memory_budget: RocksDbMemoryBudget) -> Self {
         let (large_write_buffer_size, small_write_buffer_size) =
             crate::sizing::write_buffer_sizes_for_budget(
--- a/components/indexes/rocksdb/src/options.rs
+++ b/components/indexes/rocksdb/src/options.rs
@@ -25,13 +25,22 @@ pub struct RocksIndexOptions {
 
 impl RocksIndexOptions {
     /// Create options with the memory resources used to open and maintain the index.
+    ///
+    /// Per-column-family write-buffer sizes are captured from the shared
+    /// manager's current budget and remain fixed for the lifetime of these
+    /// options.
     pub fn new(archive_enabled: bool, direct_io: bool, memory_budget: RocksDbMemoryBudget) -> Self {
+        let (large_write_buffer_size, small_write_buffer_size) =
+            crate::sizing::write_buffer_sizes_for_budget(
+                memory_budget.write_buffer_manager().get_buffer_size(),
+            );
+
         Self {
             archive_enabled,
             direct_io,
             memory_budget,
-            large_write_buffer_size: crate::sizing::DEFAULT_LARGE_WRITE_BUFFER_SIZE,
-            small_write_buffer_size: crate::sizing::DEFAULT_SMALL_WRITE_BUFFER_SIZE,
+            large_write_buffer_size,
+            small_write_buffer_size,
         }
     }
 
--- a/components/indexes/rocksdb/src/sizing.rs
+++ b/components/indexes/rocksdb/src/sizing.rs
@@ -22,30 +22,37 @@
 //! the memtable bloom filters on the point-lookup CFs are allocated at 2% of
 //! `write_buffer_size` (~1.3 MiB each at the 64 MiB default, ~5.2 MiB per
 //! query across the four such CFs), and any CF that accumulates more than the
-//! arena's 2 KiB inline block allocates a full arena block. Sizing both
-//! explicitly cuts the touched per-query floor roughly 5x and halves the
-//! memtable ceiling under load, at throughput parity (see issue #692 for
-//! measurements).
+//! arena's 2 KiB inline block allocates a full arena block. At the default
+//! 128 MiB shared budget, sizing both explicitly cuts the touched per-query
+//! floor roughly 5x and halves the memtable ceiling under load, at throughput
+//! parity (see issue #692 for measurements). Larger budgets trade some of
+//! those savings for fewer self-triggered flushes.
 //!
-//! Two buffer sizes, assigned by per-source-change byte volume:
+//! Two buffer sizes are derived from the shared write-buffer manager's budget
+//! and assigned by per-source-change byte volume. The large tier is one eighth
+//! of the budget, clamped to 16-64 MiB; the small tier is half the large tier.
+//! The default 128 MiB budget therefore preserves the original 16/8 MiB sizes,
+//! while a larger budget lets busy column families build larger memtables
+//! before self-triggering a flush:
 //!
-//! - **Large (16 MiB)**: `elements` (full element blob per change), `inbound` /
-//!   `outbound` (adjacency entries per relation change), `values`
-//!   (aggregation accumulators), `archive` (a full element version appended
-//!   per update when enabled), `outbox` and `live_results` (a serialized
-//!   result diff / row set update per result-producing change).
-//! - **Small (8 MiB)**: everything else: same write *frequency* in some cases
-//!   (`slots` is written per element upsert) but tiny values, or genuinely
-//!   quiet CFs (`metadata`, `fqueue`, `findex`, `stream_state`, `partial`,
-//!   `sorted-sets`, and the unused `default` CF).
+//! - **Large (16-64 MiB)**: `elements` (full element blob per change),
+//!   `inbound` / `outbound` (adjacency entries per relation change), `values`
+//!   (aggregation accumulators), `archive` (a full element version appended per
+//!   update when enabled), `outbox` and `live_results` (a serialized result
+//!   diff / row set update per result-producing change).
+//! - **Small (8-32 MiB)**: everything else: same write *frequency* in some
+//!   cases (`slots` is written per element upsert) but tiny values, or
+//!   genuinely quiet CFs (`metadata`, `fqueue`, `findex`, `stream_state`,
+//!   `partial`, `sorted-sets`, and the unused `default` CF).
 //!
 //! `arena_block_size` is derived, not configurable: `write_buffer_size / 64`,
 //! clamped to [64 KiB, 1 MiB]. Left unset, RocksDB sanitizes it to
 //! `min(1 MiB, write_buffer_size / 8)`, which puts a 1 MiB floor under every
-//! CF that sees any real write; the derived value keeps barely-active CFs at
-//! 128-256 KiB instead. Allocations larger than a quarter block (e.g. the
-//! memtable bloom filters) bypass the block pool as exact-size allocations,
-//! so small blocks waste nothing.
+//! CF that sees any real write. At the default budget, the derived value keeps
+//! barely-active CFs at 128-256 KiB instead; at the maximum tier sizes those
+//! blocks grow to 512 KiB for small CFs and 1 MiB for large CFs. Allocations
+//! larger than a quarter block (e.g. the memtable bloom filters) bypass the
+//! block pool as exact-size allocations, so small blocks waste nothing.
 //!
 //! Shared cache/table configuration and the flushed-memtable history bound are
 //! applied by [`crate::cf_options`] before this module overlays the per-CF
@@ -56,13 +63,18 @@ use rocksdb::{ColumnFamilyDescriptor, Options};
 
 use crate::RocksIndexOptions;
 
-/// Default `write_buffer_size` for column families with high byte volume per
-/// source change.
-pub(crate) const DEFAULT_LARGE_WRITE_BUFFER_SIZE: usize = 16 * 1024 * 1024;
+const MIB: usize = 1024 * 1024;
 
-/// Default `write_buffer_size` for the remaining column families: quiet ones,
-/// and ones written often but with tiny values.
-pub(crate) const DEFAULT_SMALL_WRITE_BUFFER_SIZE: usize = 8 * 1024 * 1024;
+/// Floor for high-volume column families, preserving the original #692 size
+/// when the shared write-buffer budget is 128 MiB or smaller.
+const MIN_LARGE_WRITE_BUFFER_SIZE: usize = 16 * MIB;
+
+/// Ceiling for high-volume column families, limiting flush bursts and the
+/// eagerly allocated memtable bloom filters that scale with this value.
+const MAX_LARGE_WRITE_BUFFER_SIZE: usize = 64 * MIB;
+
+const LARGE_BUFFER_BUDGET_DIVISOR: usize = 8;
+const SMALL_BUFFER_DIVISOR: usize = 2;
 
 /// Column families that get the large write buffer. Everything not listed here
 /// (including the mandatory `default` CF) gets the small one. Names must
@@ -78,6 +90,15 @@ const LARGE_BUFFER_CFS: &[&str] = &[
     "outbox",
     "live_results",
 ];
+
+/// Derive the large and small per-CF write-buffer tiers from the aggregate
+/// write-buffer budget shared across query databases.
+pub(crate) fn write_buffer_sizes_for_budget(write_buffer_budget: usize) -> (usize, usize) {
+    let large_write_buffer_size = (write_buffer_budget / LARGE_BUFFER_BUDGET_DIVISOR)
+        .clamp(MIN_LARGE_WRITE_BUFFER_SIZE, MAX_LARGE_WRITE_BUFFER_SIZE);
+    let small_write_buffer_size = large_write_buffer_size / SMALL_BUFFER_DIVISOR;
+    (large_write_buffer_size, small_write_buffer_size)
+}
 
 /// Derive `arena_block_size` from a write buffer size:
 /// `write_buffer_size / 64`, clamped to [64 KiB, 1 MiB].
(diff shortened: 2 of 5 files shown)


ruokun-niu (approved):

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -79,6 +91,15 @@ const LARGE_BUFFER_CFS: &[&str] = &[
...
 
+/// Derive the large and small per-CF write-buffer tiers from the aggregate
+/// write-buffer budget shared across query databases.
+pub(crate) fn write_buffer_sizes_for_budget(write_buffer_budget: usize) -> (usize, usize) {
+    let large_write_buffer_size = (write_buffer_budget / LARGE_BUFFER_BUDGET_DIVISOR)
+        .clamp(MIN_LARGE_WRITE_BUFFER_SIZE, MAX_LARGE_WRITE_BUFFER_SIZE);
+    let small_write_buffer_size = large_write_buffer_size / SMALL_BUFFER_DIVISOR;
+    (large_write_buffer_size, small_write_buffer_size)
+}
+
```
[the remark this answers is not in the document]
amansinghoriginal: These are thresholds, not reservations. Most CFs are idle, and the shared WriteBufferManager intentionally arbitrates aggregate usage. Dividing by every CF would defeat scaling for active CFs.

components/indexes/rocksdb/src/options.rs
```diff
@@ -25,13 +25,22 @@ pub struct RocksIndexOptions {
...
     /// Create options with the memory resources used to open and maintain the index.
+    ///
+    /// Per-column-family write-buffer sizes are captured from the shared
+    /// manager's current budget and remain fixed for the lifetime of these
+    /// options.
     pub fn new(archive_enabled: bool, direct_io: bool, memory_budget: RocksDbMemoryBudget) -> Self {
+        let (large_write_buffer_size, small_write_buffer_size) =
+            crate::sizing::write_buffer_sizes_for_budget(
+                memory_budget.write_buffer_manager().get_buffer_size(),
+            );
```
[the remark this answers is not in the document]
amansinghoriginal: Write load changes usage, not get_buffer_size(), so the proposed test would not exercise snapshot behavior. Recreation coverage already verifies captured sizes are retained.

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -79,6 +91,15 @@ const LARGE_BUFFER_CFS: &[&str] = &[
     "live_results",
 ];
 
+/// Derive the large and small per-CF write-buffer tiers from the aggregate
+/// write-buffer budget shared across query databases.
+pub(crate) fn write_buffer_sizes_for_budget(write_buffer_budget: usize) -> (usize, usize) {
```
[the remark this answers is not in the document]
amansinghoriginal: The adjacent module documentation already explains the divisor, bounds, and rationale. Repeating it here would duplicate the policy.

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -116,16 +137,37 @@ pub(crate) fn descriptor(
...
+        assert_eq!(
+            write_buffer_sizes_for_budget(256 * MIB),
+            (32 * MIB, 16 * MIB)
+        );
+        assert_eq!(
+            write_buffer_sizes_for_budget(512 * MIB),
+            (64 * MIB, 32 * MIB)
+        );
+        assert_eq!(
+            write_buffer_sizes_for_budget(1024 * MIB),
```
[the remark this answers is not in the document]
amansinghoriginal: Calculation is in bytes, so MiB budgets preserve fractional MiB exactly. Truncation is at most seven bytes. Existing tests cover both clamps and pass-through sizing.

components/indexes/rocksdb/src/options.rs
```diff
@@ -25,13 +25,22 @@ pub struct RocksIndexOptions {
 
 impl RocksIndexOptions {
     /// Create options with the memory resources used to open and maintain the index.
+    ///
```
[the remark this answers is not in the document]
amansinghoriginal: Updated, thanks.

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -79,6 +91,15 @@ const LARGE_BUFFER_CFS: &[&str] = &[
     "live_results",
 ];
 
+/// Derive the large and small per-CF write-buffer tiers from the aggregate
+/// write-buffer budget shared across query databases.
+pub(crate) fn write_buffer_sizes_for_budget(write_buffer_budget: usize) -> (usize, usize) {
+    let large_write_buffer_size = (write_buffer_budget / LARGE_BUFFER_BUDGET_DIVISOR)
+        .clamp(MIN_LARGE_WRITE_BUFFER_SIZE, MAX_LARGE_WRITE_BUFFER_SIZE);
+    let small_write_buffer_size = large_write_buffer_size / SMALL_BUFFER_DIVISOR;
```
danielgerlag, line 106: I know you already answered the bot on this, and I agree these are thresholds rather than reservations. The concern I still have is about *which* trigger fires.

At a 512 MiB write budget this becomes `64/32`, so one query DB has `7×64 + 8×32 = 704 MiB` of per-CF ceilings against a 512 MiB shared manager — before a second query is opened. In that shape, busy CFs are less likely to self-flush and more likely to wait for the manager’s global path. That is workable, but it is a different policy than “a larger budget buys bigger self-triggered flushes.”

Would you be open to either:
- keeping `/8` but documenting the multi-query / multi-CF overshoot and showing a multi-query measurement, or
- sizing the large tier from a smaller busy-set (for example a handful of hot CFs) so self-flush still happens inside the budget?

I do not think we should divide by every CF — that would defeat scaling for the active ones, as you said.

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -22,30 +22,37 @@
...
-//! - **Small (8 MiB)**: everything else: same write *frequency* in some cases
-//!   (`slots` is written per element upsert) but tiny values, or genuinely
-//!   quiet CFs (`metadata`, `fqueue`, `findex`, `stream_state`, `partial`,
-//!   `sorted-sets`, and the unused `default` CF).
+//! - **Large (16-64 MiB)**: `elements` (full element blob per change),
+//!   `inbound` / `outbound` (adjacency entries per relation change), `values`
+//!   (aggregation accumulators), `archive` (a full element version appended per
+//!   update when enabled), `outbox` and `live_results` (a serialized result
+//!   diff / row set update per result-producing change).
+//! - **Small (8-32 MiB)**: everything else: same write *frequency* in some
```
danielgerlag, line 44: The small-tier writeup here is convincing that these CFs are high *frequency* with tiny values, or genuinely quiet. That makes lockstep scaling (`small = large / 2`) the part I am least sure about.

Growing `elements` to 64 MiB can produce larger SSTs. Growing `slots` / `metadata` / `default` to 32 MiB mostly keeps more tiny keys in the skiplist, and it also grows the eager 2% memtable bloom on `slots` and `metadata`. If the goal is “busy column families flush less often,” could the small tier stay at the #692 8 MiB floor unless we have data that scaling it helps?

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -22,30 +22,37 @@
 //! the memtable bloom filters on the point-lookup CFs are allocated at 2% of
 //! `write_buffer_size` (~1.3 MiB each at the 64 MiB default, ~5.2 MiB per
```
danielgerlag: Small doc nit, but it is now a bit misleading: 64 MiB is the *cap*, not the default, and two of the four point-lookup CFs (`slots`, `metadata`) are small-tier.

At the real default that is closer to `0.02 × (16 + 8 + 16 + 8) = 0.96 MiB` of eager blooms, and at the cap about `0.02 × (64 + 32 + 64 + 32) = 3.84 MiB`, not 5.2. Worth updating so the floor we are trading away is the one operators will actually see.

components/indexes/rocksdb/src/sizing.rs
```diff
@@ -56,13 +63,18 @@ use rocksdb::{ColumnFamilyDescriptor, Options};
...
-pub(crate) const DEFAULT_SMALL_WRITE_BUFFER_SIZE: usize = 8 * 1024 * 1024;
+/// Floor for high-volume column families, preserving the original #692 size
+/// when the shared write-buffer budget is 128 MiB or smaller.
+const MIN_LARGE_WRITE_BUFFER_SIZE: usize = 16 * MIB;
+
+/// Ceiling for high-volume column families, limiting flush bursts and the
+/// eagerly allocated memtable bloom filters that scale with this value.
+const MAX_LARGE_WRITE_BUFFER_SIZE: usize = 64 * MIB;
+
+const LARGE_BUFFER_BUDGET_DIVISOR: usize = 8;
```
danielgerlag, line 83: Would you mind tying this to `DEFAULT_WRITE_BUFFER_BUDGET_BYTES` so the “16 MiB at the 128 MiB default” invariant cannot drift? Something as small as:

```rust
const _: () = assert!(
    crate::DEFAULT_WRITE_BUFFER_BUDGET_BYTES / LARGE_BUFFER_BUDGET_DIVISOR
        == MIN_LARGE_WRITE_BUFFER_SIZE
);
```

No need to over-engineer it — just a compile-time breadcrumb if the default budget ever moves.

components/indexes/rocksdb/tests/cf_sizing_tests.rs
```diff
@@ -123,51 +122,68 @@ fn effective_sizes_match_the_tier_policy_on_every_cf() {
...
-    );
-    assert_eq!(
-        sizes["stream_state"],
-        (
-            SMALL_WRITE_BUFFER_SIZE,
-            SMALL_WRITE_BUFFER_SIZE / 64,
-            HISTORY_BYTES
-        )
-    );
+    assert_tier_sizes(&sizes, 32 * MIB as u64, 16 * MIB as u64);
```
danielgerlag, line 173: This assertion is correct, but `512 * MIB` here is a *total* budget (`with_memory_budget_bytes` halves it), so the derived tiers are 32/16 rather than 64/32. That is easy to misread later, especially next to the unit test that feeds `512 * MIB` *as a write budget* and expects 64/32.

Could we derive the expected sizes from `provider.memory_budget().write_buffer_manager().get_buffer_size()` (or construct the budget with `RocksDbMemoryBudget::new` so the write limit is explicit)? That would also document the total-vs-write split for the next person who touches this.

danielgerlag (changes requested):
Thanks for the careful work on this, Aman. The capture-once wiring, the default-budget compatibility with #692, and the OPTIONS-file coverage are all in good shape. I want to request a few changes before we merge, because I think the policy and the validation still leave operators — and a multi-query fleet — in a slightly different place than the PR description suggests.

**What I like**
- Deriving sizes once on `RocksIndexOptions::new` is the right freeze given `TransactionDB` has no `set_options_cf`.
- The default 128 MiB write budget stays at 16/8, so unconfigured deployments do not change.
- Recreate still goes through `sized()` / `descriptor()`, so we will not silently fall back to RocksDB's 64 MiB default.

**What I would like to see before approving**

1. **The `/8` model versus a shared budget.** I take your point that these are thresholds, not reservations, and that `WriteBufferManager` is the designed backstop. That said, at a 512 MiB *write* budget a single DB already has `7×64 + 8×32 = 704 MiB` of per-CF ceilings — before a second query shares the same manager. In that regime the manager, not `write_buffer_size`, becomes the flush trigger, which undercuts the “fewer self-triggered flushes” goal. Could we either (a) document that tradeoff clearly and show a multi-query measurement, or (b) size the large tier so a typical busy set of CFs still self-flushes inside the budget?

2. **Small-tier scaling.** Small CFs are high-frequency / tiny-value or idle. Growing them to 32 MiB does not buy bigger SSTs the way `elements` does; it mostly keeps more tiny keys in the skiplist. I would prefer the small tier stay at 8 MiB unless we have data that scaling it helps.

3. **Validation.** The 512 MiB run going 9,434 → 8,779 ev/s with 12 → 1 flush is honest, but one flush on 600k events means we did not observe SST size or compaction under the new policy. #710 asked for SST distributions, and `query-perf` still hardcodes `RocksDbMemoryBudget::default()`, so the table is not reproducible from this branch. A longer run (large tier flushes more than once), the exact command, and ideally a multi-query shape closer to #692 would make the tradeoff much easier to accept.

4. **Operator coupling.** Raising `memoryBudgetBytes` now also changes flush granularity. That is a real behavior change for anyone who already grew the #634 budget. Either keep a small override (`with_write_buffer_sizes`) or call this out explicitly — and consider a config-version note, since `config_version` is still `1.1.0`.

None of this is “the wiring is wrong.” It is “let’s make sure the policy we ship is the one we meant.” Happy to talk through (a) vs (b) on the divisor if that would help.

components/indexes/rocksdb/src/sizing.rs
amansinghoriginal: Updated the figures.

components/indexes/rocksdb/src/sizing.rs
amansinghoriginal: Existing tests already protect this invariant end to end: they verify the default write budget, the 128 MiB to 16/8 MiB derivation, and the effective OPTIONS values for every CF. A compile-time assertion would duplicate that coverage and unnecessarily make todays default-to-floor relationship permanent.

components/indexes/rocksdb/tests/cf_sizing_tests.rs
amansinghoriginal: Clarified the budget split.

components/indexes/rocksdb/src/sizing.rs
amansinghoriginal: Documented the tradeoff.

components/indexes/rocksdb/src/sizing.rs
amansinghoriginal: Keeping `small = large / 2`. #692 established and measured the 2:1 tier policy on a 20-query workload, and #710 explicitly carries that relationship forward while preserving the validated 8 MiB floor and bounding growth at 32 MiB. Fixing the small tier at 8 MiB would introduce a different, unmeasured policy; we can revisit it if comparative data shows it performs better.

danielgerlag:
## Local query-perf follow-up

I ran the missing measurements locally against this branch’s sizing (release `query-perf`, seed 42, `ROCKS_WRITE_BUFFER_BUDGET_BYTES` set so the write budget is explicit). `query-perf` on main still hardcodes the default budget, so this used a local-only env hook.

The 600k averaging numbers line up with the table in the PR description, so I think we were looking at `single_path_averaging_projection`:

| Events | Write budget | ev/s | Flushes (`Flushing memtable` in LOG) | Peak RSS | Largest flush SST |
|---:|---:|---:|---:|---:|---:|
| 600k | 128 MiB | 9,372 | 12 | 69 MiB | 293 KB |
| 600k | 512 MiB | 9,436 | 1 | 137 MiB | 295 KB |
| 2.4M | 128 MiB | 9,635 | 49 | 76 MiB | 293 KB |
| 2.4M | 512 MiB | 9,090 | 12 | 184 MiB | 294 KB |

PR table for comparison: 9,434 / 12 / 72 MiB vs 8,779 / 1 / 145 MiB.

What I take from this:

- The 12 → 1 flush drop at 600k is real. Extending to 2.4M scales it to 49 → 12, so the larger budget does flush less often.
- Every flush is logged as `Flushing memtable with next log file` (WAL rotation). Payloads stay in a tight band: elements ~152–294 KB, slots ~50–105 KB, `values` ~3 KB. We never saw a 16 MiB or 64 MiB SST, even at 2.4M / 64 MiB tiers.
- After the 2.4M / 512 MiB run the DB dir was ~627 MiB with ~800 KB of SST. Almost all of the bytes were still in the WAL.
- Throughput is essentially flat at 600k on this machine, and ~6% slower at 2.4M with the larger budget (9,635 → 9,090 ev/s).

So the advertised “fewer flushes” effect is visible, but it does not look like “busy CFs build larger memtables and emit larger SSTs.” `max_total_wal_size` is left at 0, so the implicit WAL limit grows with `write_buffer_size` and we rotate less often. That is a useful observation, just a different one than the PR text.

I still think a multi-query run would be the other missing datapoint, but I did not have a harness for that today. Happy to compare notes if you already have SST histograms from a different scenario.

Commands (after the local budget hook):

```bash
ROCKS_WRITE_BUFFER_BUDGET_BYTES=134217728 cargo run --release -p query-perf --   -s single_path_averaging_projection -e rocksdb -r rocksdb -i 600000 --seed 42
ROCKS_WRITE_BUFFER_BUDGET_BYTES=536870912 cargo run --release -p query-perf --   -s single_path_averaging_projection -e rocksdb -r rocksdb -i 2400000 --seed 42
```

danielgerlag (approved):

amansinghoriginal:
> ## Local query-perf follow-up
> I ran the missing measurements locally against this branch’s sizing (release `query-perf`, seed 42, `ROCKS_WRITE_BUFFER_BUDGET_BYTES` set so the write budget is explicit). `query-perf` on main still hardcodes the default budget, so this used a local-only env hook.

Thanks for sharing these findings. My logs showed  Write Buffer Full  at approximately the configured 16/8 and 64 MiB thresholds. However, the findings shared above do reveal a real tradeoff:
• Larger memtables retain WAL files longer.
• Disk usage and RSS increase.
• This workload showed no throughput improvement.

Created a follow up issue to investigate this later - https://github.com/drasi-project/drasi-core/issues/782

PR #813, opened 2026-08-31
feat: allow Publish Plugins workflow to publish a single plugin

### LINKED ISSUE #776: Allow Publish Plugins workflow to publish a single plugin
### Area for Improvement

Plugin publish pipeline: `.github/workflows/publish-plugins.yml` and `xtask` `build-plugins` / `publish-plugins`.

These always discover and operate on **every** crate with the `dynamic-plugin` feature. There is no way to build or publish one plugin.

### Observed behavior

`Publish Plugins` is a `workflow_dispatch` workflow and already accepts `ref`, `pre_release`, `tag`, `registry`, `sign`, `dry_run`, and `skip_visibility`.

It does **not** accept a plugin selector. The jobs run:

```text
cargo run -p xtask -- build-plugins --release --target ...
cargo run -p xtask -- publish-plugins --release --target ... --registry ...
```

`xtask` has no `--plugin` / `--only` filter. `discover_dynamic_plugins()` includes every `dynamic-plugin` crate (`source/*`, `reaction/*`, `bootstrap/*`, etc.).

To preview a new plugin from a feature branch (for example `source/otel` on `danielgerlag-otel-source-plan`) you must either:

- republish **all** plugins from that branch, which can overwrite GHCR tags for unrelated plugins if `pre_release` / `tag` is empty, or
- build and push the one cdylib by hand.

A pre-release label avoids clobbering stable tags, but you still pay for an 8-target matrix over every plugin.

### Desired behavior

A `workflow_dispatch` path that publishes **one** plugin from a chosen git ref, without merging to `main` and without rebuilding/republishing the rest.

Suggested UX:

1. New optional input on **Publish Plugins**, e.g. `plugin` (`source/otel`, `drasi-source-otel`, or crate name).
2. Empty `plugin` keeps today’s “publish all” behavior.
3. Matching `--plugin <id>` (or equivalent) on `xtask build-plugins` and `xtask publish-plugins`.
4. `ref` + `pre_release` / `tag` keep working so a branch preview can land as e.g. `ghcr.io/drasi-project/source/otel:0.1.0-dev.1`.
5. Visibility step only touches the selected package (plus directory if still required).
6. Invalid / unknown plugin id fails the job with a clear error, not a silent no-op.

### Proposed Fix

- Add a filter in `xtask` after `discover_dynamic_plugins()` (crate name `drasi-source-otel` and/or `type/kind` `source/otel`).
- Thread that filter through `build-plugins` and `publish-plugins`.
- Add a `plugin` input to `.github/workflows/publish-plugins.yml` and pass it into both xtask invocations.
- Restrict the set-public-visibility loop to the selected package when `plugin` is set.
- Document the input in the workflow description and `xtask` help.

### Drasi Version

drasi-core `main` / current `Publish Plugins` workflow. Not a runtime version issue.

### Operating system

N/A (CI / GHCR publish)

### Additional context

Triggered while trying to publish the OTel source plugin from a feature branch for use with drasi-server 0.2.2, without merging PR #750 or republishing every existing plugin.

Related:

- Workflow: `.github/workflows/publish-plugins.yml`
- Discovery / CLI: `xtask/src/main.rs` (`build-plugins`, `publish-plugins`, `list-plugins`)
- Existing branch checkout input: `ref`
- Draft plugin: `components/sources/otel` (`drasi-source-otel`)

swagatobauri:
# Description

Currently, the `Publish Plugins` workflow always discovers and builds **every** dynamic plugin in the workspace. There is no way to target a single plugin. This becomes a problem when previewing a new plugin from a feature branch (e.g., `source/otel`) — you either have to republish all plugins (risking overwriting stable GHCR tags) or manually push the binary by hand.

This PR solves that by adding an optional `plugin` selector to both the `xtask` CLI and the GitHub Actions workflow.

### What changed

**`xtask/src/main.rs`**
- Added `--plugin <id>` flag support to `build-plugins`, `publish-plugins`, and `list-plugins` commands.
- The filter accepts either the **crate name** (e.g. `drasi-source-otel`) or the **type/kind format** (e.g. `source/otel`).
- If a filter is given but matches nothing, the command **fails immediately with a clear error** — no silent no-ops.
- If no filter is given, all plugins are discovered and processed exactly as before.

**`.github/workflows/publish-plugins.yml`**
- Added a new optional `plugin` string input to both `workflow_dispatch` and `workflow_call`.
- When set, the value is passed as `--plugin` to the `build-plugins`, `publish-plugins`, and `list-plugins` xtask invocations.
- The `set-public-visibility` step is also scoped to only the selected plugin, preventing unnecessary visibility changes to unrelated packages.
- **Default is empty string** — all existing runs (nightly, release-plz, manual without the field set) behave exactly as they did before. Zero breaking changes.

### Example usage

To preview only the OTel source plugin from a feature branch:

1. Go to **Actions → Publish Plugins → Run workflow**
2. Set `plugin` = `source/otel`
3. Set `pre_release` = `dev.1`
4. Set `ref` = your feature branch name

This will build and push **only** `ghcr.io/drasi-project/source/otel:0.1.0-dev.1` without touching any other plugin.

## Type of change

Fixes #776

### DIFF
--- a/.github/workflows/publish-plugins.yml
+++ b/.github/workflows/publish-plugins.yml
@@ -226,6 +226,8 @@ jobs:
 
       - name: Build plugins
         shell: bash
+        env:
+          INPUT_PLUGIN: ${{ inputs.plugin }}
         run: |
           # For -gnu targets, append the glibc-version suffix so cargo-zigbuild
           # pins the floor (e.g. x86_64-unknown-linux-gnu.2.28). xtask strips the
@@ -245,12 +247,14 @@ jobs:
             echo "${TOOLCHAIN_ENV}=${PWD}/ci/cmake/librdkafka-features.cmake" >> "$GITHUB_ENV"
             export "${TOOLCHAIN_ENV}=${PWD}/ci/cmake/librdkafka-features.cmake"
           fi
-          
-          ARGS=(build-plugins --release --target "$BUILD_TARGET")
-          if [ -n "${{ inputs.plugin }}" ]; then
-            ARGS+=(--plugin "${{ inputs.plugin }}")
+          if [ -n "$INPUT_PLUGIN" ] && ! [[ "$INPUT_PLUGIN" =~ ^[A-Za-z0-9._/-]+$ ]]; then
+            echo "Invalid plugin format: $INPUT_PLUGIN" >&2; exit 1
           fi
-          
+          ARGS=(build-plugins --release --target "$BUILD_TARGET")
+          if [ -n "$INPUT_PLUGIN" ]; then
+            ARGS+=(--plugin "$INPUT_PLUGIN")
+          fi
+
           cargo run -p xtask -- "${ARGS[@]}"
 
       - name: Verify glibc floor (-gnu plugins)
@@ -309,6 +313,7 @@ jobs:
           INPUT_TAG: ${{ inputs.tag }}
           INPUT_PRE_RELEASE: ${{ inputs.pre_release }}
           INPUT_SIGN: ${{ inputs.sign }}
+          INPUT_PLUGIN: ${{ inputs.plugin }}
           TARGET: ${{ matrix.target }}
           ARCH_SUFFIX: ${{ matrix.arch_suffix }}
         run: |
@@ -323,6 +328,9 @@ jobs:
           if ! [[ "$INPUT_REGISTRY" =~ ^[A-Za-z0-9._/:-]+$ ]]; then
             echo "Invalid registry format: $INPUT_REGISTRY" >&2; exit 1
           fi
+          if [ -n "$INPUT_PLUGIN" ] && ! [[ "$INPUT_PLUGIN" =~ ^[A-Za-z0-9._/-]+$ ]]; then
+            echo "Invalid plugin format: $INPUT_PLUGIN" >&2; exit 1
+          fi
           # Build the args as an array so each element is passed verbatim (no
           # word-splitting or globbing).
           ARGS=(--release --target "$TARGET" --registry "$INPUT_REGISTRY" --arch-suffix "$ARCH_SUFFIX")
@@ -335,8 +343,8 @@ jobs:
           elif [ -n "$INPUT_PRE_RELEASE" ]; then
             ARGS+=(--pre-release "$INPUT_PRE_RELEASE")
           fi
-          if [ -n "${{ inputs.plugin }}" ]; then
-            ARGS+=(--plugin "${{ inputs.plugin }}")
+          if [ -n "$INPUT_PLUGIN" ]; then
+            ARGS+=(--plugin "$INPUT_PLUGIN")
           fi
           cargo run -p xtask -- publish-plugins "${ARGS[@]}"
 
@@ -347,6 +355,7 @@ jobs:
           INPUT_REGISTRY: ${{ inputs.registry }}
           INPUT_TAG: ${{ inputs.tag }}
           INPUT_PRE_RELEASE: ${{ inputs.pre_release }}
+          INPUT_PLUGIN: ${{ inputs.plugin }}
           TARGET: ${{ matrix.target }}
           ARCH_SUFFIX: ${{ matrix.arch_suffix }}
         run: |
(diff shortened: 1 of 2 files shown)


danielgerlag (commented):
line-resolution probe

xtask/src/main.rs
```diff
@@ -201,6 +193,23 @@ fn discover_dynamic_plugins() -> DiscoveryResult {
...
+            eprintln!("Error: no plugin matched filter '{filter}'");
+            std::process::exit(1);
+        }
+    }
+
+    let plugin_names: BTreeSet<String> = plugins
+        .iter()
+        .map(|p| p.package.name.clone())
+        .collect();
+    let build_batches = plugin_build_batches(&metadata.packages, &plugin_names);
```
danielgerlag, line 207: This will fail to compile: `metadata.packages` was already moved by `.into_iter()` a few lines above, so this borrow is a use-after-move.

Could we keep `packages` as a slice until after `plugin_build_batches`, then consume it? `Package` already derives `Clone`, so iterating with `.iter()` and cloning into `PluginInfo` would work too:

```rust
let mut plugins: Vec<PluginInfo> = metadata
    .packages
    .iter()
    .filter(|p| is_dynamic_plugin(p))
    .filter_map(|p| {
        let (plugin_type, kind) = parse_plugin_type_kind(&p.name)?;
        Some(PluginInfo {
            package: p.clone(),
            plugin_type,
            kind,
        })
    })
    .collect();
// filter, then:
let plugin_names: BTreeSet<String> = plugins.iter().map(|p| p.package.name.clone()).collect();
let build_batches = plugin_build_batches(&metadata.packages, &plugin_names);
```

Fork CI is still `action_required` on this SHA, so this wouldn't have been caught automatically. A local `cargo check -p xtask` should confirm.

.github/workflows/publish-plugins.yml
```diff
@@ -235,7 +245,13 @@ jobs:
             echo "${TOOLCHAIN_ENV}=${PWD}/ci/cmake/librdkafka-features.cmake" >> "$GITHUB_ENV"
             export "${TOOLCHAIN_ENV}=${PWD}/ci/cmake/librdkafka-features.cmake"
           fi
-          cargo run -p xtask -- build-plugins --release --target "$BUILD_TARGET"
+          
+          ARGS=(build-plugins --release --target "$BUILD_TARGET")
+          if [ -n "${{ inputs.plugin }}" ]; then
+            ARGS+=(--plugin "${{ inputs.plugin }}")
```
danielgerlag: Could we pass `plugin` the same way this file already handles `tag` / `pre_release` / `registry` — env var + charset check, then `"$INPUT_PLUGIN"`?

`${{ inputs.plugin }}` is interpolated into the script body here, and again in both publish steps plus `set-public-visibility`. A value with quotes or `$(...)` could change the command.

Something like:

```bash
env:
  INPUT_PLUGIN: ${{ inputs.plugin }}
run: |
  if [ -n "$INPUT_PLUGIN" ] && ! [[ "$INPUT_PLUGIN" =~ ^[A-Za-z0-9._/-]+$ ]]; then
    echo "Invalid plugin format: $INPUT_PLUGIN" >&2; exit 1
  fi
  ...
  if [ -n "$INPUT_PLUGIN" ]; then
    ARGS+=(--plugin "$INPUT_PLUGIN")
  fi
```

.github/workflows/publish-plugins.yml
```diff
@@ -409,7 +431,11 @@ jobs:
           # Set each plugin package to public. list-plugins prints plugin lines
           # as "  <type>/<kind> v<version> (<path>)"; match only those so the
           # indented "SDK: ..." summary line is excluded.
-          PLUGINS=$(cargo run -p xtask -- list-plugins \
+          ARGS=(list-plugins)
+          if [ -n "${{ inputs.plugin }}" ]; then
+            ARGS+=(--plugin "${{ inputs.plugin }}")
```
danielgerlag: Two related issues on this job:

1. Same interpolation of `${{ inputs.plugin }}` as in the build/publish steps — please use `INPUT_PLUGIN` here too.
2. This job still checks out the triggering ref (no `ref: ${{ inputs.ref }}`), while `build-and-publish` honors `inputs.ref`. The documented preview path (`plugin=source/otel` + `ref=<feature branch>`) will run `list-plugins --plugin` against main, fail if the plugin isn't there yet, and skip marking the new GHCR package public.

Worth adding `ref: ${{ inputs.ref }}` to this job's checkout so visibility uses the same tree as the build.

danielgerlag (changes requested):
Thanks for this — the overall shape matches #776 well: optional `--plugin` / workflow `plugin` input, fail-fast on an unknown id, and an empty value keeps the publish-all path for nightly and release-plz.

A few things I'd like to see before merge:

1. **Compile error** in `discover_dynamic_plugins` (`packages` is moved, then borrowed again).
2. **`${{ inputs.plugin }}` is interpolated into bash** in four places. The rest of this workflow already passes free-form inputs via env + a charset check.
3. **`set-public-visibility` does not check out `inputs.ref`**, so the documented feature-branch preview path will fail for a plugin that isn't on main yet.

Inline comments have the details. Happy to re-review once those are in.

.github/workflows/publish-plugins.yml
swagatobauri: Great catch, thanks! I've updated it to use env: INPUT_PLUGIN with a charset check across all four steps.

.github/workflows/publish-plugins.yml
swagatobauri: Done! Both fixed—added ref: ${{ inputs.ref }} to the checkout and switched to using INPUT_PLUGIN.

xtask/src/main.rs
swagatobauri: Thanks for catching the use-after-move! I've updated it to use .iter() and p.clone() exactly as you suggested so the slice stays available.

Issue #824, opened 2026-09-01
Reject durable reactions subscribed to volatile queries


danielgerlag:
### Steps to reproduce

1. Build `DrasiLib` with an in-memory / volatile query index and a state store that reports `is_durable() == true`.
2. Add a reaction with `is_durable() == true`.
3. Call `start_reaction`.

Reproduced by PR https://github.com/drasi-project/drasi-core/pull/745, test `test_durable_reaction_rejects_volatile_query` in `lib/tests/reaction_recovery_e2e.rs`.

### Observed behavior

The reaction starts. `ReactionManager::validate_startup_config` checks durable store presence and archetype/policy pairs, but not query persistence. A durable checkpoint cannot be recovered if the query outbox and snapshot die with the process.

### Desired behavior

Startup rejects `is_durable = true` against a volatile query, per design [00 compatibility matrix](https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Reaction-Checkpoints/00-reaction-recovery-overview.md) and [02 §3](https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Reaction-Checkpoints/02-checkpointing-and-orchestration.md).

- `start_reaction` / `add_reaction`+auto-start returns an error.
- Reaction status is not `Running`.
- Lifecycle metric for this rejection is recorded (same family as durable-on-volatile-store).
- Check runs every start, including after query storage backend changes.

Part of #818. Original miss: #400 listed four compatibility rules and implemented three.

### Workaround

N/A

### Drasi Version

Current `drasi-core` main. Failing test is on PR #745.

### Operating system

N/A

### Additional context

**Required tests:**

1. **`test_durable_reaction_rejects_volatile_query` (PR #745)** — durable reaction + memory query + durable state store → start error, status not Running.

2. **Durable reaction + persistent query + durable store** — start succeeds (positive control).

3. **Durable reaction + persistent query + volatile/missing store** — still rejected (existing rule; keep a regression).

4. **Volatile reaction + volatile query** — still allowed (at-most-once).

5. **Multi-query reaction** — `is_durable = true` and *any* subscribed query is volatile → reject the whole reaction.

6. **Backend change across restart** — reaction started against RocksDB, query later recreated as memory-backed; next `start_reaction` must fail.

Assert the error message names the reaction id and that the query is volatile / non-durable.

### Would you like to support us?

N/A

Issue #819, opened 2026-09-01
Fresh trigger reactions must not replay retained outbox history


danielgerlag:
### Steps to reproduce

1. Start a query with `outbox_capacity >= 2` and produce two results (sequences 1 and 2).
2. Add a **new** reaction with `needs_snapshot_on_fresh_start = false` (trigger / event archetype) and no prior checkpoint.
3. Observe deliveries before any further source change, then insert one more row.

Reproduced by PR https://github.com/drasi-project/drasi-core/pull/745, test `test_fresh_trigger_does_not_replay_retained_history` in `lib/tests/reaction_recovery_e2e.rs`.

### Observed behavior

`ReactionManager::handle_fresh_start` calls `fetch_outbox(0)` and enqueues every retained outbox entry. A new HTTP/webhook-style reaction fires side effects for historical query results.

This was added so source-replay results produced before subscribe are not skipped. It violates the trigger fresh-start flow in design [00 §6](https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Reaction-Checkpoints/00-reaction-recovery-overview.md) and [02 §3 / §5](https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Reaction-Checkpoints/02-checkpointing-and-orchestration.md): skip snapshot, set checkpoint to the query's **current** sequence, open the live gate.

Same-process tests in `reaction_recovery_e2e.rs` never added a trigger to a query that already had outbox data, so they stayed green.

### Desired behavior

On fresh start (`no checkpoint`) when `needs_snapshot_on_fresh_start == false`:

- Do not enqueue retained outbox history.
- Persist checkpoint `(current_sequence, config_hash)`.
- Deliver only results with `sequence > current_sequence`.
- Results produced during *this* startup after subscribe (true live) are delivered.
- State-sync reactions (`needs_snapshot_on_fresh_start == true`) still `fetch_snapshot` and must not use this trigger path.

If source replay can emit results before the reaction subscribes, solve that with the bootstrap gate / subscribe-before-bootstrap ordering, not by replaying the entire ring buffer to a brand-new trigger.

Part of #818. Original miss: #400 flowchart.

### Workaround

N/A

### Drasi Version

Current `drasi-core` main. Failing test is on PR #745.

### Operating system

N/A

### Additional context

**Required tests:**

[...]

Issue #818, opened 2026-09-01
Complete remaining durable reaction recovery work


danielgerlag:
### Overview of feature request

Finish the Query-to-Reaction durability contract from #393 and the [Reaction Checkpoints](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints) design.

#395–#408 are closed, but the closed work did not deliver process-crash recovery. Same-process `stop_reaction` / `start_reaction` tests against in-memory query state are green. Destroying the process and reconstructing a new `DrasiLib` from RocksDB/Garnet query storage plus a durable reaction store is not.

Draft PR https://github.com/drasi-project/drasi-core/pull/745 (`test: add reaction recovery conformance coverage`) already fails in three ways:

1. Persistent query output state is not rehydrated after restart (`rocksdb_redb_full_process_reconstruction`).
2. Fresh trigger reactions replay retained history (`test_fresh_trigger_does_not_replay_retained_history`).
3. Durable reactions can start against volatile queries (`test_durable_reaction_rejects_volatile_query`).

#743 remains the overall conformance-test umbrella. This epic is the remaining *implementation* so those tests can pass. Do not close #393, #743, or this epic while PR #745 is red.

Related existing issues (do not duplicate):

- #743 — durable reaction recovery conformance tests
- #775 — `QueryOutputState` resets to sequence 0 after restart
- #384 — aggregation identity rows leak into live results
- #586 / #564 — recovery policy is inert for non-durable reactions

### Acceptance criteria

- [ ] All child issues in this epic are closed.
- [ ] PR #745 is green, or its tests are merged equivalent and green on RocksDB + Redb.
- [ ] #743 acceptance criteria that this epic owns are satisfied: process reconstruction, atomic commit, sequence HWM, snapshot/outbox after restart, downtime catch-up, checkpoint-after-side-effect, reconfigure/delete wipe, durable-on-volatile rejection.
- [ ] A new `DrasiLib` built against the same persistent query index and durable reaction store continues result sequences at `N+1`, restores snapshot and outbox, and delivers retained results at least once.
- [ ] Failures between persistence stages cannot advance a source checkpoint or result sequence without the corresponding query output.
- [ ] Garnet is either covered by the same reconstruction tests or explicitly deferred in #743 with a follow-up issue.

### Child issues

Implementation children (filed under this epic):

1. Hydrate `QueryOutputState` from persistent storage before processing after restart.
[...]

Issue #402, opened 2026-04-24
Resumable Reactions 08 - Persistent storage traits & transaction integration


amansinghoriginal:
## Background

Part of the [Reaction Recovery](https://github.com/drasi-project/drasi-core/issues/393) initiative. See [design docs](https://github.com/drasi-project/design-documents/tree/main/drasi-lib/Reaction-Checkpoints), specifically doc 01 section 3, section 4, and section 7.

**This issue depends on Source Recovery work** for the CheckpointWriter trait and outer transaction wrapping.

## Work

- Define `OutboxWriter` trait in lib: `append(sequence, result)`, `read_all()` (startup only), `clear()`
- Define `LiveResultsWriter` trait in lib: `apply_diffs(diffs)`, `read_snapshot()` (startup only), `clear()`
- Extend `CheckpointWriter` with `stage_result_sequence(sequence)`
- Extend the outer transaction scope: after core returns from `process_source_change`/`process_due_futures`, when `persistence_enabled` is true and diffs are non-empty, add `stage_result_sequence`, `outbox_writer.append`, and `live_results_writer.apply_diffs` inside the transaction before commit
- Hydrate `QueryOutputState` from persistent storage on crash recovery startup: read `live_results` CF, `outbox` CF, and `stream_state` result sequence

## Dependencies

- Source Recovery #367 (CheckpointWriter trait & implementations)
- Source Recovery #370 (Query processor checkpoint integration)

PR #830, opened 2026-09-01
fix: persist query output atomically with index writes

### LINKED ISSUE #822: Persist query output atomically with index writes and source checkpoints
### Overview of feature request

Make result sequence, outbox append, and live-result mutations commit in the same outer transaction as core index writes and source checkpoints.

#402 required this. PR #459 closed it with best-effort writes **after** in-memory apply and **outside** the index transaction. Current code in `lib/src/queries/manager.rs`:

```text
// Persist to outbox and live results writers if available (best-effort).
// These writes are NOT transactional with the index updates — on crash between
// index commit and outbox write, reactions will re-read from checkpoint sequence.
```

That comment is false if sequences restart or if a source checkpoint advances without output. Design [01 §4](https://github.com/drasi-project/design-documents/blob/main/drasi-lib/Reaction-Checkpoints/01-query-outbox-and-sequencing.md) requires:

```text
session_control.begin()
  core index writes
  stage_checkpoint(src, seq)          // process_source_change only
  if diffs non-empty:
    stage_result_sequence(++last_result_seq)
    outbox_writer.append(...)
    live_results_writer.apply_diffs(...)
session_control.commit()
then update QueryOutputState
then dispatch
```

`process_due_futures` uses the same wrapping, without a source checkpoint.

On rollback, durable index, checkpoints, result sequence, outbox, and live results must all be unchanged. In-memory output must not advance past durable state.

Part of #818. Blocking for #743 atomic-commit and fault-injection ACs. Failing reconstruction in PR https://github.com/drasi-project/drasi-core/pull/745 cannot be considered complete until this is true.

### Acceptance criteria

- [ ] For persistent queries, non-empty diffs stage result sequence, outbox, and live results inside `session_control` **before** commit, together with index writes and (for source changes) the source checkpoint.
- [ ] Volatile queries still bypass durable output I/O.
- [ ] Commit failure / rollback leaves source checkpoint, result sequence, outbox, live results, and graph index unchanged.
- [ ] In-memory `QueryOutputState` and live dispatch happen only after successful commit.
- [ ] `write_result_sequence` is not a separate post-hoc write that can succeed when outbox or live results failed, or fail after they succeeded.
- [ ] Both `process_source_change` and `process_due_futures` are covered.
- [ ] Required tests below are in CI on RocksDB; Garnet unless deferred in #743.

### Required tests

Use a test `CheckpointWriter` / `OutboxWriter` / `LiveResultsWriter` / `SessionControl` that can inject failure at a named stage. PR #745 explicitly deferred these; add them in `lib-integration-tests` (preferred) or `lib` with a real RocksDB index.

For each injected failure, restart a **new** `DrasiLib` on the same store and assert durable state:

1. **Fail after index writes, before source checkpoint stage** — source checkpoint must not advance; output sequence/outbox/live results must not contain the event; on restart the source event is reprocessed.

2. **Fail after source checkpoint stage, before outbox append** — commit must not succeed. Source checkpoint and output must both be absent for that event.

3. **Fail after outbox append, before live_results apply** — commit must not succeed. Restart must not expose an outbox entry whose row is missing from the snapshot (or vice versa).

4. **Fail after live_results apply, before result-sequence write** — if these are no longer separate writes, this case collapses into commit atomicity. If they remain separate, restart must not advertise a sequence whose outbox/snapshot are missing.

5. **Fail after durable commit, before in-memory `QueryOutputState` update / dispatch** — restart hydrates the committed output; the result is not lost; at-least-once dispatch to the reaction is allowed; duplicate sequences must be deduped by checkpoint.

6. **Fail during `process_due_futures` output persist** — same atomicity without a source checkpoint. Time-triggered diffs cannot persist in the index without outbox/live results, or the reverse.

7. **Happy path after the above harness** — one source change produces one committed tuple: source checkpoint `S`, result sequence `N`, outbox key `N`, live row set matching snapshot at `N`.

Do not accept log-and-continue `warn!` on persist failure as fulfilling this issue.

### Additional context

Depends on child "hydrate QueryOutputState" for restart assertions to be meaningful. Original miss: #402 / PR #459.

N/A

danielgerlag:
Fixes #822. Related #743, #818, #825, #826.

## Problem

#402 required staging result sequence, outbox append, and live-result mutations in the same outer transaction as core index writes and source checkpoints. PR #459 closed that with best-effort writes **after** in-memory apply and **outside** the index transaction:

```
// These writes are NOT transactional with the index updates — on crash between
// index commit and outbox write, reactions will re-read from checkpoint sequence.
```

That is false if sequences restart or a source checkpoint advances without output.

## Change

Persistent queries now stage durable output **inside** `session_control` **before** commit:

1. Core index writes
2. `stage_checkpoint(src, seq)` (source changes only)
3. If diffs are non-empty: `stage_result_sequence`, `outbox.append`, `live_results.apply_mutations`
4. `session_control.commit()`
5. Then in-memory `QueryOutputState` update and dispatch

RocksDB and Garnet outbox/live-results writers join the active session transaction (same pattern as `stage_checkpoint`). `write_result_sequence` is no longer a separate post-hoc write that can succeed when outbox/live results failed, or fail after they succeeded. Persist failure aborts the transaction; `warn!` and continue is not used.

Volatile queries still bypass durable output I/O.

## Tests

Added `lib-integration-tests/tests/query_output_atomicity.rs` with a fault-injecting `IndexBackendPlugin` wrapping RocksDB (`CheckpointStore` / `OutboxWriter` / `LiveResultsWriter` / `SessionControl`). Each case restarts a new `DrasiLib` on the same store.

**Passed**
- `fail_after_index_before_source_checkpoint_rolls_back`
- `fail_after_source_checkpoint_before_outbox_rolls_back`
- `fail_after_outbox_before_live_results_rolls_back`
- `fail_after_live_results_before_result_sequence_rolls_back`
- `fail_after_commit_before_in_memory_update_hydrates_on_restart`
- `fail_during_process_due_futures_output_persist_rolls_back`
- `happy_path_commits_source_checkpoint_result_sequence_outbox_and_live_rows`
- `rocksdb_redb_full_process_reconstruction`
- `rocksdb_redb_sequence_hwm_and_snapshot_parity`
- `rocksdb_redb_corrupt_output_strict_fails_start`
- `rocksdb_redb_corrupt_output_autoreset_wipes`
- `drasi-index-rocksdb` `outbox_live_results_tests` (15)
- `drasi-lib --lib queries` (183)

**Still failing (later stack layers, not this PR)**
- `test_fresh_trigger_does_not_replay_retained_history` (#819)
- `test_durable_reaction_rejects_volatile_query` (#824)

PR base is `danielgerlag-hydrate-query-output-state` (layer 2 / #826), not `main`.


components/indexes/garnet/src/outbox.rs
```diff
@@ -60,11 +71,21 @@ impl GarnetOutboxWriter {
 impl OutboxWriter for GarnetOutboxWriter {
     async fn append(&self, query_id: &str, sequence: u64, data: &[u8]) -> Result<(), IndexError> {
         let _ = query_id; // query_id is already bound in self
-        let mut con = self.connection.clone();
         let outbox_key = self.outbox_key();
         let data_key = self.data_key();
         let seq_str = sequence.to_string();
 
+        if let Some(session_state) = &self.session_state {
```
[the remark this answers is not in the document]
danielgerlag: Fixed. Added `GarnetSessionState::with_active_buffer` (same idea as RocksDB `with_txn_or_db`) and routed outbox, live-results, and result-sequence staging through it. Direct Redis writes remain the fallback when no session is active.

core/src/interface/checkpoint_store.rs
```diff
@@ -129,14 +129,27 @@ pub trait CheckpointStore: Send + Sync {
...
+    ///
+    /// For volatile backends, applies immediately (same as
+    /// [`write_result_sequence`](Self::write_result_sequence)).
     ///
-    /// Called after outbox and live-results writes succeed, outside any session
-    /// transaction. Records the highest sequence that was durably persisted so
-    /// that recovery can detect outbox gaps (compare with the index's committed
-    /// sequence from `stage_checkpoint`).
+    /// Default: delegates to [`write_result_sequence`](Self::write_result_sequence).
+    async fn stage_result_sequence(&self, query_id: &str, sequence: u64) -> Result<(), IndexError> {
```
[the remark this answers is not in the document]
danielgerlag: Fixed. Removed the default that delegated to `write_result_sequence`. Persistent backends must implement `stage_result_sequence` (RocksDB/Garnet already do). Volatile `InMemoryCheckpointStore` applies immediately, which is correct for that backend.

lib/src/queries/manager.rs
```diff
@@ -335,16 +316,145 @@ async fn dispatch_query_results(
...
+    outbox_writer: &Option<Arc<dyn OutboxWriter>>,
+    live_results_writer: &Option<Arc<dyn LiveResultsWriter>>,
+    checkpoint_store: &Option<Arc<dyn CheckpointStore>>,
+) -> Result<(), IndexError> {
+    if diffs.is_empty() {
+        return Ok(());
+    }
+    if outbox_writer.is_none() && live_results_writer.is_none() {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
danielgerlag: Fixed. The skip guard is now `outbox_writer.is_none() && live_results_writer.is_none() && checkpoint_store.is_none()`, so a present checkpoint store still gets `stage_result_sequence`.

lib/src/queries/manager.rs
```diff
@@ -335,16 +316,145 @@ async fn dispatch_query_results(
...
+
+    let next_seq = output_state.read().await.as_of_sequence().saturating_add(1);
```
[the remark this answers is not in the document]
danielgerlag: Fixed. Staging now records the durable sequence, and post-commit dispatch applies that same value via `apply_committed_sequence` instead of independently incrementing. In-memory `as_of_sequence` still only moves after commit, so a failed txn cannot advance past durable state.

lib/src/queries/manager.rs
```diff
@@ -335,16 +316,145 @@ async fn dispatch_query_results(
...
-        .collect();
+        .collect()
+}
+
+fn output_persist_error(message: impl Into<String>) -> IndexError {
+    IndexError::other(std::io::Error::new(
+        std::io::ErrorKind::Other,
+        message.into(),
+    ))
+}
```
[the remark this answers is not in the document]
danielgerlag: Leaving `std::io::Error::new(ErrorKind::Other, …)` here. `drasi-lib` clippy is configured with MSRV 1.70, and `Error::other` is stable only since 1.74, so `clippy::incompatible_msrv` fails `-D warnings` if we switch.

lib/src/queries/manager.rs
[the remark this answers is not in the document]
danielgerlag: Fixed without advancing in-memory state before commit (that would be the #822 regression). The pre-commit hook now stores the staged sequence, and `dispatch_query_results` applies that exact value after commit. Evaluation+commit remain serialized on `change_lock`; this query’s processor is a single task, so dispatch runs before the next `process_*_with_hook`. A concurrent stager cannot observe a bumped `as_of_sequence` that never committed.

core/src/query/continuous_query.rs
```diff
@@ -138,7 +139,7 @@ impl ContinuousQuery {
         let changes = self.execute_source_middleware(change).await?;
         let result = self.process_changes_inner(changes).await?;
 
-        pre_commit_hook().await?;
+        pre_commit_hook(&result).await?;
```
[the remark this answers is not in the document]
danielgerlag: Disagree. Staging outbox/live results inside the outer txn requires the evaluation diffs, which only exist after `process_changes_inner` and before commit. A parameterless `FnOnce() -> Fut` cannot see those diffs unless core exposes them some other way (still coupling). Passing `&[QueryPartEvaluationContext]` into the pre-commit hook is the intended extension of the existing checkpoint hook so durable output is part of the same session, not a post-commit write.

lib/src/queries/manager.rs
```diff
@@ -335,27 +320,184 @@ async fn dispatch_query_results(
...
+    outbox_writer: &Option<Arc<dyn OutboxWriter>>,
+    live_results_writer: &Option<Arc<dyn LiveResultsWriter>>,
+    checkpoint_store: &Option<Arc<dyn CheckpointStore>>,
+) -> Result<Option<u64>, IndexError> {
+    if diffs.is_empty() {
+        return Ok(None);
+    }
+    if outbox_writer.is_none() && live_results_writer.is_none() && checkpoint_store.is_none() {
+        return Ok(None);
+    }
```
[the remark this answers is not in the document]
danielgerlag: Fixed. `checkpoint_store` is always `Some` because of the in-memory fallback, so the `is_none()` guard never skipped volatile queries. Staging now runs only when there is an outbox writer, a live-results writer, or a **persistent** checkpoint store. Volatile queries return `None` and do not call `stage_result_sequence`.

lib/src/queries/manager.rs
```diff
@@ -335,27 +320,184 @@ async fn dispatch_query_results(
...
+                Ok(())
+            }
+            Err(_) => Err(output_persist_error("staged output sequence slot poisoned")),
+        }
+    }
+
+    fn take(&self) -> Option<u64> {
+        self.inner.lock().ok().and_then(|mut slot| slot.take())
+    }
+}
```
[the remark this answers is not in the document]
danielgerlag: Not changing this. The slot is recorded once in the pre-commit hook and taken once after commit on the same single-task processor — there is no contention, so the mutex cannot block the executor in practice. An atomic would be equivalent style, not a correctness fix.

ruokun-niu (approved):

agentofreality (approved):

lib/src/queries/manager.rs
```diff
@@ -2702,7 +2776,17 @@ impl Query for DrasiQuery {
...
+                                            let seq = stage_durable_query_output(
+                                                &diffs,
+                                                &source_id_for_stage,
+                                                &query_id_for_stage,
+                                                &output_state_for_stage,
+                                                &outbox,
+                                                &live,
+                                                &checkpoint_for_output,
+                                            )
+                                            .await?;
```
amansinghoriginal, line 2859: A staging failure here rolls back the event, but the processing loop just logs the error and moves on. If event 2 fails and event 3 succeeds, the source checkpoint advances to 3. On restart, event 2 gets skipped even though it was never successfully processed.

We should put the query into `Error` and stop processing further events until recovery has reconciled the durable state. That also matters when a commit succeeds but its acknowledgment fails: continuing with the old in-memory sequence can overwrite an already-committed outbox entry.

components/indexes/rocksdb/src/outbox.rs
```diff
@@ -69,27 +71,33 @@ fn make_prefix(query_id: &str) -> Vec<u8> {
...
     async fn append(&self, query_id: &str, sequence: u64, data: &[u8]) -> Result<(), IndexError> {
         let db = self.db.clone();
+        let session_state = self.session_state.clone();
         let key = make_key(query_id, sequence);
         let data = data.to_vec();
 
         task::spawn_blocking(move || {
             let cf = db.cf_handle(OUTBOX_CF).expect("outbox cf not found");
-            db.put_cf(&cf, &key, &data).map_err(IndexError::other)
+            session_state.with_txn_or_db(
```
amansinghoriginal: Canceling the future waiting on `spawn_blocking` doesn't cancel the queued write. If the query stops and its transaction rolls back before this worker runs, `with_txn_or_db` sees no active transaction and takes the direct-write branch. That leaves an outbox entry for an event whose index changes and checkpoint were rolled back.

We need to keep the write tied to the transaction it was scheduled for, so cancellation can't turn it into a standalone write. The live-results writer has the same pattern and needs the same protection.

lib/src/queries/manager.rs
```diff
@@ -2648,7 +2687,33 @@ impl Query for DrasiQuery {
...
+                                        let query_id_for_stage = query_id.clone();
+                                        let staged_seq = StagedOutputSequence::new();
+                                        let staged_seq_for_hook = staged_seq.clone();
+                                        match continuous_query_for_processor
+                                            .process_due_futures_with_hook(move |results| {
+                                                let diffs = evaluation_contexts_to_diffs(results);
+                                                async move {
+                                                    let seq = stage_durable_query_output(
+                                                        &diffs,
+                                                        crate::sources::FUTURE_QUEUE_SOURCE_ID,
```
amansinghoriginal: We're persisting `__future_queue__` as the source here, but live delivery uses `due_result.source_id`. That means the same result sequence reports its original source when delivered live and the internal queue when replayed after restart. Constructing the persisted and live results separately also gives them different timestamps.

We should create one result with the original source identity and timestamp, persist it, and dispatch that same result after commit. A regression test comparing the live result with the result replayed after restart would help keep these paths consistent.

lib-integration-tests/tests/query_output_atomicity.rs
```diff
@@ -0,0 +1,892 @@
...
+                include_reaction: false,
+                captured: Arc::new(RwLock::new(Vec::new())),
+                wal: paths.wal_b.clone(),
+            },
+        )
+        .await?;
+        start_running(&core).await?;
+        // Checkpoint did not advance, so a restarted source would replay. This
+        // in-process fixture has no WAL lock we can reopen; re-insert simulates replay.
+        insert_person(&source, "p1", "Alice").await?;
```
amansinghoriginal: Restarting with a different WAL and manually sending the event again bypasses source recovery. This could pass even if replay from the original WAL is broken. We should reopen the same stores and let recovery deliver the failed event without another `insert_person` call. If file locks prevent that in-process, we can use separate child processes like the existing reconstruction test.

lib-integration-tests/tests/query_output_atomicity.rs
```diff
@@ -0,0 +1,892 @@
...
+        .read_from(QUERY_ID, 0)
+        .await?
+        .into_iter()
+        .map(|(seq, _)| seq)
+        .collect();
+    let live_row_count = created
+        .live_results_writer
+        .as_ref()
+        .context("live results")?
+        .row_count(QUERY_ID)
```
amansinghoriginal: We're checking sequence numbers, outbox keys, and the live-row count, but not the graph contents or actual output values. That leaves room for a rolled-back graph mutation or incorrect row data to go unnoticed. We should inspect the affected graph elements and compare the outbox payloads and snapshot rows against the expected contents.

lib-integration-tests/tests/query_output_atomicity.rs
```diff
@@ -0,0 +1,892 @@
...
+            outbox
+                .results
+                .iter()
+                .map(|result| result.sequence)
+                .collect::<Vec<_>>(),
+            vec![1],
+            "hydrated outbox must contain the committed sequence"
+        );
+        // Live dispatch after hydrate is at-least-once and may be absent for a
+        // non-durable capturing reaction; durable replay is covered by hydrate.
```
amansinghoriginal: This confirms that the committed result is available in the outbox, but not that a reaction receives it after restart. The restarted reaction is non-durable, and its captured results aren't asserted. We should use a durable reaction and checkpoint store here, then assert that the missed result is processed and its checkpoint advances only after the side effect completes.

amansinghoriginal (changes requested):
The overall approach makes sense, but I don't think this is ready to merge yet. I've left the remaining concerns inline.

lib-integration-tests/tests/query_output_atomicity.rs
danielgerlag: Fixed. `inspect_durable` now checks graph membership for `p1` and deserializes outbox/live-result payloads. Rollback asserts the graph has no person and empty payloads; recovery asserts Alice in both outbox and live rows.

components/indexes/rocksdb/src/outbox.rs
danielgerlag: Fixed. Outbox `append` and live-results `apply_mutations` snapshot `has_active_session()` before `spawn_blocking`. If the write was scheduled inside a session, the worker uses `with_txn` only — cancellation/rollback cannot fall through to a standalone DB write.

lib-integration-tests/tests/query_output_atomicity.rs
danielgerlag: Fixed. The after-commit case uses a durable recording reaction and redb state store. After restart we assert the journal records sequence 1 and the reaction checkpoint advances to 1 only after that side effect.

lib/src/queries/manager.rs
danielgerlag: Fixed. Persist/eval failures now set the query to `Error` and break the processor so a later event cannot advance the source checkpoint past a rolled-back one. After-commit ack failures take the same path, so in-memory sequence cannot overwrite a committed outbox entry. The atomicity tests wait for `ComponentStatus::Error` after each injected fault.

lib/src/queries/manager.rs
danielgerlag: Fixed. `process_due_futures_with_hook` now passes the original `source_id`. Staging returns the `QueryResult` that was written to the outbox, and dispatch uses that same value (source id, timestamp, sequence) instead of reconstructing one. Recovery asserts `outbox_source_ids == [people-source]`.

lib-integration-tests/tests/query_output_atomicity.rs
danielgerlag: Fixed. Fault injection and same-WAL restart now run in separate child processes (redb exclusive locks cannot be released in-process). The recover child does not call `insert_person`; the application source replays WAL from sequence 0 when no checkpoint exists.

agentofreality:
Preservation mapping: existing implementation for #822. Historical input-progress/output crash-window observations map here rather than creating duplicate atomicity issues. Tracked by #889/#890.

lib-integration-tests/tests/query_output_atomicity.rs
```diff
@@ -0,0 +1,1352 @@
...
+        "restart must hydrate the committed output"
+    );
+    let records = read_journal(&paths.journal)?;
+    assert!(
+        journal_has_sequence(&records, 1),
+        "durable reaction must process the missed result after restart"
+    );
+    let checkpoint = read_reaction_checkpoint(&paths.state).await?;
+    assert_eq!(
+        checkpoint.sequence, 1,
```
amansinghoriginal, line 1209: This checks the checkpoint after the side effect, but doesn't establish that it stayed unchanged beforehand. When I blocked the replay side effect, the persisted checkpoint still advanced to 1: `handle_outbox_catchup` checkpoints after enqueueing.

Please hold the side effect pending and assert that the checkpoint remains at its previous sequence, then release it and assert advancement. That should expose the existing premature-checkpoint behavior rather than pass despite it.

lib-integration-tests/tests/query_output_atomicity.rs
```diff
@@ -0,0 +1,1352 @@
...
+
+fn person_name_from_diff(diff: &ResultDiff) -> Option<String> {
+    let data = match diff {
+        ResultDiff::Add { data, .. }
+        | ResultDiff::Update { after: data, .. }
+        | ResultDiff::Aggregation { after: data, .. } => data,
+        ResultDiff::Delete { data, .. } => data,
+        ResultDiff::Noop => return None,
+    };
+    data.get("name")
```
amansinghoriginal, line 751: NIT (non-blocking): This is stronger, but the assertions still reduce output to names and graph state to membership. An Add for Alice and a Delete for Alice both satisfy the name check, despite having opposite meanings.

It would be useful to compare the expected diff types, complete row values and row signatures, hydrated snapshot contents, and affected graph properties, rather than just names and existence.

components/sources/application/src/lib.rs
```diff
@@ -660,26 +660,30 @@ impl Source for ApplicationSource {
...
-                    resume_from.len()
-                ));
-            }
+                settings.resume_sequence.unwrap_or(0)
+            };
+            let wal_clone = wal.clone();
+            drop(wal_guard);
+            return self
+                .base
+                .subscribe_with_replay(&settings, wal_clone.as_ref(), resume_seq, "Application")
```
amansinghoriginal: With WAL enabled, every subscription now returns through `subscribe_with_replay`, so fresh queries never reach `subscribe_with_bootstrap`, even with `enable_bootstrap: true` and a configured provider. I reproduced that provider being bypassed.

We need zero-checkpoint WAL recovery without skipping explicitly requested initial bootstrap. Please preserve both paths and cover a fresh WAL-enabled subscription with a bootstrap provider.

lib/src/queries/manager.rs
```diff
@@ -342,144 +329,139 @@ async fn dispatch_query_results(
...
+                "processed_by".to_string(),
+                serde_json::Value::String("drasi-core".to_string()),
+            );
+            meta.insert(
+                "result_count".to_string(),
+                serde_json::Value::Number(diffs.len().into()),
+            );
+            meta
+        },
+        crate::profiling::ProfilingMetadata::new(),
```
amansinghoriginal: Now that live dispatch uses this staged result, creating fresh profiling metadata here discards the source/query timings already collected. Persistent results now have `source_send_ns`, `query_receive_ns`, and `query_core_call_ns` unset.

Please carry the available pre-commit profiling metadata into the staged result so the envelope-parity fix doesn't remove profiling information from live delivery.

amansinghoriginal (changes requested):
Thanks for the updates. Several earlier concerns are resolved, but the remaining correctness gaps and regressions noted inline still need addressing.

lib-integration-tests/tests/query_output_atomicity.rs
danielgerlag: Declining for this PR. Holding the side effect and asserting the reaction checkpoint does not advance is #820 (checkpoint-after-effect), which this stack explicitly deferred. The #822 coverage here is that a post-commit failure still leaves durable query output committed and that a durable reaction processes the missed result after restart. Premature `handle_outbox_catchup` checkpointing belongs in that later layer.

components/sources/application/src/lib.rs
danielgerlag: Fixed. WAL subscribe no longer always takes `subscribe_with_replay`. Checkpointed resume (`resume_from` / `resume_sequence`) still replays; zero-checkpoint recovery with bootstrap off still replays from seq 0; a fresh subscribe with `enable_bootstrap` uses `subscribe_with_bootstrap` even when WAL is enabled. Covered by `test_wal_enabled_fresh_subscribe_still_bootstraps`.

lib-integration-tests/tests/query_output_atomicity.rs
danielgerlag: Declining. This is a non-blocking nit. Rollback vs commit already distinguishes absence vs presence of the Alice row; Add-vs-Delete identity is not a #822 atomicity gap. Happy to tighten payload shape in a follow-up if we want more snapshot-parity coverage.

lib/src/queries/manager.rs
danielgerlag: Fixed. Staging now copies the pre-commit `ProfilingMetadata` into the durable `QueryResult` (`query_receive_ns`, `query_core_call_ns`, and any source timings already present). Dispatch overlays `query_core_return_ns` / `query_send_ns` onto that same envelope so live delivery keeps the timings. Happy-path atomicity test asserts those fields.

amansinghoriginal (approved):
Thanks for addressing the feedback. This looks good to me for the scope of this PR. The separately tracked checkpoint-ordering work remains in drasi-project/drasi-core#820.

PR #832, opened 2026-09-01
fix: fresh trigger reactions must not replay retained outbox history

### LINKED ISSUE #819: Fresh trigger reactions must not replay retained outbox history

danielgerlag:
Fixes #819. Related #743, #818, #825.

## Problem

`ReactionManager::handle_fresh_start` treated a brand-new trigger reaction (`needs_snapshot_on_fresh_start == false`, no checkpoint) like catch-up: it called `fetch_outbox(0)` and enqueued every retained outbox entry. HTTP/webhook-style reactions therefore fired side effects for historical query results.

That replay was added so source-replay results produced before subscribe would not be skipped. It violates the trigger fresh-start flow in Reaction Checkpoints 00 §6 and 02 §3 / §5: skip snapshot, set the checkpoint to the query's **current** sequence, open the live gate.

## Fix

On fresh start with `needs_snapshot_on_fresh_start == false`:

- Do **not** enqueue retained outbox history.
- Persist checkpoint `(current_sequence, config_hash)`.
- Deliver only results with `sequence > current_sequence`.
- True live results produced after subscribe still flow through the bootstrap gate.

State-sync reactions (`needs_snapshot_on_fresh_start == true`) still `fetch_snapshot` and bootstrap. Restart of an existing trigger with checkpoint `N` still catches up `N+1..` via `handle_outbox_catchup`.

If source replay can emit results before the reaction subscribes, that is handled by subscribe-before-bootstrap ordering, not by replaying the entire ring buffer to a brand-new trigger.

## Tests

Ran locally (RocksDB / Redb available):

- `cargo test -p drasi-lib --test reaction_recovery_e2e -- --test-threads=1` — **17 passed**, including:
  - `test_fresh_trigger_does_not_replay_retained_history`
  - `test_reaction_outbox_catchup_on_restart`
  - `test_durable_reaction_rejects_volatile_query` (layer 4)
  - `test_fresh_state_sync_bootstraps_snapshot_and_dedupes_live` (new)
  - `test_trigger_autoreset_is_rejected` (new)
- `cargo test -p drasi-lib reactions::manager -- --test-threads=1` — **16 passed**, including `fresh_trigger_does_not_enqueue_retained_outbox`
- `cargo test --test reaction_recovery_conformance -- --test-threads=1` — **6 passed**, including:
  - `rocksdb_redb_full_process_reconstruction`
  - `rocksdb_redb_fresh_trigger_no_history_replay` (new process-reconstruction variant)
- `cargo test --test query_output_atomicity -- --test-threads=1` — **7 passed**

This is layer 5 of the recovery stack. Base is `danielgerlag-reject-durable-on-volatile` (PR #831), not `main`.


lib/src/reactions/manager.rs
```diff
@@ -551,52 +555,23 @@ impl ReactionManager {
...
         } else {
-            // No snapshot needed — replay any outbox entries produced during this
-            // startup cycle (e.g., from source replay) and record the checkpoint.
-            // Without this replay, results produced before the reaction subscribes
-            // to the broadcast channel would be silently skipped.
+            // Trigger / event reactions must not fire side effects for retained
+            // history. Record the query head and let the live gate deliver only
+            // later sequences. Do not replay `fetch_outbox(0)`.
             metrics.record_fetch_outbox();
             let seq = match query.fetch_outbox(0).await {
```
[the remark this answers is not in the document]
danielgerlag: Agreed this fetch is heavier than a head-sequence read, but it is the same pre-existing `fetch_outbox(0)` pattern already used by AutoSkipGap. Adding `Query::current_sequence()` would be a trait-wide API change outside #819. This PR only stops enqueueing `resp.results`; I am not restoring history replay. A lighter head API can be a follow-up.

lib/src/reactions/manager.rs
```diff
@@ -551,52 +555,23 @@ impl ReactionManager {
             bootstrap_queries.push((query_id.to_string(), query.clone()));
             Ok(cp)
         } else {
-            // No snapshot needed — replay any outbox entries produced during this
-            // startup cycle (e.g., from source replay) and record the checkpoint.
-            // Without this replay, results produced before the reaction subscribes
-            // to the broadcast channel would be silently skipped.
+            // Trigger / event reactions must not fire side effects for retained
+            // history. Record the query head and let the live gate deliver only
+            // later sequences. Do not replay `fetch_outbox(0)`.
```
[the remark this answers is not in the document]
danielgerlag: Updated the comment: we still call `fetch_outbox(0)` for `latest_sequence`, but we do not enqueue any returned entries.

lib/src/reactions/manager.rs
```diff
@@ -551,52 +555,23 @@ impl ReactionManager {
...
-                        if last_ok_seq != resp.latest_sequence {
-                            info!(
-                                "[{reaction_id}] Partial outbox replay for query '{query_id}' — \
-                                 replayed up to seq={last_ok_seq}, latest_seq={}",
-                                resp.latest_sequence
-                            );
-                        }
-                    }
+                    info!(
+                        "[{reaction_id}] Fresh start for query '{query_id}' — \
```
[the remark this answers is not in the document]
danielgerlag: Updated. Logs and comments now talk about “fresh start without snapshot bootstrap” / “skipping retained history” instead of hard-coding “trigger reaction” in the manager.

lib/src/reactions/manager.rs
```diff
@@ -551,52 +555,23 @@ impl ReactionManager {
             bootstrap_queries.push((query_id.to_string(), query.clone()));
             Ok(cp)
         } else {
```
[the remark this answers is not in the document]
danielgerlag: Leaving `needs_snapshot_on_fresh_start()` as the existing Reaction Checkpoints 00/02 discriminator. In that design, `false` is the trigger/event fresh-start path: skip snapshot, checkpoint at current sequence, do not replay retained history. Splitting the trait or adding `FreshStartMode` would be a public-API change outside #819. A future reaction that wants a different combination can be a follow-up; this PR restores the documented mapping rather than inventing a third mode.

lib/tests/reaction_recovery_e2e.rs
```diff
@@ -313,6 +335,16 @@ impl Reaction for RecordingReaction {
...
     }
+
+    async fn bootstrap(&self, context: BootstrapContext) -> Result<()> {
+        let snapshot = context
+            .fetch_snapshot()
+            .await
+            .map_err(|error| anyhow::anyhow!("fetch reaction bootstrap snapshot: {error}"))?;
+        self.bootstrap_count.fetch_add(1, Ordering::SeqCst);
+        *self.bootstrap_as_of.lock().expect("bootstrap_as_of lock") = Some(snapshot.as_of_sequence);
+        Ok(())
```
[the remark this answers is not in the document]
danielgerlag: Switched `bootstrap_as_of` to `AtomicU64` with `u64::MAX` as the unset sentinel, so `bootstrap()` no longer takes a blocking mutex.

lib/src/reactions/manager.rs
```diff
@@ -551,52 +555,24 @@ impl ReactionManager {
...
-            // No snapshot needed — replay any outbox entries produced during this
-            // startup cycle (e.g., from source replay) and record the checkpoint.
-            // Without this replay, results produced before the reaction subscribes
-            // to the broadcast channel would be silently skipped.
+            // Fresh start without snapshot bootstrap must not fire side effects
+            // for retained history. Still call `fetch_outbox(0)` to learn the
+            // current head sequence, but do not enqueue any returned entries.
+            // The live gate then delivers only later sequences.
             metrics.record_fetch_outbox();
             let seq = match query.fetch_outbox(0).await {
```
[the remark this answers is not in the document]
danielgerlag: Fixed without a new `Query` API. Fresh start now calls `fetch_outbox(u64::MAX)`, so `fetch_outbox_after` hits the `after_sequence >= as_of_sequence` path and returns an empty vec. We still persist `latest_sequence` and still do not enqueue retained history. Restart catchup continues to use `fetch_outbox(checkpoint.sequence)`.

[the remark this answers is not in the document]
danielgerlag: Follow-up: `fetch_outbox(u64::MAX)` is intentional, not a new head API. In-process `DrasiQuery` takes the `after_sequence >= as_of_sequence` path and returns an empty vec, so retained entries are not cloned. `latest_sequence` is still `as_of_sequence`. Catchup is unchanged (`fetch_outbox(checkpoint.sequence)`). A dedicated `current_sequence()` can still be a follow-up if we want to drop the magic number.

lib/src/reactions/manager.rs
```diff
@@ -561,52 +565,25 @@ impl ReactionManager {
...
-            // Without this replay, results produced before the reaction subscribes
-            // to the broadcast channel would be silently skipped.
+            // Fresh start without snapshot bootstrap must not fire side effects
+            // for retained history. Ask for entries after `u64::MAX` so
+            // `fetch_outbox` returns only `latest_sequence` without cloning
+            // retained outbox entries. The live gate then delivers later
+            // sequences.
             metrics.record_fetch_outbox();
-            let seq = match query.fetch_outbox(0).await {
+            let seq = match query.fetch_outbox(u64::MAX).await {
```
amansinghoriginal: **This can silently drop live events during startup.** Subscriptions are already wired before this call. If we subscribe at sequence 2 and result 3 arrives before this read, we checkpoint 3 and the forwarder discards that buffered result—even though it arrived after subscription and was never processed. Restarting won’t replay it either. This violates drasi-project/drasi-core#819’s requirement to deliver post-subscription events. The initial cutoff must be coordinated with subscription, rather than taken from a later query head.

lib/src/reactions/manager.rs
```diff
@@ -561,52 +565,25 @@ impl ReactionManager {
...
                     resp.latest_sequence
                 }
                 Err(FetchError::OutboxGap(gap)) => {
                     info!(
-                        "[{reaction_id}] Fresh start for query '{query_id}' — outbox gap, latest_seq={}",
+                        "[{reaction_id}] Fresh start for query '{query_id}' — \
+                         outbox gap on fresh start without snapshot, latest_seq={}",
                         gap.latest_sequence
                     );
                     gap.latest_sequence
```
amansinghoriginal: Regarding the `NotRunning` / `TimedOut` fallback immediately below:

**An unavailable query does not imply sequence 0.** If the query already contains historical results 1 and 2, this fallback persists checkpoint 0 for a fresh trigger. Once the query is running and the reaction restarts, catchup replays those historical results and fires unwanted side effects. Although pre-existing, this still defeats the PR’s no-history guarantee. Please defer or fail startup without writing a checkpoint until the actual query head can be determined.

amansinghoriginal (changes requested):
Requesting changes for the two correctness concerns noted inline.

lib/src/reactions/manager.rs
danielgerlag: Agreed — we no longer fall back to checkpoint 0 when the query is unavailable.

The trigger path no longer calls `fetch_outbox` on fresh start, so the `NotRunning`/`TimedOut` → 0 branch is gone. The cutoff is the subscribe-time head from output state. If that head cannot be determined, startup fails and no checkpoint is written.

lib/src/reactions/manager.rs
danielgerlag: Agreed — sampling the outbox head after subscribe could skip a live result already in the receiver.

Trigger fresh start now uses `QuerySubscriptionResponse.as_of_sequence`, captured from output state before the receiver is attached. Later live results (`sequence > subscribe_as_of`) are delivered; retained history is still not replayed. Catchup from an existing checkpoint is unchanged.

agentofreality:
Preservation mapping: existing implementation for #819 under #818. Keep and enrich this PR rather than creating a replacement. Tracked by #889/#890.

lib/src/queries/manager.rs
```diff
@@ -3039,8 +3039,12 @@ impl Query for DrasiQuery {
             reaction_id, self.base.config.id
         );
 
+        // Sample the head *before* attaching the receiver so a later
+        // `fetch_outbox` cannot raise the skip cutoff past results already
+        // buffered for this subscriber.
+        let as_of_sequence = self.output_state.read().await.as_of_sequence();
         self.base
-            .subscribe(&reaction_id)
+            .subscribe(&reaction_id, as_of_sequence)
```
amansinghoriginal: **Moving the read earlier leaves a race before receiver attachment.** If we read head 2, result 3 can be dispatched before the receiver is attached. We then checkpoint 2, receive result 4, and `Strict` policy detects a missing sequence and stops the reaction. I reproduced this in both Channel and Broadcast modes. Please coordinate head capture and receiver registration with result publication, so neither operation leaves an uncovered window.

lib/src/reactions/manager.rs
```diff
@@ -567,64 +575,21 @@ impl ReactionManager {
...
-                         query not yet running, starting from sequence 0"
-                    );
-                    0
-                }
+            // Fresh start without snapshot bootstrap must not fire side effects
+            // for retained history. Use the sequence captured at subscribe so
+            // results that arrived after the receiver was attached are not
+            // skipped. Do not sample the outbox head later, and do not invent
+            // checkpoint 0 when the head is unknown.
+            let Some(seq) = subscribe_as_of else {
```
amansinghoriginal, line 583: **This guard does not detect an uninitialized query head.** A reconstructed persistent query initially has in-memory sequence 0 before its saved state is loaded, so this path receives `Some(0)` and persists it. With RocksDB already containing results 1 and 2, starting the trigger before the query initializes still creates checkpoint 0; a later reaction restart replays those historical results. Please defer or fail startup without writing a checkpoint until the actual persisted head is known.

amansinghoriginal (changes requested):
Requesting changes for the two remaining correctness concerns noted inline.

lib/src/reactions/manager.rs
danielgerlag: Agreed — sequence 0 from uninitialized in-memory state is not a real head.

`subscribe` now waits until the query is Running so durable hydrate has finished before we capture `as_of_sequence`. If the query never becomes ready, startup fails and no checkpoint is written (including 0). An empty Running query can still legitimately checkpoint 0.

lib/src/queries/manager.rs
danielgerlag: Agreed — sampling the head and then attaching the receiver left a publish window.

`dispatch_query_results` now publishes while still holding `output_state`. `subscribe` waits until the query is Running, then holds the same lock across the head sample and receiver attach. A result is either included in the cutoff or delivered to the new subscriber; Strict should no longer see a hole from this race. Catchup from an existing checkpoint is unchanged.

lib/src/queries/manager.rs
```diff
@@ -606,6 +609,20 @@ async fn dispatch_query_results(
...
+        // Publish while still holding `output_state` so subscribe cannot sample
+        // this head and then attach after this result has already been sent.
+        debug!(
+            "Query '{query_id}' sending {} results to reactions (seq={})",
+            result.results.len(),
+            result.sequence
+        );
+        let dispatchers = dispatchers.read().await;
+        for dispatcher in dispatchers.iter() {
+            if let Err(e) = dispatcher.dispatch_change(result.clone()).await {
```
amansinghoriginal: **Please avoid awaiting bounded sends while holding `output_state`.** When a Channel fills, this await holds the lock until the receiver drains it. A receiver fetching a snapshot or outbox before draining waits for that same lock, creating a deadlock. I reproduced this with capacity 1; explicitly draining the receiver unblocks the snapshot. Please preserve atomic subscription/publication coordination without holding the snapshot-state lock across a potentially blocking send.

lib/src/queries/base.rs
```diff
@@ -135,13 +129,11 @@ impl QueryBase {
...
         let receiver: Box<dyn ChangeReceiver<QueryResult>> = match dispatch_mode {
             DispatchMode::Broadcast => {
-                // For broadcast mode, use the single dispatcher
-                let dispatchers = self.dispatchers.read().await;
-                if let Some(dispatcher) = dispatchers.first() {
-                    dispatcher.create_receiver().await?
-                } else {
-                    return Err(anyhow::anyhow!("No broadcast dispatcher available"));
-                }
+                let dispatcher = crate::channels::UnboundedChangeDispatcher::<QueryResult>::new();
```
amansinghoriginal: **This replaces bounded Broadcast buffering with an unbounded queue for the subscriber’s entire lifetime.** With `dispatch_buffer_capacity=1`, a paused receiver retained all 256 results. The new `pending` queue in `ReactionManager` also bypasses Channel backpressure while bootstrap runs. Slow consumers or lengthy bootstrap can therefore grow memory without limit, regardless of outbox capacity. Please keep both buffering paths bounded, using backpressure or outbox recovery rather than unlimited in-memory retention.

amansinghoriginal (dismissed):
Thanks for addressing the earlier startup findings. Requesting changes for the two new regressions noted inline.

lib/src/queries/manager.rs
danielgerlag: Agreed. Holding `output_state` across a bounded Channel `send().await` can deadlock with snapshot/outbox fetches when the queue is full. We’ll keep subscribe/publish coordination without awaiting a blocking send under that lock.

lib/src/queries/base.rs
danielgerlag: Agreed. Per-subscriber unbounded queues (and the bootstrap `pending` buffer) drop backpressure for the whole subscription lifetime, not just gate wait. We’ll keep both paths bounded and recover via outbox rather than unlimited in-memory retention.

lib/src/queries/manager.rs
danielgerlag: Fixed. `dispatch_query_results` no longer awaits a bounded Channel send while holding `output_state`. Publication under that lock uses `try_dispatch_change` (`try_send` / non-blocking broadcast). Snapshot and outbox fetches can take the same lock without waiting on a full queue.

Subscribe still samples the head and attaches the receiver under `output_state`, so the publish window stays closed. If a Channel is full, the send is dropped rather than blocking; the reaction recovers those sequences from the outbox after the bootstrap gate opens (and again on a live sequence gap).

lib/src/queries/base.rs
danielgerlag: Fixed. Query Broadcast is bounded again: one shared ring created at `QueryBase::new`, subscribers reuse `create_receiver`, and `UnboundedChangeDispatcher` is gone. The forwarder no longer keeps an unbounded `pending` drain while the gate is closed.

Dropped ring entries are recovered from the outbox after the gate opens, and again if live delivery sees a sequence gap. Slow consumers therefore backpressure or lag against the configured buffer/outbox, instead of growing an unlimited in-memory queue for the whole subscription.

lib/src/reactions/manager.rs
```diff
@@ -1189,16 +1156,70 @@ impl ReactionManager {
                     // Track the last forwarded sequence for gap detection.
                     let mut last_forwarded_seq = initial_seq;
 
+                    match query_clone.fetch_outbox(initial_seq).await {
```
amansinghoriginal: Non-blocking follow-up: this can declare an unrecoverable gap even when the live buffer contains everything needed. With checkpoint 2, channel `[3,4]`, and outbox `[4]`, the Strict reaction stops without processing either event. The base branch consumes both normally. If we keep Channel delivery lossless, I’d remove this extra post-gate outbox fetch and drain the live buffer instead, while retaining normal restart catchup.

lib/src/queries/manager.rs
amansinghoriginal: Non-blocking follow-up: a full Channel can still leave its last event unprocessed. I reproduced outbox `[1,2,3,4]` with only `[1,2,3]` processed and the reaction still Running; result 4 was recovered only after publishing 5. For a simpler initial implementation, I’d favor bounded Channel backpressure with publication/subscription coordination separate from `output_state`, rather than live drop-and-replay.

amansinghoriginal (commented):
Non-blocking follow-up notes

lib/src/queries/manager.rs
danielgerlag: Fixed. Sequence assign still happens under `output_state`; Channel `send` happens after that lock is released. A full queue now backpressures the query instead of dropping the tail, so result 4 is delivered without waiting for 5. Snapshot/outbox fetches no longer share a lock with a blocking send.

lib/src/reactions/manager.rs
danielgerlag: Fixed. The forwarder no longer fetches the outbox after the gate opens. It drains the live buffer, so Channel `[3,4]` is delivered even if the outbox has already wrapped to `[4]`. Restart catchup from an existing checkpoint still uses `fetch_outbox` before the forwarder starts. Live sequence-gap recovery is unchanged.

Issue #828, opened 2026-09-01
Migrate all sources to dispatch_event and deprecate dispatch_from_task


ruokun-niu:
Part of #817. Depends on #827.

## Background
Sources can send events two ways today:
- `dispatch_event` (and its batch form): stamps the monotonic sequence and does the correct bookkeeping.
- `dispatch_from_task`: a static shortcut used by sources that dispatch from a background task. It does **not** stamp a sequence.

14 sources use the shortcut. 11 never get a sequence; the other 3 (http, grpc, application) only get one when durability is on. Events without a sequence can't be ordered reliably or de-duplicated. These sources predate the `SourceBase::clone_shared()` pattern and were never updated.

## What to do
Give each background task an **owned** `SourceBase` via `clone_shared()` and call `dispatch_event` / `dispatch_events_batch` on it instead of the static `dispatch_from_task`. Event-building code stays the same; WAL sequences and source positions are preserved (so native-cursor sources like mysql/postgres keep their checkpointing).

Effort varies by source "shape":
- **Shape A (one line):** the struct already holds a `SourceBase` — just change the call. e.g. `mysql`.
- **Shape B (one signature):** the spawn passes `dispatchers` into a single function — hand it a `base` instead. e.g. `http`, `open511`.
- **Shape C (a few signatures):** `dispatchers` threaded through nested helpers — pass `base` down the same path. e.g. `kubernetes`, `ris-live`.

Then:
- Mark `dispatch_from_task` `#[deprecated]` (and remove once no in-tree callers remain).
- Migrate the test mock sources too (`shared-tests/src/mock_source.rs`, `lib/tests/mock_source/mod.rs`, and reaction test mocks).
- Update source-authoring docs/comments that currently point at the shortcut (`components/sources/source-documentation-standards.md`, `components/sources/README.md`) and add one clear "dispatch from a background task" example.

## Sources to migrate (from #817)
Never-stamped (11): cloudflare-radar, dataverse, gtfs-rt, hyperliquid, kubernetes, mock, mysql, open511, platform, ris-live, sqlite.
WAL-only (3): http, grpc, application.

## Dependency / ordering
- **Blocked by #827** for the native-cursor sources: once they start stamping, restart-monotonicity must already be handled by the framework, or migrating them would introduce a restart-drop regression. Volatile and sequence-as-position sources can be migrated independently.

## Acceptance criteria
[...]

Issue #827, opened 2026-09-01
Carry last sequence number back to sources on resubscribe (restart monotonicity)


ruokun-niu:
Part of #817.

## Background
When a source omits a sequence, `dispatch_event` fills one in from a per-source counter (`next_sequence`) that **starts at 1 every process start**. On restart, the query's dedup filter still remembers the last checkpoint (say `500`) and **drops anything ≤ 500**, so freshly-filled sequences `1, 2, 3…` would be wrongly dropped.

Two classes of source behave differently:
- **Sequence-as-position sources** (http, grpc, application, otel): their durable position *is* the framework sequence (`source_position = seq.to_be_bytes()`), so they already restore the counter with `set_next_sequence(head)`.
- **Native-cursor sources** (postgres LSN, kafka offset, mysql binlog, mssql, oracle): their durable position is a store-native cursor — **not** the framework sequence — so they never restore the counter. This is the gap.

## What to do
The query already reads `SourceCheckpoint { sequence, source_position }` but only passes the *position* back to the source (`resume_from`). Pass the sequence back too:

1. Add `resume_sequence: Option<u64>` to `SourceSubscriptionSettings` (next to `resume_from`).
2. Populate it in `QueryManager` from the checkpoint (it already reads `cp.sequence`).
3. In `SourceBase`, on subscribe raise `next_sequence` up to that floor with `fetch_max` (never lowers). With multiple resuming queries, take the **max** floor; per-subscriber position filtering handles the overlap.
4. Add `resume_sequence` to the plugin subscribe FFI (`components/plugin-sdk/src/ffi/vtable_gen.rs`) so **out-of-process** plugin sources receive the floor too. Today the FFI carries only `resume_from` (the position); without this an out-of-process source never sees the floor.

This makes restart-monotonicity a **framework** behavior instead of each source hand-rolling it. It is compatible with the existing `set_next_sequence(head)` restore — whichever floor is higher wins.

## Notes
- **Non-breaking and additive.** It is a no-op until a source actually produces checkpointed sequences, so it is safe to land first.
- **Ordering:** this must land **before** native-cursor sources start getting a framework sequence in #817's migration work — otherwise that migration would *introduce* a restart-drop regression for them.

## Acceptance criteria
- [ ] `SourceSubscriptionSettings` has a `resume_sequence: Option<u64>` field.
- [ ] `QueryManager` populates `resume_sequence` from the persisted checkpoint on resubscribe.
[...]

Issue #817, opened 2026-09-01
Sources that use `dispatch_from_task` don't stamp a monotonic `sequence`


ruokun-niu:
## Summary

Sources that dispatch via `SourceBase::dispatch_from_task` do **not** get a
framework-assigned monotonic `sequence`. This is a gap between the sequence/position
design and its implementation: `dispatch_event` auto-stamps a monotonic sequence
for every event even when durability is disabled, but the `dispatch_from_task`
path skips that stamping. As a result these sources emit events with
`sequence = None` — either always, or whenever durability is disabled.

This leaves those events with no per-source ordinal, which is what the priority
queue needs to order same-timestamp events deterministically (see #814).

There are multiple parts to this issue. Specifically:

1. We should examine what sources are using the outdated `dispatch-from-task` function and replace them with `dispatch_event`
2. We should investigate if it is possible to enforce sources to do a sequence. If this is not possible, we should fill in a sequence number when it is omitted by the Source
3. Think about how we can make the developer experience of new sources easier

Please see the additional comments in this issue for more details.

## Affected sources

Sources are grouped by whether/when they carry a `sequence`.

### ✅ Not affected — always stamped
Dispatch via `dispatch_event` / `dispatch_events_batch` / `dispatch_source_change`
(which calls `dispatch_event`), so the framework always assigns a sequence:
`kafka`, `mssql`, `oracle`, `otel`, `postgres`, `neo4j`, `sui-deepbook`,
`here-traffic`.

### ⚠️ Affected — sequence only when durability is ON (`None` when OFF)
Dispatch via `dispatch_from_task`; set `sequence` from the WAL only:
- `http`
- `grpc`
- `application`

### ❌ Affected — sequence NEVER stamped (`None` regardless of durability)
Dispatch via `dispatch_from_task`; never set `sequence` at all:
- `cloudflare-radar`
- `dataverse`
- `gtfs-rt`
- `hyperliquid`
- `kubernetes`
- `mock`
- `mysql`
- `open511`
- `platform`
- `ris-live`
- `sqlite`

Related: #814 (priority-queue ordering relies on this sequence).

ruokun-niu:
## Research Brief

### Problem statement and scope

[...]

PR #858, opened 2026-09-03
Make the source-event sequence mandatory at the type level

### LINKED ISSUE #829: Make sequence mandatory at the type level (unstamped vs stamped event types)
Part of #817. Follows #828. **Breaking change** — coordinated release.

## Background
After #827 and #828, every in-tree source gets a sequence *by convention* (use `dispatch_event`, the only remaining path). But the wrapper type still carries `sequence: Option<u64>`, so "an event with no sequence" is still representable — a new source author could reintroduce the gap, and all downstream code must keep handling the `None` case.

## What to do
Make the sequence impossible to omit by splitting the type:
- An **unstamped, source-authored** type carrying just the payload (the change/event) — with **no sequence field to set or forget**.
- A **stamped, downstream** type where `sequence: u64` is mandatory (not `Option`), so "no sequence" is unrepresentable past the source boundary.
- `dispatch_event` / `dispatch_events_batch` become the **only** bridge: the author hands in the unstamped payload, the framework stamps it and produces the downstream event. No other constructor exists.
- Remove `Option` handling for `sequence` in all downstream code.
- Add a short source-author template/checklist so the "right way" is the fastest way to start.

## Developer-experience goal
- **One obvious way in:** a new author calls `dispatch_event` with a payload and never touches sequence numbers, ordering, or positions.
- **Can't forget the sequence:** with the shortcut gone (#828) and the field removed from the author-facing type, an unstamped event cannot be built.
- **Simpler downstream code:** `sequence` is always present.

## Breaking-change surface
- The event wrapper type and its serde representation change.
- The plugin ABI/FFI changes → all out-of-process plugins must be rebuilt.
- Schedule as a coordinated/breaking release; this is intentionally the **last** step.

## Acceptance criteria
- [ ] Source authors construct only an unstamped payload type with no `sequence` field.
- [ ] The downstream event type exposes `sequence: u64` (non-optional); no public constructor allows building it without a framework-assigned sequence.
- [ ] `dispatch_event` / `dispatch_events_batch` are the only way to turn an unstamped payload into a stamped downstream event.
- [ ] All downstream `Option<u64>` sequence handling is removed.
- [ ] Plugin SDK/host FFI updated for the new types; a rebuilt out-of-process plugin round-trips sequences correctly.
- [ ] serde round-trip tests updated and passing for the new representation.
- [ ] A source-author template/checklist is added to the docs.
- [ ] `cargo fmt` + `cargo clippy --all-targets --all-features` clean; full workspace tests pass.

ruokun-niu:
# Description

Closes #829. Part of #817; builds on #827 and #828.

Make source-event sequences mandatory while retaining a single `SourceEventWrapper`. Previously, `sequence: Option<u64>` allowed events without a sequence and required downstream consumers to handle that case.

The updated contract guarantees that a sequence is **present**, not that it was assigned by the framework. Sources may supply their own source-local monotonic sequence or allocate one through `SourceBase`.

### Single Event Wrapper
- Retain `SourceEventWrapper` with required `sequence: u64`.
- Require an explicit sequence in `new()` and `with_profiling()`.
- Remove the proposed `SourceEventDraft` / `StampedSourceEvent` split and optional `supplied_sequence`.
- Preserve profiling metadata, opaque source positions, and zero-copy extraction.

### Sequence Allocation and Dispatch
- Expose `SourceBase::next_sequence()` using the existing source-local `AtomicU64` allocator.
- `dispatch_event()` and `dispatch_events_batch()` preserve supplied sequences and advance the fallback counter past them without lowering it.
- `dispatch_source_change()` and `broadcast_control()` allocate and dispatch under the same ordering lock.
- Preserve WAL-head and query-resume sequence-floor restoration.

Explicit-wrapper producers must serialize allocation and dispatch together across concurrent tasks. Atomic allocation guarantees unique numbers, but does not itself guarantee delivery order.

### Source Migration
- WAL-backed application, HTTP, gRPC, and OTEL sources use WAL-assigned sequences when durability is enabled, otherwise the fallback allocator.
- Native-cursor sources retain `source_position` metadata and allocate required event sequences.
- Other sources use explicit allocation or ordered convenience methods.
- Coordinate gRPC unary and streaming producers with a shared lock covering WAL append or allocation through dispatch.

### Downstream Processing
- Remove optional-sequence handling from deduplication, checkpoint staging, and position-handle updates.
- Use `SourceEventWrapper` throughout query queues, subscriptions, and SDK mappings.
- Give `FutureQueueSource` notifications their own required sequence numbers.

### Future-Notification Deduplication Fix
Apply source deduplication only to `SourceEvent::Change`.

An application source named `__future_queue__` can otherwise establish a data checkpoint that suppresses independently numbered `FuturesDue` notifications. Control notifications now bypass data-sequence deduplication; data changes retain their existing checkpoint behavior.

### Plugin FFI and Documentation
- Change `SourceEventPayload.sequence` from `Option<u64>` to `u64`.
- Retain `from_wrapper()` / `into_wrapper()` helpers and reconstruct `SourceEventWrapper` at the receiving boundary.
- Update source-authoring guides, API references, and agent instructions to describe required sequences and concurrent-producer responsibilities.

## Compatibility

This is a breaking Rust API and plugin FFI change. Callers must provide a sequence when constructing events, and dynamically loaded plugins must be rebuilt against the matching SDK.

The allocator is in-memory; allocating or dispatching an event does not persist it. WAL-backed recovery continues to depend on durable append before acknowledgment and replay using persisted WAL sequences.

## Validation
- Compilation passed for all source crates and affected library, SDK, shared-test, integration-test, and reaction test targets.
- Passed 71 source-base tests, 160 query/checkpoint tests, 23 event tests, and eight FFI payload tests.
- Passed application, HTTP, and gRPC sequencing tests, including concurrent unary/streaming gRPC requests with durability enabled and disabled.
- Added a temporal-query regression covering ordinary and `__future_queue__` source IDs after 129 data events. Confirmed it fails with the old dedup placement and passes with the fix.
- Formatting and whitespace checks passed.

The full workspace test suite was not run. Three existing WAL tests in the selected ingress test runs remained ignored.

## Type of Change
- This pull request adds or changes features of Drasi and has an associated issue: #829.


components/sources/mock/src/mock.rs
```diff
@@ -656,7 +650,7 @@ impl MockSource {
     ///
     /// # Returns
     ///
-    /// A boxed receiver that yields [`SourceEventWrapper`](drasi_lib::channels::SourceEventWrapper)
+    /// A boxed receiver that yields [`SourceEventDraft`](drasi_lib::channels::SourceEventDraft)
     /// for each event generated or injected.
     ///
```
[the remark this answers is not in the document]
ruokun-niu: updated

lib/src/sources/tests.rs
```diff
@@ -50,6 +53,7 @@ impl TestMockSource {
             auto_start: true,
             status_handle,
             dispatchers: Arc::new(RwLock::new(Vec::new())),
+            next_sequence: Arc::new(std::sync::atomic::AtomicU64::new(0)),
```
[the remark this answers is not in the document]
ruokun-niu: already addressed

lib/src/sources/future_queue_source.rs
```diff
@@ -163,11 +169,16 @@ impl FutureQueueSource {
...
+                    // Item is due — dispatch FuturesDue signal. This source
+                    // bypasses SourceBase, so it stamps a monotonic sequence
+                    // itself to honor the mandatory-sequence contract.
+                    let draft = SourceEventDraft::new(
                         FUTURE_QUEUE_SOURCE_ID.to_string(),
                         SourceEvent::Control(SourceControl::FuturesDue),
                         timestamp,
                     );
+                    let seq = next_sequence.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+                    let event_wrapper = StampedSourceEvent::stamp(draft, seq);
```
amansinghoriginal: Could we apply source deduplication only in the `SourceEvent::Change` branch? `__future_queue__` is currently accepted as an application source ID, so its data checkpoint can suppress these independently numbered `FuturesDue` notifications.

I reproduced this in a fresh in-memory instance: after 129 source events, a `drasi.trueLater` result appeared with an ordinary source ID but missed the same deadline with `__future_queue__`.

lib/src/channels/events.rs
```diff
@@ -218,42 +218,45 @@ pub enum SourceEvent {
...
-/// Wrapper for source events with metadata
+/// Unstamped, source-authored event payload.
+///
+/// This is the **only** type a source author constructs. It deliberately has
+/// **no `sequence` field**: the framework assigns the monotonic sequence when
+/// the draft is dispatched via [`SourceBase::dispatch_event`] /
+/// [`SourceBase::dispatch_events_batch`], which is the sole bridge that turns a
+/// draft into a downstream [`StampedSourceEvent`]. Because "unstamped" and
+/// "stamped" are distinct types, it is a *compile-time* guarantee that no event
+/// reaches the query side without a framework-assigned sequence.
```
amansinghoriginal: Small wording question: could we describe this as guaranteeing that the sequence is present, rather than framework-assigned? The public `from_ffi_parts()` constructor and mutable `sequence` field still allow callers to supply their own values. I don’t think we need additional enforcement here—just wording that matches what the types guarantee.

amansinghoriginal (changes requested):
Thanks for simplifying the downstream event handling. Could we address the future-notification case below before merging? The other comment is just a non-blocking wording clarification.

ruokun-niu:
NOTE:
- single struct with sequence number being required

lib/src/sources/future_queue_source.rs
ruokun-niu: Good point. I'll fix this

agentofreality (approved):

components/sources/hyperliquid/src/stream.rs
```diff
@@ -432,6 +432,7 @@ pub(crate) async fn dispatch_changes(
             SourceEvent::Change(change),
             chrono::Utc::now(),
             profiling,
+            base.next_sequence(),
         );
 
         if let Err(e) = base.dispatch_event(wrapper).await {
```
amansinghoriginal: Could we use `base.dispatch_source_change(change).await` here? The WebSocket and funding-poll tasks both call this helper, but `next_sequence()` now runs before the dispatch lock. They can dispatch `N+1` before `N`, causing query deduplication to discard a genuine update.

I reproduced this with two concurrent callers. The existing helper keeps sequence allocation and dispatch under the same lock, so it avoids the race without adding another ordering mechanism.

amansinghoriginal (changes requested):
Thanks for addressing the earlier comments. Could we fix the remaining sequencing issue below before merging?

components/sources/hyperliquid/src/stream.rs
ruokun-niu: Good catch. I'll fix this

ruokun-niu:
Hi @amansinghoriginal I have addressed your latest feedback. Please review again when you have the time. Thanks!

amansinghoriginal (approved):
thanks

Issue #791, opened 2026-08-27
HTTP source silently drops events under batch-ingestion backpressure


ruokun-niu:
## Summary

When a producer POSTs source change events to the HTTP source's
`/sources/{id}/events/batch` endpoint faster than the downstream query can
drain them, the source plugin **discards a fraction of the events without
error**. Every batch POST returns `200 OK` and the producer sees zero failures,
but the query emits fewer results than events ingested and then stalls forever.

Loss scales with ingest pressure (~0.19% on a MacBook Pro, ~0.02% on a faster
GitHub-hosted 8-vCPU VM for the identical workload), which is the signature of a
bounded-channel overflow / lossy send under backpressure rather than a fixed
unflushed tail.

## Environment

- `drasi-server` 0.2.2 (release binary).
- HTTP source plugin `source/http` **v0.2.10** (sdk=0.13.0, core=0.2.10,
  commit `c688c39f`), loaded as a cdylib.
- Source starts with:
  `HttpSource using batch channel capacity: 5000 (max_batch_size: 1000 x 5)`
  and
  `AdaptiveBatchConfig { max_batch_size: 1000, min_batch_size: 10, max_wait_time: 100ms, min_wait_time: 1ms, throughput_window: 5s, adaptive_enabled: false }`.
- Reproduced on macOS aarch64 (local) and Linux x86_64 (GitHub Actions).

## Reproduction

Harness:
[`e2e-test-framework/examples/building_comfort/local/perf_sweep`](../e2e-test-framework/examples/building_comfort/local/perf_sweep).
A single per-room Cypher query is run so the expected result count is exact:

```
elements = B + 2*(B*F) + 2*(B*F*R)
expected = B*F*R + CHANGE_COUNT - elements
```

For a 1×3×4 graph @ 100000 events this is **99981**.

Crucially, in this harness the **reaction delivers egress per-result** (no
`{"batch":[...]}` coalescing, no `adaptive` block on the reaction), so egress
batching is *not* involved — the shortfall is entirely on the ingress path.

```bash
cd e2e-test-framework/examples/building_comfort/local/perf_sweep

DRASI_SERVER_BIN="$PWD/.local_run/drasi-server-bin/drasi-server-0.2.2-aarch64-apple-darwin" \
TIMEOUT_SECS=180 \
ADAPTIVE=1 BATCH_EVENTS=1 BATCH_SIZE=1000 BATCH_TIMEOUT_MS=50 \
[...]

PR #860, opened 2026-09-04
fix(queries): order same-timestamp events deterministically via (timestamp, source_rank, sequence) composite key

### LINKED ISSUE #814: Priority queue reorders same-timestamp events, corrupting continuous-query results
# Priority queue reorders same-timestamp events, corrupting continuous-query results

## Summary

Each query has an in-memory priority queue (`lib/src/channels/priority_queue.rs`)
that orders incoming events by **timestamp only**. Timestamps are not unique — a
batched or fast producer stamps many events with the same value — and the
underlying Rust `BinaryHeap` is **not stable**, so equal-timestamp events are
dequeued in an **arbitrary, heap-structure-dependent order**.

Because a continuous query's output depends on the order events are applied,
this reordering **changes query results**. It manifests in two different ways
depending on whether source **persistence (WAL durability)** is enabled — and it
goes wrong in **both** cases:

- **Persistence disabled (default):** events carry no sequence number, so nothing
  is dropped, but the query **evaluates same-timestamp events in the wrong order
  and emits an incorrect / short result set**.
- **Persistence enabled:** the same reordering additionally trips the per-source
  **sequence dedup**, which **silently drops** the out-of-order event.

This is the underlying cause of #791 (HTTP source "losing" events under batch
ingestion). The event is not dropped on the wire — the query produces the wrong
results.

## The trigger (common to both cases)

- A source stamps events with `chrono::Utc::now()` at ingest. Under batch
  ingestion (e.g. HTTP `/events/batch`) many events are stamped inside one clock
  tick, so clusters of events share an identical timestamp.
- Those same-timestamp events land in the query's priority queue together.
- `BinaryHeap` pops equal-timestamp events in an arbitrary order — the queue does
  **not** preserve arrival order for ties.

Sources that ingest one event at a time (e.g. gRPC, paced by network reads)
rarely produce same-timestamp clusters, so they seldom expose this. Any source
that ingests in bulk triggers it reliably.

## Case 1 — persistence disabled (default): wrong / short result set

With durability off, events carry `sequence = None`. The sequence dedup is inert
(`should_skip(None)` returns `false`), so **every event is processed and nothing
is dropped**. The loss comes entirely from the query evaluating same-timestamp
events in the wrong order.

### Worked example

Query: `MATCH (s:Sensor) WHERE s.v > 10 RETURN s.name`
(a sensor is in the result set only while `v > 10`)

Start: `s1` exists with `v = 0` → not in result set → current set `{}`.

Two updates to `s1` arrive in the **same batch** → **same timestamp**, **no
sequence**:

- `U20`: set `v = 20`
- `U5` : set `v = 5`

**Order A — pop `U20` then `U5`:**

| Step | Event | v before→after | Membership | Diff emitted |
|---|---|---|---|---|
| 1 | U20 | 0 → 20 | crosses above 10 → ADD | Add |
| 2 | U5  | 20 → 5 | drops below 10 → REMOVE | Remove |

→ **2 diffs emitted**, final set `{}`

**Order B — pop `U5` then `U20`:**

| Step | Event | v before→after | Membership | Diff emitted |
|---|---|---|---|---|
| 1 | U5  | 0 → 5  | still ≤ 10 → no change | (none / Noop) |
| 2 | U20 | 5 → 20 | crosses above 10 → ADD | Add |

→ **1 diff emitted**, final set `{s1}`

Same two events, same timestamps, no sequence numbers — only the pop order
differs, yet **both the number of results emitted (2 vs 1) and the final result
set (`{}` vs `{s1}`) differ**. A reconcile that expects one delivery per change
sees a shortfall (the missing diff) and, waiting for a result that never comes,
**stalls**.

## Case 2 — persistence enabled: silent drop by sequence dedup

With durability on, each event carries a per-source `sequence` and the query runs
a sequence dedup (`lib/src/queries/sequence_dedup.rs`, `should_skip`): it drops
any event whose sequence is `<=` the last one processed (per-source high-water
mark), to discard post-crash replays.

Events are enqueued in per-source sequence order, but the queue can release a
higher sequence before a lower one when timestamps tie. The dedup then treats the
lower one as an already-seen duplicate and drops it:

```
Queue holds (seq=6, ts=100) and (seq=7, ts=100)   <- same timestamp, same source
Releases seq=7 first (ties are arbitrary) -> high-water = 7
Releases seq=6 next -> "6 <= 7 -> duplicate" -> DROP   <- silently lost event
```

Here the event really is discarded before the query sees it, and the query stalls
waiting for a result that will never be produced.

## Evidence

**Perf run (persistence disabled).** In a `building_comfort` performance run the
`http_adaptive` variant fails to reconcile (records short, run stalls) while
`grpc_adaptive`, `grpc_standard`, and `http_standard` all pass with matching
determinism hashes.

- Run: https://github.com/ruokun-niu/test-infra/actions/runs/33202348567/job/98954697202
- Config confirmed durability **off**: `persistIndex=false, stateStore=false`.

**Probe proving dedup is not the cause here.** Re-running the lossy repro with
`drasi_lib::queries` debug logging: `SequenceDedup::should_skip` fired **0
times** across ~200k debug lines, no drop/capacity/overflow logged, yet the query
emitted fewer results than events ingested (99,803 of 99,981). This rules out the
dedup path for the durability-off case and points at query evaluation on
out-of-order events.

**Reproduced loss counts (durability off):**

| Run | Events | Expected | Delivered | Lost | Loss % |
|---|---:|---:|---:|---:|---:|
| local 20k | 20,000 | 19,981 | 19,942 | 39 | 0.195% |
| local 100k | 100,000 | 99,981 | 99,791 | 190 | 0.190% |
| local 100k (debug probe) | 100,000 | 99,981 | 99,803 | 178 | 0.178% |
| CI 100k (faster host) | 100,000 | 99,981 | 99,961 | 20 | 0.020% |

Loss scales with how far ingestion outruns the consumer (deeper queue → more
same-timestamp events coexisting → more reordering). By contrast `http_standard`
(unbatched, heap depth ≈ 1, no reordering) reaches **exactly 99,981** with a
matching determinism hash on the same seed.

**Workload / config.** `building_comfort`, deterministic seeded generator:
- Graph: 1 building × 3 floors × 4 rooms = **12 rooms**
- Query: `MATCH (r:Room) RETURN elementId(r), r.temperature, r.humidity, r.co2`
- Change events: **100,000**; expected results
  `= B*F*R + CHANGE_COUNT − elements = 12 + 100000 − 31 = 99,981`
- Environment: drasi-server 0.2.2, `source/http` v0.2.10, adaptive batching
  `min_batch_size=10, max_batch_size=1000, wait=50ms`, **durability off**.

**Minimal repro** (single query, egress per-result → isolates the ingress path):

```bash
cd e2e-test-framework/examples/building_comfort/local/perf_sweep
TIMEOUT_SECS=180 ADAPTIVE=1 BATCH_EVENTS=1 BATCH_SIZE=1000 BATCH_TIMEOUT_MS=50 \
CHANGE_COUNT=100000 BUILDINGS=1 FLOORS=3 ROOMS=4 AUTO_DRASI_SERVER=1 ./run_sweep.sh
```
Symptom: `building-comfort=99791 (waiting)` plateaus and the run times out.

**Minimal confirmation.** A focused test feeding a continuous query two
same-`timestamp`, same-source updates (with `sequence = None`) in the two
possible orders produced two different result sets (0 rows vs 1 row) — matching
the worked example above.

## Expected vs actual

- **Expected:** among same-timestamp events, the queue applies a well-defined,
  deterministic order so query results do not depend on heap layout, and (with
  persistence on) the dedup never sees a sequence go backwards.
- **Actual:** same-timestamp events are dequeued in arbitrary order, so query
  results are non-deterministic and can be short/incorrect (persistence off) or
  silently dropped by dedup (persistence on).

## Goals

- Same-timestamp events must be processed in a well-defined order so that query
  results are correct and deterministic, with **and** without persistence.
- No record loss under HTTP batch ingestion in either persistence mode.
- Add coverage for **multi-source join** queries, where events from multiple
  sources share timestamps and interleave in the queue.

## Code references

- PQ ordering (timestamp only, no defined order for ties):
  https://github.com/drasi-project/drasi-core/blob/main/lib/src/channels/priority_queue.rs
- Sequence dedup (`should_skip`; `None` always passes; `seq <= high-water` drops):
  https://github.com/drasi-project/drasi-core/blob/main/lib/src/queries/sequence_dedup.rs
- Query processor where dequeue + dedup interact:
  https://github.com/drasi-project/drasi-core/blob/main/lib/src/queries/manager.rs
- HTTP source: `sequence` set only when a WAL exists (else `None`):
  https://github.com/drasi-project/drasi-core/blob/main/components/sources/http/src/lib.rs

## Notes

This issue documents the behavior and root cause only. The solution (how to
define and enforce a deterministic order for equal-timestamp events) is out of
scope here and to be designed separately. Related: #791.

ruokun-niu:
# Description

Fixes #814.

ach query's in-memory priority queue (`lib/src/channels/priority_queue.rs`)
ordered incoming events by **timestamp only**. Timestamps are not unique (a
batched or fast producer stamps many events with the same value), and the
underlying `BinaryHeap` is not stable, so equal-timestamp events were dequeued
in an **arbitrary, heap-dependent order**. Because a continuous query's output
depends on the order events are applied, this reordering produced **incorrect /
short result sets** (persistence off) or **silently dropped events** via the
sequence dedup (persistence on). This was the root cause of #791.

This PR replaces the timestamp-only ordering with a composite key:

```
(timestamp, source_rank, sequence)
```

- **timestamp** — event time (as before), the primary component.
- **source_rank** — the source's rank within the query; breaks ties between
  events from **different** sources. Inert for single-source queries.
- **sequence** — the per-source monotonic sequence (mandatory as of #817);
  breaks ties between events from the **same** source, so they dequeue in
  emission order and the dedup high-water never moves backward.

With this, same-timestamp events are applied in a well-defined, deterministic
order — with and without persistence — and no events are lost under HTTP batch
ingestion.

## How `source_rank` is resolved

A source's rank is derived at subscribe time and stamped onto each event by the
per-source forwarder at enqueue time (the shared `Arc<StampedSourceEvent>` is
broadcast across queries, so rank must be query-local and applied at enqueue).

Ranking supports **both** explicit and implicit ordering:

- New optional `priority: Option<i64>` field on `SourceSubscriptionConfig`
  (lower = evaluated first on a tie).
- Sources without a `priority` fall back to their **position in the query's
  `sources` list**.
- Effective ordering is `(priority.unwrap_or(list_index), list_index)`, so every
  source gets a **distinct, stable rank** and the list index always breaks
  remaining ties.

The synthetic `FutureQueueSource` (not in `config.sources`) is ranked last
(`u32::MAX`) so its re-evaluation signals resolve deterministically after real
sources on a tie.

## Public API additions

- `Sequenced` trait in `channels/events.rs` (implemented for
  `StampedSourceEvent` and `QueryResult`); the priority queue now requires
  `Timestamped + Sequenced`.
- `SourceSubscriptionConfig.priority: Option<i64>` (serde `default`, skipped when
  `None`).
- `Query::from_source_with_priority(source_id, priority)` builder method.

## Backward compatibility

- **Configs unchanged:** `priority` is optional and defaults to `None`
  (implicit ranking by declared order) — existing YAML/JSON configs deserialize
  unchanged, and behavior for distinct-timestamp events is identical to before.
- **Builder unchanged:** existing `from_source(...)` keeps working.
- **Behavioral change is the intended fix:** only *same-timestamp* events change
  ordering — from arbitrary to deterministic.
- Source-level note: `SourceSubscriptionConfig` gains a required Rust field, so
  external code constructing it via a bare struct literal must add
  `priority: None`. Serde and builder callers are unaffected.

## Tests

Added deterministic coverage (no reliance on heap layout or wall-clock ties):

- Priority-queue ordering: rank-only ties, sequence-only ties, full composite
  precedence, and multi-source same-timestamp interleave (issue Scenarios 1–4).
- `compute_source_ranks` unit tests: implicit fallback, explicit reordering,
  explicit+implicit combination with index tiebreak, always-distinct ranks,
  single-source.
- Builder + serde round-trip tests, including a backward-compat test that a
  subscription **without** `priority` deserializes to `None`.

## Type of change

Fixes: #issue_number


lib/src/sources/tests.rs
```diff
@@ -50,6 +53,7 @@ impl TestMockSource {
             auto_start: true,
             status_handle,
             dispatchers: Arc::new(RwLock::new(Vec::new())),
+            next_sequence: Arc::new(std::sync::atomic::AtomicU64::new(0)),
```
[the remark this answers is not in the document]
ruokun-niu: This is addressed in anotehr PR

lib/src/channels/priority_queue.rs
```diff
@@ -21,39 +21,63 @@ use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering as AtomicOrdering};
...
+    T: Timestamped + Sequenced + Clone + Send + Sync + 'static,
 {
-    fn new(event: Arc<T>) -> Self {
-        Self { event }
+    fn new(event: Arc<T>, source_rank: u32) -> Self {
+        Self { event, source_rank }
+    }
+
+    /// The composite ordering key: `(timestamp, source_rank, sequence)`.
+    fn order_key(&self) -> (chrono::DateTime<chrono::Utc>, u32, u64) {
```
danielgerlag, line 56: This key is easy to read, which I like.

One thing that jumped out: dispatch still bumps timestamps by 1ns so same-source events never actually share a timestamp. That means we now have two ways of keeping same-source order — unique-ified timestamps, and sequence in this key — and dispatch comments still talk about a timestamp-only queue.

Would it be cleaner to pick one? If we trust `(timestamp, rank, sequence)`, we could leave event time as the source stamped it. If we keep the bump, sequence in the heap is mostly belt and suspenders for the same-source case.

lib/src/channels/priority_queue.rs
```diff
@@ -181,8 +207,15 @@ where
...
+    ///
+    /// `source_rank` is the precomputed rank of the producing source within the
+    /// owning query, derived from `config.sources` and any explicit per-source
+    /// priority. It is used as a tie-breaker between events from different
+    /// sources that share a timestamp. Pass `0` for single-source queues where
+    /// the rank is inert.
+    ///
     /// Returns true if enqueued, false if queue is at capacity
-    pub async fn enqueue(&self, event: Arc<T>) -> bool {
+    pub async fn enqueue(&self, event: Arc<T>, source_rank: u32) -> bool {
```
danielgerlag, line 218: Putting `source_rank` on the generic enqueue is easy to get wrong — a lot of call sites (tests, reactions) just pass `0`.

Rank is query-local anyway, and the forwarder already closes over it. Would it be simpler to keep rank out of this signature? Even a tiny wrapper type would make “forgot the rank” a compile error instead of a silent `0`.

lib/src/config/schema.rs
```diff
@@ -130,6 +144,21 @@ pub struct SourceSubscriptionConfig {
...
+    /// **position in the `sources` list** (implicit rank).
+    ///
+    /// Explicit and implicit ranks combine deterministically: sources are
+    /// ordered by `(priority.unwrap_or(list_index), list_index)`, so an
+    /// unspecified source uses its list index as its effective priority, and
+    /// the list index always breaks any remaining ties. This guarantees every
+    /// source in a query gets a distinct, stable rank. The field is inert for
+    /// single-source queries and for events with distinct timestamps.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub priority: Option<i64>,
```
danielgerlag, line 161: Do we need a user-facing `priority` for this fix? The `sources` list already gives us a stable order.

Mixing “I set `priority: 1`” with “I didn’t set it, so I get my list index” is a bit surprising — the second source with no priority and an explicit `1` look the same until the index tiebreak.

If we just used declaration order, we could drop this field, `from_source_with_priority`, and a chunk of config. Happy to keep a knob later if someone actually needs to reorder without reshuffling the list.

lib/src/channels/events.rs
```diff
@@ -218,42 +228,45 @@ pub enum SourceEvent {
...
+    /// Optional source-supplied sequence (the WAL fast-path).
+    ///
+    /// **Most sources leave this `None`** and let the framework assign the
+    /// sequence — that is the recommended path (the source owns the *position*;
+    /// the framework owns the *sequence*). A source with its own durable,
+    /// monotonic ordinal (e.g. a WAL append index) *may* supply it here; the
+    /// framework then uses that value and advances its counter past it to stay
+    /// monotonic. It is never possible to omit the sequence downstream: if this
+    /// is `None`, the framework still stamps one.
+    pub supplied_sequence: Option<u64>,
```
danielgerlag, line 265: Draft vs stamped is the part of this PR I like most — “no sequence past the source boundary” as a type.

`supplied_sequence` undercuts that a little: drafts can carry a sequence again. Could WAL talk to `SourceBase::set_next_sequence` instead, so the draft stays sequence-free and the framework is still the only one that stamps?

lib/src/queries/manager.rs
```diff
@@ -134,6 +134,37 @@ fn convert_variable_value_to_json(value: &VariableValue) -> serde_json::Value {
...
+/// `priority.unwrap_or(list_index)`; sources are ordered by
+/// `(effective_priority, list_index)` and assigned distinct ranks `0..n` by
+/// that order.
+///
+/// This honors explicit priorities, falls back to declared order for
+/// unspecified sources, and always breaks remaining ties by list index — so
+/// every source gets a stable, distinct rank and cross-source ordering is
+/// deterministic. The returned vec is indexed by the source's list position:
+/// `ranks[i]` is the rank of the `i`-th declared source.
+pub(super) fn compute_source_ranks(priorities: &[Option<i64>]) -> Vec<u32> {
```
danielgerlag, line 151: Worth double-checking the merge rule we want on a timestamp tie.

`(timestamp, rank, sequence)` means: at time T, all of source A, then all of source B. That’s deterministic, but it’s a different join than arrival order — `A, B, A, B` becomes `A, A, B, B`.

The original bug was same-source reorder (HTTP batches). Sequence as the last key fixes that. Cross-source grouping is extra semantics. If we do want a stable cross-source order, declaration order is enough — I just want us to be intentional that this is the join policy, not only a heap tiebreak.

danielgerlag (commented):
Took a pass focused on simplifying the ordering story, not on nits.

The composite key is the right idea — each source is a totally ordered stream, and the query merges them by time. A few places I think we could make that easier to follow; notes inline. Happy to talk through any of this.

lib/src/sources/future_queue_source.rs
```diff
@@ -163,11 +169,16 @@ impl FutureQueueSource {
...
+                    // Item is due — dispatch FuturesDue signal. This source
+                    // bypasses SourceBase, so it stamps a monotonic sequence
+                    // itself to honor the mandatory-sequence contract.
+                    let draft = SourceEventDraft::new(
                         FUTURE_QUEUE_SOURCE_ID.to_string(),
                         SourceEvent::Control(SourceControl::FuturesDue),
                         timestamp,
                     );
+                    let seq = next_sequence.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+                    let event_wrapper = StampedSourceEvent::stamp(draft, seq);
```
amansinghoriginal, line 181: Could `FuturesDue` bypass source-checkpoint dedup? `__future_queue__` is also an accepted source ID, so a source checkpoint can suppress these synthetic sequences. With checkpoint `100001`, my 100 ms `drasi.trueFor` condition fires on `main` but not within one second on this branch. Handling the wakeup separately seems simpler.

lib/src/config/schema.rs
amansinghoriginal, line 161: Could we use declaration order for now and leave numeric priorities until there’s a concrete need? That would let us remove this field, the mixed explicit/implicit ranking, and `from_source_with_priority`, while retaining a stable tie-break between buffered sources.

lib/src/channels/priority_queue.rs
```diff
@@ -21,39 +21,63 @@ use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering as AtomicOrdering};
...
+    fn new(event: Arc<T>, source_rank: u32) -> Self {
+        Self { event, source_rank }
+    }
+
+    /// The composite ordering key: `(timestamp, source_rank, sequence)`.
+    fn order_key(&self) -> (chrono::DateTime<chrono::Utc>, u32, u64) {
+        (
+            self.event.timestamp(),
+            self.source_rank,
+            self.event.sequence(),
```
amansinghoriginal, line 60: Would it make sense to settle on one per-source ordering rule? `dispatch_event` already makes wrapper timestamps strictly increasing, so same-source ties normally never reach this key. If we rely on sequence for ties instead, could we keep timestamps nondecreasing so clock rollback still cannot reverse events?

lib/src/queries/manager.rs
```diff
@@ -134,6 +134,37 @@ fn convert_variable_value_to_json(value: &VariableValue) -> serde_json::Value {
...
+/// order: `Some(p)` for an explicit `priority`, `None` to fall back to the
+/// source's list position. A source's effective priority is
+/// `priority.unwrap_or(list_index)`; sources are ordered by
+/// `(effective_priority, list_index)` and assigned distinct ranks `0..n` by
+/// that order.
+///
+/// This honors explicit priorities, falls back to declared order for
+/// unspecified sources, and always breaks remaining ties by list index — so
+/// every source gets a stable, distinct rank and cross-source ordering is
+/// deterministic. The returned vec is indexed by the source's list position:
```
amansinghoriginal, line 149: Could “deterministic” here be limited to events already buffered? With the same timestamps and ranks, I reproduced `A -> B` when both were queued and `B -> A` when A arrived after B was processed. I’d document that limit rather than add cross-source coordination for it.

lib/src/channels/priority_queue.rs
amansinghoriginal, line 218: nit: Would a query-specific queue entry keep `source_rank` out of this generic API? Reactions and helpers currently have to pass `0`, although source ranking only has meaning at the query boundary.

lib/src/channels/events.rs
amansinghoriginal, line 265: Could ordinary drafts stay sequence-free, with an explicit trusted path for WAL/replay reconstruction? Allowing `supplied_sequence` here weakens the ownership distinction this split is trying to establish. WAL-assigned sequence values would still need to remain stable during replay.

amansinghoriginal (changes requested):
Thanks for the work. I’ve left a few concerns inline.

ruokun-niu:
- remove priority order

lib/src/channels/priority_queue.rs
ruokun-niu: Agreed. Removed the 1 ns bump: equal timestamps now remain equal, and sequence breaks same-source ties

lib/src/channels/priority_queue.rs
ruokun-niu: Good suggestion. I have applied.

lib/src/config/schema.rs
ruokun-niu: As discussed in our PR review meeting, we have decided to remove the priority field.

lib/src/queries/manager.rs
ruokun-niu: Updated the documentation

lib/src/channels/events.rs
ruokun-niu: We have removed `supplied_sequence` field based on our discussion

ruokun-niu: We have removed `supplied_sequence` field and the `SourceEventDraft` struct based on our discussion

lib/src/channels/priority_queue.rs
ruokun-niu: Agreed. I have already moved source_rank into a query-specific RankedSourceEvent wrapper. Generic enqueue and enqueue_wait now take only the event, so reactions and helpers no longer pass 0. Query forwarders explicitly construct ranked entries before enqueueing.

lib/src/queries/manager.rs
ruokun-niu: Already addressed

lib/src/channels/priority_queue.rs
ruokun-niu: Agreed. Removed the 1 ns increment and now clamp timestamps to be nondecreasing. Equal timestamps remain equal, with sequence breaking same-source ties.

lib/src/config/schema.rs
ruokun-niu: The `priority` field has already been removed

lib/src/sources/future_queue_source.rs
ruokun-niu: fixed

lib/src/channels/events.rs
amansinghoriginal: nit: `supplied_sequence` is still present in this revision. Keeping the WAL path seems reasonable for now; could we adjust the docs and earlier reply to reflect that, rather than saying it was removed? This doesn't need to block approval.

amansinghoriginal (approved):
Thanks for the updates. The revised implementation looks good to me.

agentofreality (approved):

PR #863, opened 2026-09-08
fix(release): validate package visibility access early

### LINKED ISSUE #862: Fail releases early when GHCR visibility credentials are invalid
## Area for Improvement

Validate GHCR package-visibility credentials before publishing crates or building plugins.

## Observed Behavior

[Release run 33420502105](https://github.com/drasi-project/drasi-core/actions/runs/33420502105) successfully published all crates and eight plugin targets, then failed because `Set Package Visibility to Public` received HTTP 404 responses.

The workflow silently falls back to `GITHUB_TOKEN` when `PACKAGES_ADMIN_TOKEN` is unavailable, but `GITHUB_TOKEN` cannot administer organization package visibility. Credential problems are therefore detected only after the release work completes.

## Desired Behavior

Fail immediately with a clear error if package-visibility access is missing or invalid, before publishing any artifacts.

## Proposed Fix

- Store a PAT classic with `write:packages` in a dedicated, `main`-restricted environment.
- Add a preflight API call against a known package before publishing.
- Remove the ineffective `GITHUB_TOKEN` fallback.
- Generate the plugin list before exposing the privileged token.
- Expose the token only to package-visibility API calls.
- Optionally run the same preflight on a schedule to detect expiration before a release.

## Additional Context

The release artifacts were published successfully; only the final visibility operation failed. Related improvements: #767, #773, and #784.

amansinghoriginal:
# Description

- validate `PACKAGES_ADMIN_TOKEN` before publishing crates or starting plugin builds
- remove the ineffective `GITHUB_TOKEN` fallback and pass only the required secret to the reusable workflow
- keep the privileged token out of Cargo execution and make visibility updates idempotent and verified
- add focused shell tests for missing, denied, public, private, and invalid-package cases

## Type of change

Fixes: #862


.github/workflows/release-plz.yml
```diff
@@ -121,4 +127,5 @@ jobs:
     with:
       sign: true
       dry_run: false
-    secrets: inherit
\ No newline at end of file
+    secrets:
+      PACKAGES_ADMIN_TOKEN: ${{ secrets.PACKAGES_ADMIN_TOKEN }}
```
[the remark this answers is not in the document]
amansinghoriginal: `PACKAGES_ADMIN_TOKEN` is configured as a repository secret, not an environment secret; `drasi-core-release` does not contain a secret with that name. The explicit forwarding therefore works as written. Moving the token to a protected environment would require separate workflow architecture and manual configuration, so I suggest treating that as follow-up scope.

.github/workflows/publish-plugins.yml
```diff
@@ -85,8 +89,22 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  visibility-preflight:
+    name: Validate Package Visibility Access
+    runs-on: ubuntu-latest
+    steps:
```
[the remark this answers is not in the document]
amansinghoriginal: Addressed in 5956fab6. `visibility-preflight` now has job-level `contents: read` permissions, removing its inherited `packages: write` and `id-token: write` access.

.github/workflows/publish-plugins.yml
```diff
@@ -359,6 +377,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
+          ref: ${{ inputs.ref }}
```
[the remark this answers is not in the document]
amansinghoriginal: Addressed in 5956fab6. Package discovery from `inputs.ref` now runs in a separate read-only job without the PAT. The visibility job checks out the trusted workflow revision and validates all package names and namespaces before making API calls.

.github/scripts/package-visibility.sh
```diff
@@ -0,0 +1,97 @@
...
+}
+
+validate_package() {
+    local package="$1"
+
+    if ! [[ "$package" =~ ^[a-z0-9][a-z0-9._-]*(/[a-z0-9][a-z0-9._-]*)*$ ]]; then
+        fail "Invalid package name: $package"
+    fi
+    case "$package" in
+        drasi-plugin-directory | source/* | reaction/* | bootstrap/* | identity/* | secret-store/*) ;;
```
danielgerlag, line 25: We should not restrict by these package prefixes

.github/workflows/publish-plugins.yml
```diff
@@ -75,6 +75,10 @@ on:
         required: false
         type: string
         default: ''
+    secrets:
+      PACKAGES_ADMIN_TOKEN:
```
danielgerlag, line 79: Will this expire?

danielgerlag (changes requested):

.github/workflows/publish-plugins.yml
amansinghoriginal: Expires September 7, 2027; manual rotation documented in the workflow README.

.github/scripts/package-visibility.sh
amansinghoriginal: Removed prefix restrictions; syntax validation remains.

PR #903, opened 2026-09-13
Forward source subscription completion through plugin ABI

### LINKED ISSUE #774: Forward source subscription-complete lifecycle through dynamic plugin ABI
### Steps to reproduce

1. Implement a Source plugin whose `on_subscriptions_complete` records invocation or releases a visible barrier.
2. Load it dynamically through the plugin SDK/host SDK path.
3. Configure multiple persistent continuous queries.
4. Start/restart Drasi and let startup register subscriptions.
5. Observe the Source implementation.

For a WAL-backed Source, allow early subscribers to catch up before later subscribers register.

### Observed behavior

`SourceManager` invokes `on_subscriptions_complete` after auto-start query subscriptions are registered, but dynamically loaded Source plugins do not receive it. `SourceProxy` falls back to the trait default no-op because the Source plugin vtable has no corresponding function.

Sources that use this lifecycle signal to release startup barriers, such as WAL pruning fences, therefore behave differently when dynamically loaded.

A dynamic Source delayed WAL pruning until persistent query subscribers restored old watermarks. The Source expected `on_subscriptions_complete` to release that fence.

In the real host, `SourceManager::subscriptions_complete` ran, but the plugin never observed its callback and eventually used a 60-second fallback.

Before the fence, restart manifested as:

1. Three queries registered position-16 handles and replayed WAL 17–27.
2. The Source pruned through 27 using only currently registered handles.
3. Five later persistent queries registered their own position-16 checkpoints.
4. They failed with `PositionUnavailable`, earliest retained position 28.

A delay is not a correctness boundary; the host already knows when startup subscription registration is complete.

### Desired behavior

The dynamic Source receives exactly one callback after startup subscriptions register, and callback errors propagate consistently.

Please investigate representing the callback in the Source vtable; forwarding it through adapters and `SourceProxy`; propagating errors consistently; handling ABI/version compatibility explicitly; and adding integration coverage proving the dynamic Source receives the signal before pruning and late old-watermark subscribers replay successfully.

### Workaround

The Source eventually uses a 60-second fallback, but this is not a correctness boundary and does not prevent restart-time WAL pruning races.

### Drasi Version

Local prototype build of `drasi-core`; no Core patch has been committed or proposed.

### Operating system

Not provided.

### Additional context

**Impact**

- Dynamic and statically invoked Sources have different lifecycle semantics.
- Sources cannot reliably protect restart-time WAL history until persistent subscribers restore watermarks.
- Late queries may fail with `PositionUnavailable` or require timing fallbacks.
- Adding the callback affects plugin ABI compatibility/versioning.

**Relevant code**

- `lib/src/sources/traits.rs`
- `lib/src/sources/manager.rs`
- `lib/src/lifecycle.rs`
- `components/host-sdk/src/proxies/source.rs`
- `components/plugin-sdk/src/ffi/vtables.rs`
- `components/plugin-sdk/src/ffi/vtable_gen.rs`
- related ABI metadata/versioning.

Reproduced with a dynamically loaded WAL-backed GitHub WorkGraph Source during a local prototype run. No Core patch has been committed or proposed.

Not specified.

### LINKED ISSUE #902: Plugin loader skips ABI compatibility checks when metadata is missing
### Steps to reproduce

1. Build a dynamic plugin library that exports `drasi_plugin_init` and a registration/vtable layout accepted by the loader, but does **not** export `drasi_plugin_metadata`.
2. Alternatively, export `drasi_plugin_metadata` and return a null pointer.
3. Build the plugin against an older SDK whose registration or component vtable is physically smaller than the current host layout.
4. Load it with the current `PluginLoader`/`load_plugin_from_path` path.
5. Observe whether loading fails before the host reads version-gated trailing fields.

A historical regression fix is preserved in sanitized patch `44342b390c501032a24519c087cb4df4cd1e539f.patch` under draft PR #890. The same commit also added Reaction checkpoint-ownership forwarding, but missing metadata is a distinct loader safety root cause.

### Observed behavior

Current `main` at `e759606fa065bee0ef9e60017e263f0f85dd0e44` treats absent or null metadata as a warning and returns `Ok(None)` from `validate_plugin_metadata`. Loading continues after explicitly skipping SDK version and target checks.

Later code gates some trailing registration fields on the optional SDK version, but a plugin with missing metadata can still enter the loader without proving that its physical FFI layouts are compatible. This defeats the compatibility boundary and makes future vtable extension safety depend on every call site remembering to handle `None` conservatively.

### Desired behavior

Dynamic plugin metadata is mandatory for loading. A missing export, null pointer, malformed version, incompatible major/minor version, or target mismatch must fail loading before initialization or any version-dependent vtable access.

### Workaround

Require plugin producers operationally to use the current SDK. This is not enforceable by the host and does not protect against stale, malformed, or third-party binaries.

### Drasi Version

Confirmed by source inspection on `main` `e759606fa065bee0ef9e60017e263f0f85dd0e44`. Historical candidate fix: `44342b390c501032a24519c087cb4df4cd1e539f` (original object preserved as a sanitized portable patch in #890 because the commit no longer resolves through a remote ref/GitHub commit endpoint).

### Operating system

Dynamic loader behavior; reproduce on each supported plugin target because target-triple validation is part of the skipped metadata check.

### Additional context

**Impact**

- An incompatible plugin can bypass the host's declared ABI version check.
- Reading fields beyond an older physical vtable/registration layout can become undefined behavior.
- Missing target validation can load an incompatible binary far enough to fail unpredictably.

**Root-cause confidence:** high. `components/host-sdk/src/loader.rs::validate_plugin_metadata` explicitly warns and returns `Ok(None)` for both missing metadata export and null metadata.

**Possible fix**

Make `validate_plugin_metadata` return `Result<String>` and fail closed for missing/null metadata, as in the preserved historical patch. Keep inspection-only helpers such as `scan_plugin_metadata` separately optional if their API contract needs `None`; do not let the actual load path silently downgrade compatibility enforcement.

**Acceptance criteria**

- [ ] Loading fails before plugin initialization when `drasi_plugin_metadata` is absent.
- [ ] Loading fails before plugin initialization when metadata is null or malformed.
- [ ] SDK major/minor and target mismatch tests remain explicit failures.
- [ ] A compatibility fixture with an older physically smaller layout cannot reach any trailing-field read.
- [ ] Current compatible plugins continue to load.
- [ ] Loader and inspection-only metadata APIs document their different failure contracts.

Tracked by #889. The Reaction checkpoint-ownership subcase from the same historical commit remains mapped to #820 rather than this loader issue.

Yes, I would like to support you.

agentofreality:
# Description

Dynamic source plugins now receive `Source::on_subscriptions_complete` after all auto-start query subscriptions have registered. Previously, `SourceProxy` inherited the trait's no-op default because `SourceVtable` had no callback slot, so replay/pruning fences could remain active until a timing fallback.

This change:

- adds a fallible `on_subscriptions_complete_fn` to `SourceVtable` and wires both typed and boxed plugin adapters through the existing panic/error bridge;
- forwards the callback through `SourceProxy`, `SourceManager`, and `LifecycleManager`, preserving source errors at the host boundary;
- invokes each successfully auto-started source once without racing asynchronous graph status updates;
- bumps the FFI contract to `0.15.0` and requires valid plugin metadata before initialization, ensuring pre-0.15 plugins with physically smaller source vtables are rejected before any trailing slot is read;
- adds unit, lifecycle, real-cdylib, old-vtable, and missing-metadata regressions using generic counting/failing mock sources and two startup query subscriptions.

**Reproduction:** load a dynamic source that uses `on_subscriptions_complete` as a startup fence, configure multiple auto-start queries, and start Drasi. Before this change the dynamic source never observed the callback; after this change it observes exactly one callback after both subscriptions are registered, and callback errors fail startup instead of being swallowed.

**Source provenance:** selectively ported the generic lifecycle callback design from retained commit `1bd669a389924cc264ef071f5f97624b5618ef08`, verified against retained branch state `fb5d2bda2cbce4284fecbce40cae2a2d74116249`, and adapted to current `main` APIs and ABI conventions. No WorkGraph-specific code, SourceBase FIFO changes, or Reaction acknowledgement changes are included.

**Validation:** `cargo fmt --all -- --check`; targeted Clippy for `drasi-plugin-sdk`, `drasi-host-sdk`, `drasi-lib`, `drasi-source-mock`, and `drasi-source-postgres`; typed/boxed vtable tests; proxy slot tests; loader tests; real cdylib lifecycle test; old/missing-metadata ABI tests; and lifecycle ordering/error propagation tests.

## Type of change

Fixes #774
Fixes #902

### DIFF
--- a/components/host-sdk/src/loader.rs
+++ b/components/host-sdk/src/loader.rs
@@ -350,9 +350,7 @@ pub fn load_plugin_from_path(
     // safely read the trailing fields.
     let identity_provider_vtables: Option<
         Vec<drasi_plugin_sdk::ffi::IdentityProviderPluginVtable>,
-    > = if plugin_sdk_version
-        .as_deref()
-        .and_then(parse_semver)
+    > = if parse_semver(&plugin_sdk_version)
         .map(|v| v >= MIN_SDK_VERSION_WITH_IDENTITY_PROVIDERS)
         .unwrap_or(false)
         && !registration.identity_provider_plugins.is_null()
@@ -387,16 +385,15 @@ pub fn load_plugin_from_path(
     // NOTE: `set_log_level` is a trailing field appended for SDK 0.12.0; gate
     // access on the plugin's reported `sdk_version` like the identity fields
     // above, since reading it from an older (smaller) layout is UB.
-    let set_log_level_fn: Option<extern "C" fn(FfiLogLevelFilter)> = if plugin_sdk_version
-        .as_deref()
-        .and_then(parse_semver)
-        .map(|v| v >= MIN_SDK_VERSION_WITH_SET_LOG_LEVEL)
-        .unwrap_or(false)
-    {
-        Some(registration.set_log_level)
-    } else {
-        None
-    };
+    let set_log_level_fn: Option<extern "C" fn(FfiLogLevelFilter)> =
+        if parse_semver(&plugin_sdk_version)
+            .map(|v| v >= MIN_SDK_VERSION_WITH_SET_LOG_LEVEL)
+            .unwrap_or(false)
+        {
+            Some(registration.set_log_level)
+        } else {
+            None
+        };
 
     // Report the host's effective log level so the plugin can drop
     // filtered-out records before formatting or crossing the FFI.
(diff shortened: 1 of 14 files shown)


agentofreality:
This PR's metadata hardening appears to cover the implementation tracked by #902 in addition to its primary #774 Source lifecycle fix: the PR description requires valid plugin metadata before initialization and lists old-vtable plus missing-metadata regressions.

The distinct loader finding and historical fix provenance are preserved at:

- [`report-L282`](https://github.com/drasi-project/drasi-core/blob/a7cb89f2a7281bee28c0f79acea3761c9a8d22b8/docs/history/workgraph-core-findings/evidence.md#report-l282)
- [sanitized `44342b39` patch](https://github.com/drasi-project/drasi-core/blob/a7cb89f2a7281bee28c0f79acea3761c9a8d22b8/docs/history/workgraph-core-findings/patches/44342b390c501032a24519c087cb4df4cd1e539f.patch)

No separate implementation should be opened while #903 is the active candidate. #902 should remain linked until its missing export, null metadata, malformed metadata, target/version mismatch, and physically smaller legacy-layout acceptance criteria are confirmed against this PR.

PR #932, opened 2026-09-18
fix: trim outbox inside the output transaction and test overflow gaps after restart

danielgerlag:
Related #743, #818, #822, #835. Replaces #927, which GitHub would not retarget off the stack after #835 merged to main.

## Why

`stage_durable_query_output` only appended the new outbox key inside the session. RocksDB and Garnet `append` are puts. They do not evict. After commit, `dispatch_query_results` called `trim_to_capacity` as a standalone write and ignored failure with `warn!`.

A crash between commit and that trim, or a failed trim, left N+1 durable keys. After restart, hydrate could pretend the prefix was gone while disk still held it. `fetch_outbox` then lied about `earliest_available`.

## What changed

Evict keys below `next_seq - capacity + 1` in the same session as append, live-results, and result sequence. Trim failure aborts the transaction. The post-commit best-effort trim is gone.

RocksDB deletes go through `with_txn` when a session is active. Garnet deletes go through the session buffer. After commit, the in-memory ring and the durable outbox share the same earliest sequence.

## Overflow reconstruction

Full-process seed/recover in `reaction_recovery_conformance.rs` (RocksDB + Redb, `outbox_capacity = 2`):

- `rocksdb_redb_overflow_gap_strict` after hydrate, durable key count is 2, `fetch_outbox(C)` is `OutboxGap { requested=C, earliest_available=E, latest_sequence=HWM }`, reaction does not reach Running
- `rocksdb_redb_overflow_gap_autoreset` snapshot at HWM, checkpoint becomes (HWM, hash), no replay of C+1..E-1 as live side effects
- `rocksdb_redb_overflow_gap_autoskip` checkpoint jumps to HWM, no replay of the missing range

The parent process counts durable keys after the seed child exits, so a crash between commit and the old post-commit trim cannot leave N+1 keys.

## Tests

Passed on the stacked PR:

- `cargo test -p lib-integration-tests --test reaction_recovery_conformance`
- `cargo test -p lib-integration-tests --test query_output_atomicity`
- `cargo test -p drasi-lib --test reaction_recovery_e2e`
- `cargo clippy -p drasi-lib --all-targets -- -D warnings`
- `cargo fmt --all -- --check`

Garnet outbox tests remain `#[ignore]` (need Redis). They compile.


danielgerlag:
Duplicate of #927. Opened by mistake; restoring the original PR so reviews/approvals stay there.

